PR #79, opened 2023-08-14
add in trend percentage parameter

npusarla:


src/seer/trend_detection/trend_detector.py
```diff
@@ -18,7 +18,7 @@
 from seer.trend_detection.detectors.cusum_detection import CUSUMDetector
 
 
-def find_trends(txns_data, sort_function, zerofilled, pval=0.01, trend_perc=0.05):
+def find_trends(txns_data, sort_function, zerofilled, trend_perc, pval=0.01):
```
udameli, line 21: can we keep the default as `0.05` for backwards compatibility

udameli: so that you can deploy this change without needing to change the backend

npusarla: I'm specifying the default value to be 0.05 when i get the parameter in the first place here: 
`lower_limit_trend_percentage = float(data.get('trend_percentage()', 0.05))`

npusarla: or i see this is to not change the backend - got it will keep it

src/seer/seer.py
```diff
@@ -66,11 +66,18 @@ def breakpoint_trends_endpoint():
 
         sort_function = data.get('sort', "")
 
+        lower_limit_trend_percentage = data.get('trend_percentage()', 0.05)
+
+        if lower_limit_trend_percentage[0] == ">":
```
udameli, line 71: you know maybe it doesn't make sense for me to send you `>` if there's no upper limit. I can just send you a value directly `trend_percentage():0.25`

Zylphrex: Since this is a query parameter, I think it makes sense to just send the numeric value without the `>`.

src/seer/seer.py
```diff
@@ -66,11 +66,18 @@ def breakpoint_trends_endpoint():
...
+            lower_limit_trend_percentage = lower_limit_trend_percentage[1:]
+
+        lower_limit_trend_percentage = float(lower_limit_trend_percentage)
+
         with sentry_sdk.start_span(
                 op="cusum.detection", description="Get the breakpoint and t-value for every transaction"
         ) as span:
 
-            trend_percentage_list = find_trends(txns_data, sort_function, zerofilled)
+            trend_percentage_list = find_trends(txns_data, sort_function, zerofilled, lower_limit_trend_percentage)
```
Zylphrex, line 80: You could do something like this to avoid having to change the function signature at all. Though considering this is the only caller, and you have the default value in place, I don't have strong opinions here.

```suggestion
            trend_percentage_list = find_trends(txns_data, sort_function, zerofilled, trend_perc=lower_limit_trend_percentage)
```

Zylphrex (approved):

udameli (approved):

PR #81, opened 2023-08-16
add severity endpoint v0

trillville:
still WIP, but can use this as a reference for inputs/outputs


src/seer/seer.py
```diff
@@ -40,19 +41,35 @@ def traces_sampler(sampling_context):
...
+        app.logger.exception("Error processing request")
+        return {"Error": str(e)}, 500
+
+
 @app.route("/issues/severity-score", methods=["POST"])
 def def_severity_endpoint():
     try:
         data = request.get_json()
-        severity = float(data.get("severity", 0.5))
+        severity = embeddings_model.severity_score(data["message"])
```
wedamija, line 67: Nit: `data.get("message", "")`?

PR #87, opened 2023-09-21
for trends boolean

npusarla:
add in a boolean to check if breakpoint analysis is being used for trends or not 
- will continue with midpoint analysis for trends, otherwise will just move on to the next transaction

### DIFF
--- a/src/seer/seer.py
+++ b/src/seer/seer.py
@@ -86,6 +86,8 @@ def breakpoint_trends_endpoint():
 
         sort_function = data.get("sort", "")
 
+        trends_boolean = data.get("trends", True)
+
         lower_limit_trend_percentage = float(data.get('trend_percentage()', 0.1))
 
         with sentry_sdk.start_span(
@@ -96,6 +98,7 @@ def breakpoint_trends_endpoint():
                 txns_data,
                 sort_function,
                 zerofilled,
+                trends_boolean,
                 trend_perc=lower_limit_trend_percentage,
             )
 
--- a/src/seer/seer.py
+++ b/src/seer/seer.py
@@ -86,7 +86,7 @@ def breakpoint_trends_endpoint():
 
         sort_function = data.get("sort", "")
 
-        trends_boolean = data.get("trends", True)
+        allow_midpoint  = data.get("allow_midpoint", "1") == "1"
 
         lower_limit_trend_percentage = float(data.get('trend_percentage()', 0.1))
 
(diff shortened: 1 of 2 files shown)


src/seer/trend_detection/trend_detector.py
```diff
@@ -98,9 +98,13 @@ def find_trends(txns_data, sort_function, zerofilled, trend_perc=0.1, pval=0.01)
...
-        # if breakpoint is in the very beginning or no breakpoints are detected, use midpoint analysis instead
-        if num_breakpoints == 0 or change_index <= req_start_index + 15 or change_index >= len(timestamps) - 10:
-            change_point = (req_start + req_end) // 2
+        #check the trends boolean - don't get midpoint of the request period if the trends boolean is false 
+        if not trends:
+            continue
+        else:
+            # if breakpoint is in the very beginning or no breakpoints are detected, use midpoint analysis instead
+            if num_breakpoints == 0 or change_index <= req_start_index + 15 or change_index >= len(timestamps) - 10:
+                change_point = (req_start + req_end) // 2
```
Zylphrex, line 107: else block seems unnecessary

src/seer/seer.py
```diff
@@ -86,6 +86,8 @@ def breakpoint_trends_endpoint():
 
         sort_function = data.get("sort", "")
 
+        trends_boolean = data.get("trends", True)
```
Zylphrex, line 89: If I understood this correctly, then the only way to disable it is to pass `trend=` with the empty string? I'm not even sure that works.

Can we change this to something like

```python
allow_midpoint  = data.get("allow_midpoint", "1") == "1"
```

So we can pass `"1"` and `"0"` to toggle it?

src/seer/trend_detection/trend_detector.py
npusarla: oh thats true haha

src/seer/seer.py
npusarla: you would have to pass {"trends": False}, but ya let me update it to that

npusarla: updated

Zylphrex: Because flask deserializes these values from the query parameters, best we can do is pass `{"trends": "False"}` which then means the boolean will actually contain the string `"False"` which is truthy. So converting this to something more obvious will be less ambiguous.

Zylphrex (approved):

PR #88, opened 2023-09-29
fix: Correctly disable midpoint fallback

Zylphrex:
Only check the allow_midpoint fallback right before the fallback. Having it outside means there will never be a breakpoint when midpoints are disallowed.


trillville (approved):

Zylphrex:
Closing in favour of #89

PR #94, opened 2023-10-07
feat: add minimum ms parameter

trillville:
- black format trend_detector.py
- add `min_ms_change` parameter with a default value of 25 (miliseconds)
- remove unused `zerofilled` parameter
- make variable names more readable
- remove duplication of default value logic

@udameli @Zylphrex lmk what you guys think makes sense for a default value for `min_ms_change`. Also, we may want to set it to a lower value or remove it for trends so stuff like this still shows up
image

### DIFF
--- a/src/seer/seer.py
+++ b/src/seer/seer.py
@@ -83,7 +83,7 @@ def breakpoint_trends_endpoint():
                 sort_function,
                 allow_midpoint,
                 min_pct_change,
-                min_ms_change
+                min_change
             )
 
         trends = {"data": [x[1] for x in trend_percentage_list]}
--- a/src/seer/seer.py
+++ b/src/seer/seer.py
@@ -68,14 +68,11 @@ def breakpoint_trends_endpoint():
         data = request.get_json()
         txns_data = data["data"]
 
-        # new format has zerofilled parameter - if it's not being sent to microservice default value is True
-        zerofilled = data.get("zerofilled", True)
-
         sort_function = data.get("sort", "")
-
         allow_midpoint  = data.get("allow_midpoint", "1") == "1"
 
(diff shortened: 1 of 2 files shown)


src/seer/seer.py
```diff
@@ -68,14 +68,11 @@ def breakpoint_trends_endpoint():
...
-        # new format has zerofilled parameter - if it's not being sent to microservice default value is True
-        zerofilled = data.get("zerofilled", True)
-
         sort_function = data.get("sort", "")
-
         allow_midpoint  = data.get("allow_midpoint", "1") == "1"
 
-        lower_limit_trend_percentage = float(data.get('trend_percentage()', 0.1))
+        min_pct_change = float(data.get('trend_percentage()', 0.1))
+        min_ms_change = float(data.get('min_ms_change()', 25.0))
```
Zylphrex, line 75: I think we should default this to 0 and let the caller decide. Functions use nanoseconds as the unit so forcing a unit here is not ideal.

```suggestion
        min_change = float(data.get('min_change()', 0))
```

src/seer/trend_detection/trend_detector.py
```diff
@@ -18,23 +18,29 @@
 from seer.trend_detection.detectors.cusum_detection import CUSUMDetector
 
 
-def find_trends(txns_data, sort_function, zerofilled, allow_midpoint, trend_perc=0.1, pval=0.01):
+def find_trends(
+    txns_data,
+    sort_function,
+    allow_midpoint,
+    min_pct_change,
+    min_ms_change,
```
Zylphrex, line 26: And update this argument name too

```suggestion
    min_change,
```

Zylphrex (approved):

PR #95, opened 2023-10-16
feat: require last 24h to also be regressed 

trillville:
- add an additional requirement that the last 24h of data meets `min_change` and `min_pct_change`

### DIFF
--- a/src/seer/trend_detection/trend_detector.py
+++ b/src/seer/trend_detection/trend_detector.py
@@ -196,5 +196,3 @@ def find_trends(
             trend_percentage_list.append((trend_percentage, output_dict))
 
     return trend_percentage_list
-
-
--- a/src/seer/trend_detection/trend_detector.py
+++ b/src/seer/trend_detection/trend_detector.py
@@ -156,6 +156,20 @@ def find_trends(
         # TREND LOGIC:
         #  1. p-value of t-test is less than passed in threshold (default = 0.01)
         #  2. trend percentage is greater than passed in threshold (default = 10%)
+        #  3. last 24 hours are also greater than threshold
+
+        timestamp_24h_ago = req_end - 24 * 60 * 60
+
+        # Filter out the data for the last 24 hours
+        last_24h_data = [
+            metric for timestamp, metric in zip(timestamps, metrics)
+            if timestamp_24h_ago <= timestamp <= req_end
+        ]
+
+        # Calculate the trend percentage and change for the last 24 hours
+        mu_24h = np.average(last_24h_data)
+        trend_percentage_24h = mu_24h / mu0 if mu0 != 0 else mu_24h
+        trend_change_24h = mu_24h - mu0
 
         # most improved - get only negatively significant trending txns
         if (
@@ -163,6 +177,8 @@ def find_trends(
             and mu1 + min_change <= mu0
             and scipy_t_test.pvalue < pval
             and abs(trend_percentage - 1) > min_pct_change
+            and abs(trend_percentage_24h - 1) > min_pct_change
+            and abs(trend_change_24h) > min_change
         ):
             output_dict["change"] = "improvement"
             trend_percentage_list.append((trend_percentage, output_dict))
@@ -173,8 +189,12 @@ def find_trends(
             and mu0 + min_change <= mu1
             and scipy_t_test.pvalue < pval
             and abs(trend_percentage - 1) > min_pct_change
+            and abs(trend_percentage_24h - 1) > min_pct_change
+            and abs(trend_change_24h) > min_change
         ):
             output_dict["change"] = "regression"
             trend_percentage_list.append((trend_percentage, output_dict))
 
     return trend_percentage_list
+
+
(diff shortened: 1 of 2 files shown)


src/seer/trend_detection/trend_detector.py
```diff
@@ -156,13 +156,29 @@ def find_trends(
         # TREND LOGIC:
         #  1. p-value of t-test is less than passed in threshold (default = 0.01)
         #  2. trend percentage is greater than passed in threshold (default = 10%)
+        #  3. last 24 hours are also greater than threshold
+
+        timestamp_24h_ago = req_end - 24 * 60 * 60
```
Zylphrex, line 161: This creates an implicit assumption that trends are >24h long. i.e. for the trends page, if the time period selected is <24h, we should never surface a trend.

Should we consider making this optional and having statistical detectors opt into them?

Additionally, what if the breakpoint is in the last 24 hours? Then the last 24 hours of data will contain some data from before the breakpoint, the actual regression, and some data after. So the results may not be very reliable.

src/seer/trend_detection/trend_detector.py
```diff
@@ -172,7 +188,9 @@ def find_trends(
             (sort_function == "-trend_percentage()" or sort_function == "")
             and mu0 + min_change <= mu1
             and scipy_t_test.pvalue < pval
-            and abs(trend_percentage - 1) > min_pct_change
+            and trend_percentage - 1 > min_pct_change
+            and trend_percentage_24h - 1 > min_pct_change
```
Zylphrex, line 192: I think you need to keep the `abs` here. If the last 24 hours average is slightly less than before the breakpoint. This can happen since the timeseries is noisy and there will inevitably be some differences. Then `trend_percentage_24h` will be less than 1 and this condition will not be true.

The same reasoning applies for `trend_percentage`.

trillville: Agreed that this could potentially filter out a real regression, but I think it's the most straightforward way for us to handle the case where a regression appears and then resolves before an issue is created. 

@Zylphrex what happens with the workflow if it fails at this step? I think ideally we'd retry a few times. Even if we can't though. Overall though I think I'd rather avoid a situation where we are generating a regression where it's clearly already returned to the baseline. This is the notion doc I am going off of: https://www.notion.so/sentry/Known-false-positive-examples-35155f435e114b9b97b3f0505a576569

src/seer/trend_detection/trend_detector.py
trillville: > This creates an implicit assumption that trends are >24h long. i.e. for the trends page, if the time period selected is <24h, we should never surface a trend.
> 
> Should we consider making this optional and having statistical detectors opt into them?

Good point, I'll do this

> Additionally, what if the breakpoint is in the last 24 hours? Then the last 24 hours of data will contain some data from before the breakpoint, the actual regression, and some data after. So the results may not be very reliable.

Hmm good point - remind me, what's the current logic for calling the microservice from sentry? Maybe we add more of a delay there, and I make it so the last 24h have to be after the breakpoint.

Zylphrex: Currently we just have a naive 12 hour delay after the EMA analysis. I think what we could do is have an optional parameter to specify how many hours to use for checking if the timeseries is still in a regressed state and decide on the sentry side. So we can keep these constants in 1 place instead of 2.

src/seer/trend_detection/trend_detector.py
Zylphrex: currently we just try once, and im thinking about how we can implement a retry mechanism

trillville: 👍 got it, i think this is a reasonable change to make even without a retry mechanism, but that will also help mitigate the downside

src/seer/trend_detection/trend_detector.py
trillville: that makes sense, that would imply no code changes on the microservice side right?

Zylphrex: I'm not sure what you mean here exactly.

What I was thinking was the microservice would expose a parameter like `validate_tail` that takes an integer > 0 and defaults to none/0

Then the microservice will take the parameter and validate that the at most the last `n` buckets (specified by `validate_tail`) is still in a regressed state using the algorithm you have here. If it was not specified, it skips this additional validation.

trillville: oh, i gotcha. I thought you were talking about just parameterizing the delay on the sentry side and calling it a day. This makes more sense 👍

src/seer/trend_detection/trend_detector.py
```diff
@@ -156,23 +157,59 @@ def find_trends(
         # TREND LOGIC:
         #  1. p-value of t-test is less than passed in threshold (default = 0.01)
         #  2. trend percentage is greater than passed in threshold (default = 10%)
+        #  3. last validate_tail_hours hours are also greater than threshold
+
+        if validate_tail_hours > 0:
+            validation_start = req_end - validate_tail_hours * 60 * 60
+
+            # Filter out the data for the last 24 hours
```
udameli, line 165: this comment can be updated since it looks like we're not hardcoding the tail to 24 hr

udameli (approved):

PR #116, opened 2023-11-16
Enable tracing

corps:
Allow us to track performance in sentry.io to investigate timeout issues.

### DIFF
--- a/src/seer/seer.py
+++ b/src/seer/seer.py
@@ -1,7 +1,6 @@
 import os
 import time
 
-import numpy as np
 import pandas as pd
 import sentry_sdk
 from flask import Flask, request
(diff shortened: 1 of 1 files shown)


src/seer/seer.py
```diff
@@ -59,54 +59,48 @@ def model_path(subpath: str) -> str:
...
-            )
-
-        trends = {"data": [x[1] for x in trend_percentage_list]}
-        app.logger.debug("Trend results: %s", trends)
-
-        return trends
-    except Exception as e:
-        app.logger.exception("Error processing request")
-        return {"Error": str(e)}, 500
+    data = request.get_json()
```
corps, line 77: We shouldn't need the try/except wrapping, the `FlaskIntegration` from sentry handles this (and does so better).

src/seer/seer.py
```diff
@@ -27,6 +26,7 @@ def traces_sampler(sampling_context):
     integrations=[FlaskIntegration()],
     traces_sampler=traces_sampler,
     profiles_sample_rate=1.0,
+    enable_tracing=True,
```
corps, line 29: This puts performance data of endpoints into sentry.io so we can watch capacity issues.

jennmueng (approved):
lgtm

corps:
🙏

PR #126, opened 2023-11-30
ref: Change severity inference input format

jennmueng:
To match how the model was trained. Don't merge this without coordinate a model switch out.


trillville (approved):

jennmueng:
Note: The failing test case passes with the new model, just tried it locally

PR #127, opened 2023-12-06
add initial grouping model PoC

trillville:


jennmueng (approved):

src/seer/app.py
```diff
@@ -89,6 +98,13 @@ def breakpoint_trends_endpoint(data: BreakpointRequest) -> BreakpointResponse:
     return trends
 
 
+@json_api("/v0/issues/similar-issues")
+def grouping_endpoint(data: GroupingRequest) -> List[GroupingResponse]:
+    with sentry_sdk.start_span(op="seer.grouping", description="grouping lookup") as span:
+        grouping_results = grouping_lookup().find_nearest_neighbor(data)
```
jangjodi, line 104: should this be `.get_nearest_neighbors`?

trillville: oops good catch

src/seer/grouping/grouping.py
```diff
@@ -0,0 +1,144 @@
...
+    group_id: int
+    stacktrace: str
+    k: int = 1
+    threshold: float = 0.99
+
+
+class GroupingRecord(BaseModel):
+    group_id: int
+    stacktrace: str
+    embeddings: Any
```
trillville, line 21: let me know if there is a better way of doing this - it looks like you can't by default require that a data type is `np.ndarray` @corps

corps: numpy typing is still, not great in python.  I'm ok with the any here for now, and I'll keep an eye out for something better.

PR #128, opened 2023-12-08
bump up sentry version and make span instrumentation consistent acros…

trillville:
…s endpoints

### DIFF
--- a/src/seer/seer.py
+++ b/src/seer/seer.py
@@ -60,7 +60,7 @@ if not os.environ.get("PYTEST_CURRENT_TEST"):
 
 
 @app.route("/v0/issues/severity-score", methods=["POST"])
-def def_severity_endpoint():
+def severity_endpoint():
     data = request.get_json()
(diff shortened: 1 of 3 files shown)


src/seer/seer.py
```diff
@@ -60,7 +60,7 @@ def model_path(subpath: str) -> str:
 
 
 @app.route("/v0/issues/severity-score", methods=["POST"])
-def def_severity_endpoint():
+def severity_endpoint():
```
trillville, line 63: lol i randomly noticed this 🤦

src/seer/severity/severity_inference.py
```diff
@@ -1,6 +1,5 @@
 import numpy as np
 import pandas as pd
-import sentry_sdk
```
trillville, line 3: AFAIK this should have been fine, but it's the only thing I could find that is not consistent across endpoints so getting rid of it

jennmueng (approved):

PR #131, opened 2024-01-04
feat(suggested-fix): Initial suggested fix POC implementation

jennmueng:
Initial implementation of suggested fix POC in seer:

`/v0/automation/autofix`: Triggers a celery task to run autofix given an issue title, stack trace, git commit sha (getsentry/sentry), and any additional info. Calls seer rpc endpoint on sentry

- Uses cached embeddings that will be populated in gcs models folder prior to merging this PR.
- Calling the endpoint schedules a celery task
- Celery worker picks up task, embeds the delta between the cached commit and the provided commit sha in the request
- Runs agents with openai api
- Calls sentry backend rpc with results

### Github App
Uses environment variables `GITHUB_APP_ID` and `GITHUB_PRIVATE_KEY` to determine and use the github app.

### Sentry <-> Seer RPC
Communicates with main sentry backend using `SENTRY_BASE_URL`  through a custom RPC client made for this seer service.

### Celery Queue
Uses RabbitMQ celery broker queue, set through `CELERY_BROKER_URL`


src/seer/automation/agent/agent.py
```diff
@@ -0,0 +1,153 @@
...
+                    and self.memory[-1].content is not None
+                )
+            )
+            and self.iterations < self.max_iterations
+        ):
+            # runs until the assistant sends a message with no more tool calls.
+            self.run_iteration()
+
+        if self.iterations == self.max_iterations:
+            raise Exception(f"Agent {self.name} reached maximum iterations without finishing.")
```
trillville, line 92: have you seen this happen? should we try to do a final "wrap things up" prompt rather than just returning the last call?

src/seer/automation/autofix/agent_context.py
```diff
@@ -0,0 +1,354 @@
...
+from llama_index.node_parser import CodeSplitter
+from llama_index.readers import SimpleDirectoryReader
+from llama_index.schema import BaseNode, Document
+
+from seer.automation.autofix.types import AutofixAgentsOutput, FileChange
+from seer.automation.autofix.utils import SentenceTransformersEmbedding
+
+logger = logging.getLogger("autofix")
+
+CACHED_COMMIT_SHA = "db53ec4557fc8e52de8b27f35f977a34e8e73213"
```
trillville, line 26: should we also implement functionality to periodically refresh the most recent sha? for example a cron job that runs every day, reindexes anything changed in last 24h, and updates this variable

trillville: (as a future item)

src/seer/automation/autofix/agent_context.py
```diff
@@ -0,0 +1,354 @@
...
+        nodes: List[BaseNode] = []
+
+        for language, language_docs in documents_by_language.items():
+            logger.debug(f"{language}: {len(language_docs)}")
+
+            splitter = CodeSplitter(
+                language=language,
+                chunk_lines=16,  # lines per chunk
+                chunk_lines_overlap=4,  # lines overlap between chunks
+                max_chars=5096,  # max chars per chunk
```
trillville, line 123: doesnt this correspond to over 1k tokens?

src/seer/automation/autofix/agent_context.py
```diff
@@ -0,0 +1,354 @@
...
+    def _embed_and_index_nodes(self, nodes: list[BaseNode]) -> VectorStoreIndex:
+        service_context = ServiceContext.from_defaults(embed_model=self.embed_model)
+        logger.debug(f"Embedding {len(nodes)} nodes")
+        index = VectorStoreIndex(nodes, service_context=service_context, show_progress=True)
+
+        return index
+
+    def _documents_to_nodes(self, documents: list[Document], max_file_size_bytes=1024 * 1024 * 2):
+        file_extension_mapping = {
+            # "ts": "typescript",
```
trillville, line 95: ahhh right, we will need to revisit this sometime soonish

src/seer/automation/autofix/autofix.py
```diff
@@ -0,0 +1,168 @@
...
+                    ),
+                )
+            ],
+        )
+
+        additional_context_str = (
+            (self.request.additional_context + "\n\n") if self.request.additional_context else ""
+        )
+        planning_response = planning_agent.run(
+            f"{additional_context_str}Note: instead of ./app, the correct directory is static/app/..."
```
trillville, line 120: does this need a TODO?

src/seer/automation/autofix/tools.py
```diff
@@ -0,0 +1,431 @@
...
+class BaseTools:
+    context: AgentContext
+    retriever: VectorIndexRetriever
+
+    retrieved_paths: Set[str]
+    expanded_paths: Set[str]
+
+    def __init__(self, context: AgentContext):
+        self.context = context
+        self.retriever = VectorIndexRetriever(index=self.context.index, similarity_top_k=4)
```
trillville, line 24: have you tried experimenting with this value?

src/seer/automation/autofix/utils.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+    This function computes the similarity between two pieces of text.
+
+    Parameters:
+    text1 (str): The first piece of text.
+    text2 (str): The second piece of text.
+    ignore_whitespace (bool): If True, ignores whitespace when comparing the two pieces of text.
+
+    Returns:
+    float: The similarity ratio between the two pieces of text.
```
trillville, line 25: can you please also indicate what the similarity metric is? not familiar with the `difflab` package

src/seer/celery_app.py
```diff
@@ -0,0 +1,8 @@
+from celery import Celery
```
trillville, line 1: question - should we simply have a directory for celery? or is there not goign to be much baggage

trillville (approved):
🚀 this is incredible stuff

src/seer/app.py
```diff
@@ -7,8 +7,14 @@
 from flask import Flask
 from sentry_sdk.integrations.flask import FlaskIntegration
 
+from seer.automation.autofix.types import (
```
trillville, line 10: should we initialize anything when the worker is started? like loading the embeddings model or anything

src/seer/automation/agent/agent.py
jennmueng: I've seen it happen a handful of times when running the evals, we'll need to experiment with ways to solve this instead of this; I added this just to prevent infinite loops eating up our openai budget for now

src/seer/automation/autofix/agent_context.py
jennmueng: I don't think we should put the effort into periodic reindexing functionality for this poc–a month delta took <3 minutes to embed on cpu. We can just manually update the index every now and then for now.

Let's just skip straight to doing the project & release-tied embedding runs we talked about.

src/seer/automation/autofix/agent_context.py
jennmueng: Hmm yes, it's also limited by 16 lines and the embedding model truncates anything longer than the 512 token input limit anyways. I'll bump it down to 2048 characters to avoid truncation.

src/seer/automation/autofix/autofix.py
jennmueng: Yep, cuz this is specific to the sentry repo and we have to find a way to handle mismatches between the stack trace path and the actual path

src/seer/automation/autofix/tools.py
jennmueng: yes, picked 4 as the middle ground of showing enough matches but not eating up too much context

src/seer/app.py
jennmueng: hmm not too familiar with how to do that in celery but good call, will look into

src/seer/automation/autofix/agent_context.py
trillville: ohh right that makes more sense, agreed

src/seer/automation/autofix/agent_context.py
trillville: is there a reason why we want a line limit as well as a token limit? might be simpler to remove the line limit and limit to 2048 characters

src/seer/app.py
trillville: actually sorry i think my suggestion was dumb, i forgot this isnt something we want running on the API inference endpoint - since the embeddings model is only needed for the celery worker initializing it here is probably a bad idea

src/seer/automation/autofix/agent_context.py
jennmueng: Hmm this is so only the relevant smaller chunks are returned in a retrieval hit instead of a large 5096 char blocks. I also think the overlap is done via lines

src/seer/automation/autofix/agent_context.py
```diff
@@ -0,0 +1,377 @@
...
+from seer.automation.autofix.utils import (
+    MemoryVectorStore,
+    SentenceTransformersEmbedding,
+    generate_random_string,
+    sanitize_branch_name,
+)
+
+logger = logging.getLogger("autofix")
+
+CACHED_COMMIT_SHA = "ac63e3750291a6795d31404030783358ef3ea1ac"
```
trillville, line 34: should update this to the most recent sha before we deploy (if you didnt already)

src/seer/automation/autofix/utils.py
```diff
@@ -0,0 +1,233 @@
...
+
+logger = logging.getLogger("autofix")
+VALID_BRANCH_NAME_CHARS = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ-"
+
+
+def compute_similarity(text1: str, text2: str, ignore_whitespace=True) -> float:
+    """
+    This function computes the similarity between two pieces of text using the difflib.SequenceMatcher class.
+
+    difflib.SequenceMatcher uses the Ratcliff/Obershelp algorithm: it computes the doubled number of matching characters divided by the total number of characters in the two strings.
```
trillville, line 25: thank you for including the description <3

src/seer/automation/autofix/agent_context.py
```diff
@@ -0,0 +1,377 @@
...
+    def _embed_and_index_nodes(self, nodes: list[BaseNode]) -> VectorStoreIndex:
+        service_context = ServiceContext.from_defaults(embed_model=self.embed_model)
+        logger.debug(f"Embedding {len(nodes)} nodes")
+        index = VectorStoreIndex(nodes, service_context=service_context, show_progress=True)
+
+        return index
+
+    def _documents_to_nodes(self, documents: list[Document], max_file_size_bytes=1024 * 1024 * 2):
+        file_extension_mapping = {
+            # "ts": "typescript",
```
trillville, line 103: are we only enabling this for project == 1 (sentry)?

jennmueng: yes

PR #139, opened 2024-01-24
fix(autofix): Fix the vector store index not correctly being loaded + now save the latest index

jennmueng:
### Problem
1. The index was not being correctly loaded because llama_index for some reason was saving the vector store to `default__vector_store.json` when the correct directory is `vector_store.json`? 
2. We were not caching the new index at the new commit

### Fixes ++
1. Explicitly loading a `MemoryVectorStore` seems to fix this.
2. We are now caching the new index and the commit sha at this point into the local filesystem.
3. Instrument how long it takes to load the index

### DIFF
--- a/src/seer/automation/autofix/agent_context.py
+++ b/src/seer/automation/autofix/agent_context.py
@@ -1,4 +1,5 @@
 import hashlib
+import json
 import logging
 import os
 import shutil
@@ -86,6 +87,7 @@ class AgentContext:
             )
         self.tmp_dir = tmp_dir
         self.tmp_repo_path = os.path.join(self.tmp_dir, f"repo")
+        self.cached_commit_json_path = os.path.join("./", "models/autofix_cached_commit.json")
 
         logger.info(f"Using tmp dir {self.tmp_dir}")
 
@@ -194,6 +196,7 @@ class AgentContext:
             logger.error(f"Failed to delete tar file: {e}")
 
     def _load_data_from_github(self):
+        logger.debug(f"Loading data from github for {self.repo.name} on ref {self.base_sha}")
         self._load_repo_to_tmp_dir()
         documents = SimpleDirectoryReader(
             self.tmp_repo_path, required_exts=[".py"], recursive=True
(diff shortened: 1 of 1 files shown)


src/seer/automation/autofix/agent_context.py
```diff
@@ -229,23 +236,34 @@ def _get_commit_file_diffs(self) -> tuple[list[str], list[str]]:
...
-            raise Exception("No index structures found in storage context")
-        index_struct: IndexDict = index_structs[0]  # type: ignore
-
-        index = VectorStoreIndex(
-            index_struct=index_struct,
-            service_context=service_context,
-            storage_context=storage_context,
-            show_progress=True,
-        )
+        storage_path = os.path.join("./", "models/autofix_storage_context/")
```
trillville, line 241: probably a dumb question but just double checking - what happens when a celery worker is kicked off the very first time? will there be something here to begin with?

jennmueng: this will be pulled from gcs at build time

src/seer/automation/autofix/agent_context.py
```diff
@@ -229,23 +236,34 @@ def _get_commit_file_diffs(self) -> tuple[list[str], list[str]]:
...
+            service_context = ServiceContext.from_defaults(embed_model=self.embed_model)
+            memory_vector_store = MemoryVectorStore().from_persist_dir(storage_path)
+            storage_context = StorageContext.from_defaults(
+                vector_store=memory_vector_store,
+                persist_dir=storage_path,
+            )
+            index_structs = storage_context.index_store.index_structs()
+
+            if len(index_structs) == 0:
+                raise Exception("No index structures found in storage context")
```
trillville, line 256: do we want to raise an exception here or do an alternate workflow

jennmueng: hmm something has gone very wrong if there are no index structs here so I think exception is the right move

src/seer/automation/autofix/agent_context.py
```diff
@@ -229,23 +236,34 @@ def _get_commit_file_diffs(self) -> tuple[list[str], list[str]]:
...
-            show_progress=True,
-        )
+        storage_path = os.path.join("./", "models/autofix_storage_context/")
+
+        with sentry_sdk.start_span(
+            op="seer.automation.autofix.index_loading",
+            description="Loading the vector store index from local filesystem",
+        ) as span:
+            service_context = ServiceContext.from_defaults(embed_model=self.embed_model)
+            memory_vector_store = MemoryVectorStore().from_persist_dir(storage_path)
```
trillville, line 248: is it right for this to reference `storage_path` while below we have
```
memory_vector_store.persist(persist_path=os.path.join(storage_path, "vector_store.json"))
```

trillville (approved):

src/seer/automation/autofix/agent_context.py
jennmueng: yeah persist_dir is the directory and persist_path is the specific file, confusing right? lol

trillville: ahh ok gotcha, just checking

PR #140, opened 2024-01-25
chore: Update local development readme instructions

jennmueng:


trillville (approved):

trillville:
thank you! what about the model artifacts? should we add something about loading those locally as well

jennmueng:
> thank you! what about the model artifacts? should we add something about loading those locally as well

Good call

PR #144, opened 2024-01-28
chore: Add sentry sdk to celery worker

jennmueng:
Weren't getting any sentry errors even though there were errors within the worker. Uses same DSN as the normal seer.


trillville (approved):
ahh nice, i was wondering about this

PR #146, opened 2024-01-28
fix(autofix): better handle when there are no changes

jennmueng:
### Problem
Fixes [#TIMESERIES-ANALYSIS-SERVICE-1A](https://sentry.sentry.io/issues/4921956433/), [#TIMESERIES-ANALYSIS-SERVICE-1B](https://sentry.sentry.io/issues/4921956481/), [TIMESERIES-ANALYSIS-SERVICE-1C](https://sentry.sentry.io/issues/4923205088/)

This is because a pull request was attempted to be created when there were no file changes

### Fix
1. Gracefully return `None` if there are no file changes
2. Throw an error if a branch was created with no commits before creating the PR.


src/seer/automation/autofix/agent_context.py
```diff
@@ -395,6 +395,15 @@ def create_branch_from_changes(
             except Exception as e:
                 logger.error(f"Error committing file change: {e}")
 
+        # Check that the changes were made
+        comparison = self.repo.compare(base_commit_sha, branch_ref.object.sha)
+        if comparison.ahead_by < 1:
+            # Remove the branch if there are no changes
+            self.repo.get_git_ref(branch_ref.ref).delete()
+            raise Exception(
```
corps, line 403: This is a nit and doesn't have to change immediately for PoC, but in general I don't like using Exceptions as control flow.  It makes the side effect behavior of a system very hard to predict or test.

In the future, can we simply return `None` and update the types and callsites?  It's more work, but it is more robust.

src/seer/automation/autofix/autofix.py
```diff
@@ -90,6 +93,10 @@ def run(self) -> AutofixOutput | None:
         return output
 
     def _create_pr(self, combined_output: AutofixAgentsOutput):
+        if len(combined_output.changes) == 0:
+            logger.warning(f"No changes were generated by the coding agent")
```
corps, line 97: This is a good example of what I mean.  Returning None with helpful logging is an easier pattern than using Exceptions for non exceptional cases.

Use an exception because something logical or semantic or IO related has breached our expectations.   Novel cases.  Inputs that are outside of our range to support isn't an exception ,it's expected.

All robust systems must deal in failure of inputs.

corps (approved):

src/seer/automation/autofix/agent_context.py
jennmueng: I'd say this one isn't part of the expected control flow and it's for the case where the previous check of no file changes before this function was called went through but for some reason the files above weren't correctly committed. Would this fit into the case you mentioned in the other comment where it's something that breached our expectations?

corps: it's a boundary condition and I couldn't give you a proof, but I do have an intuition, based on experience from automating git in CI.

Git will do unexpected things at the intersection of its many behaviors, and that itself, I expect.  Remember that git has tons of features that expand the meaning of different signals.  It's also very possible when dealing with a remote system that there could be consistency issues (git makes "strong" guarantees until the implementation does something unexpected).

If I were designing this, I'd definitely put a message into sentry, but I wouldn't throw an exception.  I would try to behave robustly.

corps: That said, I think you know this sytem and flow better than me.

jennmueng: Good food for thought; a log to sentry and none return would be better way to short circuit when theres a failure

PR #147, opened 2024-01-29
fix(autofix): Acquire file locks when embedding repos

jennmueng:
## Problems
1. Turns out the celery process runs as many workers as the number of available CPU cores. This means there will be concurrent processes.
2. The embedding files are not locked
3. We aren't also correctly removing the temp directory if the pipeline short circuited or errored.
4. We were logging debug logs from various libraries as well because of the blanket log level debug setting

## Fixes
1. Set max concurrency on the celery process to 4 workers
2. Lock the embedding files throughout the whole data loading process.
    - The files will be locked for the entire process of loading data; that means only one worker at a time can go through the data loading process which is a bottleneck. However, this should be a temporary compromise during PoC stage to ensure that only the latest commit's embeddings are stored.
3. Remove the temp dir at the end of the data loading step
5. Remove debug log level on the celery process as that was enabling debug logs for libraries as well; instead we just need debug logging on the autofix logger.

### DIFF
--- a/celeryworker.sh
+++ b/celeryworker.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 if [ "$CELERY_WORKER_ENABLE" = "true" ]; then
-    exec celery -A src.celery_app.tasks worker --loglevel=debug
+    exec celery -A src.celery_app.tasks worker --loglevel=debug -c 4 # 4 workers for now because we don't acquire locks on the embedding files
 else
     echo "Celery worker is disabled via environment variable CELERY_WORKER_ENABLE"
     exit 0
--- a/celeryworker.sh
+++ b/celeryworker.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 if [ "$CELERY_WORKER_ENABLE" = "true" ]; then
-    exec celery -A src.celery_app.tasks worker --loglevel=debug -c 4 # 4 workers for now because we don't acquire locks on the embedding files
+    exec celery -A src.celery_app.tasks worker --loglevel=info -c 4 # 4 workers for now
 else
     echo "Celery worker is disabled via environment variable CELERY_WORKER_ENABLE"
     exit 0
--- a/src/seer/automation/autofix/agent_context.py
+++ b/src/seer/automation/autofix/agent_context.py
@@ -5,10 +5,9 @@ import os
 import shutil
 import tarfile
 import tempfile
-import time
+from random import randint
 from typing import List
 
-import joblib
 import requests
 import sentry_sdk
 import torch
@@ -82,7 +81,8 @@ class AgentContext:
 
         if tmp_dir is None:
             tmp_dir = os.path.join(
-                tempfile.gettempdir(), f"{repo_owner}-{repo_name}_{self.base_sha}"
+                tempfile.gettempdir(),
+                f"{repo_owner}-{repo_name}_{self.base_sha}_{randint(0, 100000)}",
             )
         self.tmp_dir = tmp_dir
         self.tmp_repo_path = os.path.join(self.tmp_dir, f"repo")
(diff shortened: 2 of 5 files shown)


src/seer/automation/autofix/agent_context.py
```diff
@@ -83,7 +81,8 @@ def __init__(
 
         if tmp_dir is None:
             tmp_dir = os.path.join(
-                tempfile.gettempdir(), f"{repo_owner}-{repo_name}_{self.base_sha}"
+                tempfile.gettempdir(),
+                f"{repo_owner}-{repo_name}_{self.base_sha}_{randint(0, 100000)}",
```
jennmueng, line 85: Need these temp directories to be unique, as multiple workers could be embedding the same commit sha. They'll be deleted at the end of data loading anyways

src/seer/automation/autofix/agent_context.py
```diff
@@ -234,78 +229,104 @@ def _get_commit_file_diffs(self) -> tuple[list[str], list[str]]:
...
-            self.index.delete(document.get_doc_id())
+                commit_is_newer = self.repo.compare(cached_commit_sha, self.base_sha).ahead_by > 0
 
-        with sentry_sdk.start_span(
-            op="seer.automation.autofix.indexing",
-            description="Indexing the diff between the cached commit and the requested commit",
-        ) as span:
-            new_nodes = self._documents_to_nodes(documents_to_update)
-            self.index.insert_nodes(new_nodes)
+                if commit_is_newer:
```
jennmueng, line 310: Now we will only save if the commit we've embedded is newer

src/seer/automation/autofix/agent_context.py
```diff
@@ -345,8 +366,11 @@ def _update_document(self, file_path: str, contents: str | None):
                 self.nodes.extend(new_nodes)
 
     def cleanup(self):
-        shutil.rmtree(self.tmp_dir)
-        logger.info(f"Deleted tmp dir {self.tmp_dir}")
+        if os.path.exists(self.tmp_dir):
+            shutil.rmtree(self.tmp_dir)
+            logger.info(f"Deleted tmp dir {self.tmp_dir}")
+        else:
+            logger.info(f"Tmp dir {self.tmp_dir} already removed")
```
jennmueng, line 373: Make it ok if the directory was already removed.

src/seer/automation/autofix/agent_context.py
corps: Oh by the way, there is an even easier way to do this. https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile

This will ensure better clean up,  it uses system entropy, and you can still control file name template.

corps (approved):

celeryworker.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 if [ "$CELERY_WORKER_ENABLE" = "true" ]; then
-    exec celery -A src.celery_app.tasks worker --loglevel=debug
+    exec celery -A src.celery_app.tasks worker --loglevel=info -c 4 # 4 workers for now
```
trillville, line 4: Just sanity checking - do we want to have more than 1 concurrent worker? logically it makes sense, as everything downstream from the embedding indexing stuff is I/O bound (so theoretically we can easily support multiple concurrent requests). Just double checking my logic here.

If thats the case, then I think there's a strong argument to be made to completely isolating the steps that are CPU-intensive from everything else (which is I/O) intensive. 

E.g. logically we only want 1 (powerful) machine doing embedding/indexing stuff per repo at a given point in time, while we want many more (weak) workers available to do the function calling logic

src/seer/automation/autofix/agent_context.py
trillville: should we just use https://docs.python.org/3/library/uuid.html here?

src/seer/automation/autofix/agent_context.py
```diff
@@ -234,78 +229,104 @@ def _get_commit_file_diffs(self) -> tuple[list[str], list[str]]:
...
-            index_struct: IndexDict = index_structs[0]  # type: ignore
-
-            index = VectorStoreIndex(
-                index_struct=index_struct,
-                service_context=service_context,
-                storage_context=storage_context,
-                show_progress=True,
-            )
+        # The files will be locked for the entire process of loading data; that means only one worker at a time can go through the data loading process which is a bottleneck.
+        # However, this should be a temporary compromise during PoC stage to ensure that only the latest commit's embeddings are stored.
```
trillville, line 233: 👍 lets make sure we have an efficient workflow planned here for next iteration

src/seer/automation/autofix/agent_context.py
jennmueng: looks like https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp is actually what we're looking for

src/seer/automation/autofix/agent_context.py
```diff
@@ -234,78 +229,104 @@ def _get_commit_file_diffs(self) -> tuple[list[str], list[str]]:
...
-        logger.debug(
-            f"Updating index with {len(changed_files)} changed files and {len(deleted_files)} deleted files"
-        )
+                for document in documents_to_update + documents_to_delete:
+                    self.index.delete(document.get_doc_id())
 
-        self.index = index
-        self.documents = documents
-        self.nodes = nodes
+                with sentry_sdk.start_span(
```
trillville, line 294: love the instrumentation! 👏

trillville (approved):
👏

celeryworker.sh
jennmueng: This PR would make sure that only a single worker is working on the CPU-intensive embeddings at a time due to the file lock logic. But yeah I agree that for the target solution it can make sense to completely delegate the compute-intensive task to another machine while the rest of the steps I/O intensive can easily be handled by less powerful machines.

This means something along the lines of – we completely delegate the context computing/embedding tasks to a separate more powerful machine, store the results of that, then run the I/O intensive agent cycles on less powerful machines.

corps: That is true, until it's not.  CPU isn't the only resource to bind though.  Consistency of behavior also matters.

If some interactions, having consistency (that is a synchronous relationship between input and output) will matter, esp wrt to interacting with a shared resource (git).  You can add locks to manage this, but there will be a scale where the lock management is more expensive than simply forcing interactions into a synchronous behavior.

It's another fun paradox of optimization at the limit 🤡

PR #149, opened 2024-01-29
fix(autofix): Support reverse git diffs

jennmueng:
If the second commit in a git diff is in the past it returns an empty diff; the functionality we need is essentially to be able to turn the repo from one state to another. To accomplish this we implement an arbitrary "reverse diff" where the added files are now removed files, and the modified/deleted files become the changed files.

This PR also includes improvements to logging.

Tested locally

### DIFF
--- a/src/seer/automation/autofix/agent_context.py
+++ b/src/seer/automation/autofix/agent_context.py
@@ -122,8 +122,6 @@ class AgentContext:
         nodes: List[BaseNode] = []
 
         for language, language_docs in documents_by_language.items():
-            logger.debug(f"{language}: {len(language_docs)}")
-
             splitter = CodeSplitter(
                 language=language,
                 chunk_lines=16,  # lines per chunk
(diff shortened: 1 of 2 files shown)


src/seer/automation/autofix/agent_context.py
```diff
@@ -337,7 +354,7 @@ def _update_document(self, file_path: str, contents: str | None):
         if document:
             # Delete operation
             if contents is None:
-                self.index.delete(document.get_doc_id())
+                self.index.delete_ref_doc(document.get_doc_id())
```
jennmueng, line 357: .delete() was getting deprecation warnings

src/seer/automation/autofix/agent_context.py
```diff
@@ -122,16 +122,20 @@ def _documents_to_nodes(self, documents: list[Document], max_file_size_bytes=102
         nodes: List[BaseNode] = []
 
         for language, language_docs in documents_by_language.items():
-            logger.debug(f"{language}: {len(language_docs)}")
-
             splitter = CodeSplitter(
                 language=language,
                 chunk_lines=16,  # lines per chunk
```
trillville, line 127: apologies since im pretty sure ive already asked about this, but are we confident about this 16 line limitation here? seems pretty restrictive to me intuitively (i noticed the [default](https://docs.llamaindex.ai/en/stable/api/llama_index.node_parser.CodeSplitter.html#llama_index.node_parser.CodeSplitter.chunk_lines) is 40)

trillville (approved):

src/seer/automation/autofix/agent_context.py
jennmueng: yeah we can increase it in a following iteration; originally this was done to reduce the search result size in favor of the llm asking to see the specific document instead

trillville: right right right 👍

PR #150, opened 2024-01-31
feat(ai-autofix): Implement intermediate pipeline with realtime event updates to frontend

jennmueng:
## Complicated diagram that should help explain...
Autofix diagrams

## Frontend view:
Screenshot 2024-01-30 at 1 00 03 PM


src/seer/rpc.py
```diff
@@ -1,12 +1,32 @@
...
 
-class RPCClient:
+
+class RpcClient(ABC):
+    @abstractmethod
+    def call(self, method: str, **kwargs):
+        pass
+
+
+class DummyRpcClient(RpcClient):
```
jennmueng, line 19: Used for local development

src/seer/automation/autofix/repo_client.py
```diff
@@ -0,0 +1,159 @@
...
+    else:
+        app_auth = Auth.AppAuth(app_id, private_key=private_key)  # type: ignore
+        gi = GithubIntegration(auth=app_auth)
+        installation = gi.get_repo_installation(repo_owner, repo_name)
+        github_auth = app_auth.get_installation_auth(installation.id)
+
+    return github_auth
+
+
+class RepoClient:
```
jennmueng, line 39: Mutating the repo separated out from the original agent context class

src/seer/automation/autofix/repo_client.py
```diff
@@ -0,0 +1,159 @@
...
+from seer.automation.autofix.types import FileChange
+from seer.automation.autofix.utils import generate_random_string, sanitize_branch_name
+
+logger = logging.getLogger("autofix")
+
+
+def get_github_auth(repo_owner: str, repo_name: str):
+    app_id = os.environ.get("GITHUB_APP_ID")
+    private_key = os.environ.get("GITHUB_PRIVATE_KEY")
+    github_token = os.environ.get("GITHUB_TOKEN")
```
jennmueng, line 21: for local development or for non-github app use (token will post as the user)

src/seer/automation/agent/singleturn.py
```diff
@@ -0,0 +1,37 @@
+from typing import Any, Callable
+
+from openai import OpenAI
+
+from seer.automation.agent.types import Message, Usage
+
+
+class LlmClient:
```
jennmueng, line 8: Single use LLM util unlike the multi-iteration agents

docker-compose.yml
```diff
@@ -15,8 +15,6 @@ services:
             - .env
         environment:
             - PORT=9091
-            - GITHUB_REPO_NAME=sentry-mirror-suggested-fix
```
jennmueng, line 18: GITHUB_REPO_NAME no longer passed as env variable

src/seer/automation/agent/agent.py
```diff
@@ -73,14 +49,20 @@ def run(self, prompt: str):
         )
 
         logger.debug(f"----[{self.name}] Running Agent----")
-        logger.debug(f"Prompt: {prompt}")
+        logger.debug(f"Previous messages: ")
+        for message in self.memory:
```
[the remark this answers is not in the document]
jennmueng: great idea but seer can't access gcs yet; let's put that on hubert's radar soon

src/seer/automation/agent/agent.py
```diff
@@ -73,14 +49,20 @@ def run(self, prompt: str):
...
+            logger.debug(f"{message.role}: {message.content}")
 
         while (
             self.iterations == 0
             or (
                 not (
                     self.memory[-1].role in ["assistant", "model"]
-                    and self.memory[-1].content is not None
+                    and (
```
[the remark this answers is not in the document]
trillville: might be worth at least a comment detailing the various cases we are checking for for future reference

trillville (commented):
meta comment: you are going to hate me for saying this, but the complexity here is getting to a point where we really need to start adding some unit tests. I wonder if there is also an easy way we can bake in a happy path/simple case integration test, like using a static cached repo from GCS.

jennmueng:
oh no i agree with a happy path test completely, it's getting hard to manage at this point already

src/seer/automation/agent/singleturn.py
```diff
@@ -0,0 +1,37 @@
+from typing import Any, Callable
```
trillville, line 1: question - does this need to be a completely separate class? at a high level it makes more sense to me that both this and gptagent inhereit LlmAgent (with tweaked class names to make roles more clear). Just worried about unnecessary fragmentation of the business logic

trillville: another question is whether this should simply be a special case of gptagent with iterations capped at 1

src/seer/automation/autofix/repo_client.py
```diff
@@ -0,0 +1,160 @@
...
+    def create_pr_from_branch(
+        self, branch: GitRef, title: str, description: str, issue_id: int | str, usage: Usage
+    ):
+        title = f"""🤖 {title}"""
+
+        issue_link = f"https://sentry.io/organizations/sentry/issues/{issue_id}/"
+
+        description = textwrap.dedent(
+            """\
+            👋 Hi there! This PR was automatically generated 🤖
```
trillville, line 140: thought - it might be a good idea to reformat this to be aligned with a more verbose version of the information communicated on the UI. Basically walking them through the plan and what the execution was in the plan.

So it's like:
<high level summary of problem>
<high elvel summary of fix>
<walkthrough the execution steps>

src/seer/automation/autofix/tools.py
```diff
@@ -58,17 +57,20 @@ def get_tools(self):
...
-""",
+                description=textwrap.dedent(
+                    """\
+                    Search for code snippets.
+                    - You can search for code using either a code snippet or the path.
+                    - The codebase is large, so you will need to be very specific with your query.
+                    - If the path contains relative paths such as ../, you will need to remove them.
+                    - If "code" in "file" search does not work, try searching just the code snippet.
+
+                    Example Queries:
```
trillville, line 68: per our conversation, we might want to rethink this prompt format and/pr the query format (or allow for it to choose between semantic search and keyword/exact search). worried that mixing these ideas will muddy the waters a bit WRT retrieval

trillville (approved):
lets goooooo

PR #155, opened 2024-02-04
feat(ai-autofix): Short circuit the diff embedding step with stacktrace files

jennmueng:
Will short circuit and skip codebase index update step if the files in the stacktrace do not match the ones that changed since the last commit diff. This logic is part of the new codebase context overhaul to pgvector we are doing but this is done in advance in anticipation of higher volume on Monday.

Includes a lot of refactoring cleanup too + unit tests for the short circuiting!

Tested e2e with local Sentry App

### DIFF
--- a/src/celery_app/tasks.py
+++ b/src/celery_app/tasks.py
@@ -17,7 +17,7 @@ class TaskStatusRequest(BaseModel):
     task_id: str
 
 
-@celery_app.task(time_limit=60 * 20)  # 20 minutes
+@celery_app.task(time_limit=60 * 60)  # 1 hour task timeout
 def run_autofix(data: dict[str, Any]) -> None:
     base_url = os.environ.get("SENTRY_BASE_URL")
     if not base_url:
--- a/src/celery_app/tasks.py
+++ b/src/celery_app/tasks.py
@@ -7,7 +7,7 @@ from pydantic import BaseModel
 
 from celery_app.app import app as celery_app
 from seer.automation.autofix.autofix import Autofix
-from seer.automation.autofix.types import AutofixRequest
+from seer.automation.autofix.models import AutofixRequest
 from seer.rpc import SentryRpcClient
 
 logger = logging.getLogger("autofix")
--- a/src/seer/automation/autofix/autofix.py
+++ b/src/seer/automation/autofix/autofix.py
@@ -56,7 +56,8 @@ class Autofix:
         try:
             logger.info(f"Beginning autofix for issue {self.request.issue.id}")
 
-            stacktrace = self.request.issue.events[-1].get_stacktrace()
+            events = self.request.issue.events
+            stacktrace = events[-1].get_stacktrace() if events else None
             if not stacktrace:
                 logger.warning(f"No stacktrace found for issue {self.request.issue.id}")
 
--- a/src/seer/automation/autofix/autofix.py
+++ b/src/seer/automation/autofix/autofix.py
@@ -9,8 +9,8 @@ from llama_index.schema import MetadataMode, NodeWithScore
 from seer.automation.agent.agent import GptAgent
 from seer.automation.agent.singleturn import LlmClient
 from seer.automation.agent.types import Message, Usage
+from seer.automation.autofix.autofix_context import AutofixContext
 from seer.automation.autofix.codebase_context import CodebaseContext
-from seer.automation.autofix.context_manager import ContextManager
 from seer.automation.autofix.event_manager import AutofixEventManager
 from seer.automation.autofix.models import (
     AutofixOutput,
@@ -48,7 +48,7 @@ class Autofix:
         self.request = request
         self.usage = Usage()
         self.event_manager = AutofixEventManager(rpc_client, self.request.issue.id)
-        self.context_manager = ContextManager(
+        self.autofix_context = AutofixContext(
             RepoClient("getsentry", "sentry"), self.request.base_commit_sha
         )
 
@@ -87,20 +87,20 @@ class Autofix:
                 return
 
             try:
-                self.context_manager.load_codebase()
+                self.autofix_context.load_codebase()
 
                 # Below is the short circuit logic to skip embedding the codebase context if the stacktrace
                 # does not contain any files that need to be re-indexed
-                needs_indexing = self.context_manager.diff_contains_stacktrace_files(
+                needs_indexing = self.autofix_context.diff_contains_stacktrace_files(
                     self.stacktrace
                 )
                 if needs_indexing:
                     assert (
-                        self.context_manager.codebase_context is not None
+                        self.autofix_context.codebase_context is not None
                     ), "Codebase context is not loaded"
 
                     logger.debug(f"Updating codebase index")
-                    self.context_manager.codebase_context.update_codebase_index()
+                    self.autofix_context.codebase_context.update_codebase_index()
                 else:
                     logger.debug(f"Codebase does not need to be indexed")
 
@@ -311,10 +311,10 @@ class Autofix:
         ), "PlanningInput requires at least one of the fields: 'message', 'previous_output', or 'problem'."
 
         assert (
-            self.context_manager.codebase_context is not None
+            self.autofix_context.codebase_context is not None
         ), "Planning agent needs a codebase context to run."
 
-        planning_agent_tools = BaseTools(self.context_manager.codebase_context)
+        planning_agent_tools = BaseTools(self.autofix_context.codebase_context)
 
         planning_agent = GptAgent(
             name="planner",
@@ -395,13 +395,13 @@ class Autofix:
         file_changes: list[FileChange] = [],
     ):
         assert (
-            self.context_manager.codebase_context is not None
+            self.autofix_context.codebase_context is not None
         ), "Execution agent needs a codebase context to run."
 
         # TODO: make this more robust
         try:
             context_dump = self._get_plan_step_context(
-                plan_item, self.context_manager.codebase_context
+                plan_item, self.autofix_context.codebase_context
             )
         except Exception as e:
             logger.error(f"Failed to get context for plan item: {e}")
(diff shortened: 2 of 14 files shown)


src/seer/automation/autofix/models.py
```diff
@@ -67,30 +67,53 @@ class PlanningInput(BaseModel):
     problem: Optional[ProblemDiscoveryOutput] = None
```
jennmueng, line 67: Renamed this file from `types.py` to `models.py` given that it contains a couple utils within the pydantic models.

src/seer/automation/autofix/repo_client.py
```diff
@@ -52,6 +54,45 @@ def __init__(
         self.github = Github(auth=get_github_auth(repo_owner, repo_name))
         self.repo = self.github.get_repo(repo_owner + "/" + repo_name)
 
+        self.repo_owner = repo_owner
+        self.repo_name = repo_name
+
+    def get_commit_file_diffs(self, prev_sha: str, next_sha: str) -> tuple[list[str], list[str]]:
```
jennmueng, line 60: Moved this into `RepoClient` from `CodebaseContext`.

Interactions with the repository should all occur here so we can in the future abstract away the repo provider (github, gitlab, bitbucket etc). For now the codebase context still directly interacts with the repo through the field in this class but keeping it for now for sake of not over-optimizing until a stage where our structure is more clear

tests/automation/autofix/test_autofix.py
```diff
@@ -0,0 +1,234 @@
...
+    AutofixRequest,
+    IssueDetails,
+    PlanningOutput,
+    ProblemDiscoveryOutput,
+    ProblemDiscoveryResult,
+    SentryEvent,
+)
+
+
+class TestAutoFixRunFlow(unittest.TestCase):
```
jennmueng, line 15: These unit tests mainly check the short circuit cases are working

src/seer/automation/autofix/autofix.py
```diff
@@ -28,23 +22,48 @@
...
 
     def run(self) -> None:
         try:
             logger.info(f"Beginning autofix for issue {self.request.issue.id}")
 
+            stacktrace = self.request.issue.events[-1].get_stacktrace()
+            if not stacktrace:
+                logger.warning(f"No stacktrace found for issue {self.request.issue.id}")
+
+                self.event_manager.send_no_stacktrace_error()
```
jennmueng, line 63: Screenshot 2024-02-03 at 3 25 26 PM

src/seer/automation/autofix/autofix.py
```diff
@@ -68,9 +87,23 @@ def run(self) -> None:
...
-                codebase_context = CodebaseContext(
-                    "getsentry", "sentry", self.request.base_commit_sha
+                self.context_manager.load_codebase()
+
+                # Below is the short circuit logic to skip embedding the codebase context if the stacktrace
+                # does not contain any files that need to be re-indexed
+                needs_indexing = self.context_manager.diff_contains_stacktrace_files(
+                    self.stacktrace
                 )
+                if needs_indexing:
```
jennmueng, line 97: We now only index when files that are present in the stacktrace change

src/seer/automation/autofix/codebase_context.py
```diff
@@ -255,66 +230,84 @@ def _get_data(self):
...
-                    storage_context.persist(persist_dir=storage_dir)
-                    memory_vector_store.persist(
-                        persist_path=os.path.join(storage_dir, "vector_store.json")
+        if commit_is_newer:
+            # Save only when the commit is newer
+            logger.debug(f"Saving index to storage context")
+            try:
+                with SoftFileLock(
+                    os.path.join(LOCK_PATH), timeout=60 * 15
+                ):  # 15 minute max write lock timeout
```
jennmueng, line 293: Shorter file lock timeout duration as it's not the end of the world if we don't get to save...

src/seer/automation/autofix/codebase_context.py
```diff
@@ -65,6 +61,20 @@ def __init__(self, repo_owner: str, repo_name: str, base_sha: str):
 
         self._get_data()
 
+    @staticmethod
+    def get_cached_commit_sha() -> str | None:
```
jennmueng, line 65: The cached commit logic will soon change with the pgvector overhaul

src/seer/automation/autofix/models.py
```diff
@@ -67,30 +67,53 @@ class PlanningInput(BaseModel):
...
-                    continue
-                frames = item["stacktrace"]["frames"][::-1]
-                for frame in frames[:4]:
-                    stack_str += f" {frame['function']} in {frame['filename']} ({frame['lineNo']}:{frame['colNo']})\n"
-                    for ctx in frame["context"]:
-                        is_suspect_line = ctx[0] == frame["lineNo"]
-                        stack_str += f"{ctx[1]}{'  <--' if is_suspect_line else ''}\n"
-
-                    stack_str += "------\n"
+    def get_stacktrace(self):
```
jennmueng, line 95: Entries can contain other values such as breadcrumbs, etc, this is in anticipation of us supporting other entries other than the stacktrace in the near future.

src/celery_app/tasks.py
```diff
@@ -17,7 +17,7 @@ class TaskStatusRequest(BaseModel):
     task_id: str
 
 
-@celery_app.task(time_limit=60 * 20)  # 20 minutes
+@celery_app.task(time_limit=60 * 60)  # 1 hour task timeout
```
jennmueng, line 20: Increasing the task timeout to 1 hour in anticipation that after the Shipped post it'll be like black friday on the one file-lock-at-a-time logic...

TODO: Update Sentry Application's timeout to 1 hour from 30 mins too

corps: 👍

corps: Later, I think we'll want to look at this more carefully but i understand the context here.

src/seer/automation/autofix/autofix.py
```diff
@@ -28,23 +22,48 @@
...
         self.event_manager = AutofixEventManager(rpc_client, self.request.issue.id)
+        self.context_manager = ContextManager(
+            RepoClient("getsentry", "sentry"), self.request.base_commit_sha
+        )
 
     def run(self) -> None:
         try:
             logger.info(f"Beginning autofix for issue {self.request.issue.id}")
 
+            stacktrace = self.request.issue.events[-1].get_stacktrace()
```
corps, line 59: nit: check the length of events as well here =>

```
events = self.request.issue.events
stacktrace = events[-1].get_stacktrace() if events else None
```

src/celery_app/tasks.py
jennmueng: will completely nuke this with the upcoming in-house codebase index solution backed by pgvector coming by EOW

src/seer/automation/autofix/autofix.py
```diff
@@ -68,9 +87,23 @@ def run(self) -> None:
...
+                self.context_manager.load_codebase()
+
+                # Below is the short circuit logic to skip embedding the codebase context if the stacktrace
+                # does not contain any files that need to be re-indexed
+                needs_indexing = self.context_manager.diff_contains_stacktrace_files(
+                    self.stacktrace
                 )
+                if needs_indexing:
+                    assert (
+                        self.context_manager.codebase_context is not None
```
corps: Is there a way that `load_codebase` can return a type that already has a non optional type guarantee and prevent this entire class of failure via the type system?

Also I don't like asserts -- when they fail, they don't tell us much of *why*.  No semantics about why you're confident this won't happen, no hypothesis on how this might happen, is encoded.

At the least, throw an exception that states something about what you think went wrong, and if you can't, restructure.

src/seer/automation/autofix/codebase_context.py
```diff
@@ -47,15 +44,14 @@ class CodebaseContext:
...
-    def __init__(self, repo_owner: str, repo_name: str, base_sha: str):
-        self.github = Github(auth=get_github_auth(repo_owner, repo_name))
-        self.repo = self.github.get_repo(repo_owner + "/" + repo_name)
-
+    def __init__(self, repo_client: RepoClient, base_sha: str):
+        self.repo_client = repo_client
         self.base_sha = base_sha
 
-        self.tmp_dir = tempfile.mkdtemp(prefix=f"{repo_owner}-{repo_name}_{self.base_sha}")
+        self.tmp_dir = tempfile.mkdtemp(
```
corps, line 51: is there any reason for us to *do* the side effect here?  It feels like this should be something provided *within* the context of doing work, not a constructor argument.  

Perhaps these should be arguments *to* the CodebaseContext, which are provided by a `with tempfile.mkstemp`.

Basically, I think ephemeral filesystem side effects are a bad idea inside of a constructor.

src/seer/automation/autofix/codebase_context.py
corps: Hold on, I think this is a little briddle, let's try this:

```
try:
  with open(CACHED_COMMIT_JSON_PATH, "r") as file:
        cached_commit_data = json.load(file)
        return str(cached_commit_data.get("sha"))
except IOError as e:
   logger.warn("IOError reading cached commit: %s", e)
   return None
except TypeError as e:
   logger.warn("Unexpected type error parsing commit json data: %s", e)
```

src/seer/automation/autofix/codebase_context.py
```diff
@@ -65,6 +61,20 @@ def __init__(self, repo_owner: str, repo_name: str, base_sha: str):
...
+        cached_commit_sha = None
+        with open(CACHED_COMMIT_JSON_PATH, "r") as file:
+            cached_commit_data = json.load(file)
+            cached_commit_sha = cached_commit_data.get("sha")
+
+        return cached_commit_sha
+
+    @staticmethod
+    def set_cached_commit_sha(sha: str):
+        with open(CACHED_COMMIT_JSON_PATH, "w") as sha_file:
```
corps, line 75: Same -- let's catch the IOError and warn without completely failing.
Reason is that there *can* be intermittent errors that we don't want the whole system to collapse under.  Let's send up a metric to sentry but keep processing.

PR #163, opened 2024-02-07
feat(grouping): enable multi-project grouping support

trillville:
- expect additional `project_id` parameter
- use `project_id` to identify which embeddings cache to reference
- backend model changed to https://huggingface.co/jinaai/jina-embeddings-v2-base-code

updated code for indexing [here](https://github.com/getsentry/ml-models/blob/main/grouping/grouping_artifacts.ipynb)

note - there are some schema changes unrelated to my PR, which reflect prior changes that @jennmueng made to autofix (the `generate_schemas` script just hadnt been run since then)

### DIFF
--- a/src/seer/grouping/grouping.py
+++ b/src/seer/grouping/grouping.py
@@ -1,16 +1,17 @@
 import difflib
 import pickle
-from typing import Any, List, Optional
+from typing import Any, Dict, List, Optional
 
 import faiss  # type: ignore
 import numpy as np
 import pandas as pd
-from pydantic import BaseModel, ValidationError, ValidationInfo, field_validator, validator
+from pydantic import BaseModel, ValidationInfo, field_validator, validator
 from sentence_transformers import SentenceTransformer
 
 
 class GroupingRequest(BaseModel):
     group_id: int
+    project_id: int
     stacktrace: str
     message: str
     k: int = 1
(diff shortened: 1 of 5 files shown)


jangjodi (approved):
lgtm, though I don't have a lot of context about creating indices

src/seer/grouping/grouping.py
```diff
@@ -60,24 +62,47 @@ class GroupingLookup:
...
         """
-        self.model = SentenceTransformer(model_path)
+        self.model = SentenceTransformer(model_path, trust_remote_code=True)
         with open(data_path, "rb") as file:
             self.data = pickle.load(file)
-        embeddings = np.vstack(self.data["embeddings"].values).astype("float32")
-        faiss.normalize_L2(embeddings)
-        self.index = faiss.IndexFlatIP(embeddings.shape[1])
-        self.index.add(embeddings)
+        self.indexes = self.create_indexes()
```
jennmueng, line 75: this looks like it'll take a while to initialize on the first call to the endpoint before it's cached by `functools.cache`, are we gonna be dropping the first call?

jennmueng (approved):
one question but lgtm otherwise

src/seer/grouping/grouping.py
trillville: good call, i should pull this logic into that

trillville: FWIW the indexing is actually very fast, but still makes sense

PR #179, opened 2024-02-09
Adds database to the app

corps:
Related: https://github.com/getsentry/ops/pull/9351

@hubertdeng123 Is there a safe way to ensure that this + the new gocd pipeline step actually works without firing off?  I wonder how hard a canary for seer would be.


Makefile
```diff
@@ -14,39 +14,42 @@ help:
...
 	  -v $(tmpdir):/sentry-data-schemas:ro \
 	  -v $$(pwd)/src/:/src:ro \
 	  tufin/oasdiff breaking /sentry-data-schemas/seer/seer_api.json /src/seer/schemas/seer_api.json
+
+.PHONY: migration
+migration: .env
+	docker-compose run app flask db migrate -m 'Migration'
+
+.env:
+	cp .env.example .env
```
jennmueng, line 55: .env.example would be empty unless you set the secrets, do we want to be copying the file without values set over?

src/migrations/env.py
```diff
@@ -0,0 +1,106 @@
...
+logger = logging.getLogger("alembic.env")
+
+
+def get_engine():
+    try:
+        # this works with Flask-SQLAlchemy<3 and Alchemical
+        return current_app.extensions["migrate"].db.get_engine()
+    except (TypeError, AttributeError):
+        # this works with Flask-SQLAlchemy>=3
+        return current_app.extensions["migrate"].db.engine
```
jennmueng, line 24: don't we have 3.1.1 set in the requirements.txt?

jennmueng (approved):
some comments, lgtm otherwise! Thanks for setting this up @corps!

Makefile
corps: For CI, for instance, yes.  `.env.example` is a good place to have 'defaults' given that CI will depend on the docker-compose file for running moving forward.

That said, I'm curious about options for letting users know that they should be editing this file robustly.  Like a big echo?  Mainly I want to keep CI simple (and the easy case of running tests simple).

src/migrations/env.py
corps: We do, this file is actually auto generated from the `flask db` command, so I can likely clean this stuff up.

Makefile
jennmueng: Most of the secrets are fine to get away without in CI so it should work even with them not set in this file.

We should probably add warnings when running seer without required env variables which should be more robust than just an echo

corps: The issue isn't about running seer, it's about the `docker-compose.yml` configuration that requires the `.env` file to exist, regardless of wether it is used (which it isn't in CI, but the file still must exist).

PR #188, opened 2024-02-14
fix bugs and update jia model load

trillville:
- fix bug allowing indices from outside of a project to be referenced
- switch model serialization/loading to properly load all files needed for inference


jennmueng (approved):
lgtm

PR #199, opened 2024-02-19
fix(ai-autofix): Make enum json serializable

jennmueng:
Fixes [TIMESERIES-ANALYSIS-SERVICE-29](https://sentry.sentry.io/issues/4991819556/)

### DIFF
--- a/src/seer/rpc.py
+++ b/src/seer/rpc.py
@@ -1,11 +1,12 @@
 import hashlib
 import hmac
-import json
 import logging
 import os
 from abc import ABC, abstractmethod
 
 import requests
+
+from seer.utils import json_dumps
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


src/seer/rpc.py
```diff
@@ -17,12 +18,17 @@ def call(self, method: str, **kwargs):
...
 
 class DummyRpcClient(RpcClient):
+    """Dummy RPC client that logs the method and arguments it's called with. Use for dev only."""
+
     def __init__(self, should_log: bool = False):
         self.should_log = should_log
 
     def call(self, method: str, **kwargs):
+        body_dict = {"args": kwargs}
+        json_dump = json_dumps(body_dict, separators=(",", ":"))
```
jennmueng, line 28: Was not getting this error locally because I was using this dummy client. Serializing it here to the logs so it'll match behavior of the real client

corps (approved):

PR #203, opened 2024-02-19
Adds property based testing library

corps:
Adds a property test based framework.

Add tests with type annotations, decorate with `@generate`, then have your test run with random inputs based on the types!  Make assertions into invariants about inputs without having to create fake inputs.

You can guide the generation process with arbitrary functions and with `Examples` annotations as well.

Will be using this in my async PR that follows up.


jennmueng (approved):
oooh love this

PR #213, opened 2024-02-20
feat(ai-autofix): Instrument with Langsmith + refactor llm client

jennmueng:
Instruments ai autofix with langsmith and also refactors the llm agents to use the llm client for calls.

Blocked by
- OPs secret setting: https://getsentry.atlassian.net/browse/OPS-5242
- Other env vars: https://github.com/getsentry/ops/pull/9448


src/seer/automation/autofix/autofix.py
```diff
@@ -390,19 +406,23 @@ def run_planning_agent(self, input: PlanningInput) -> PlanningOutput | None:
...
             return None
 
         return self._parse_planning_output(planning_response)
 
+    @traceable(run_type="retriever", name="Execution Agent Plan Step Context Fetcher")
     def _get_plan_step_context(self, plan_item: PlanStep):
         logger.debug(f"Getting context for plan item: {plan_item}")
 
+        def json_parser(x) -> list[str] | None:
+            return json.loads(x) if x else None
```
corps, line 422: Is it worth catching the parser here and handling it elegantly?

Ideally, if we could "guide" the generation by sampling different completions, validating each if they are valid syntax, and selecting the highest confidence one that doesn't break syntax, we may be able to increase accuracy for a pefromance cost.

corps (approved):

src/seer/automation/autofix/autofix.py
jennmueng: gonna rework the parsing of this next

PR #237, opened 2024-02-22
fix(ai-autofix): Fix tool use issues

jennmueng:
### Problems
- OpenAI Tool use passes the argument as a json string, and because we work with code, code snippets often contain JSON characters that confuse JSON parsers
- Create file and delete file weren't working

### Solution
- A JSON parser that uses tree-sitter-languages to only parse out valid characters
- Better support cases of create and delete file
- Little prompt engineering tweaks to adjust the llm tool use and also optimize the execution.

### DIFF
--- a/src/seer/automation/codebase/repo_client.py
+++ b/src/seer/automation/codebase/repo_client.py
@@ -215,27 +215,43 @@ class RepoClient:
         return ref
 
     def _commit_file_change(self, change: FileChange, branch_ref: str):
-        contents = self.repo.get_contents(change.path, ref=branch_ref)
+        contents = (
+            self.repo.get_contents(change.path, ref=branch_ref)
+            if change.change_type != "create"
+            else None
+        )
 
         if isinstance(contents, list):
-            raise Exception(f"Expected a single ContentFile but got a list for path {change.path}")
+            raise RuntimeError(
+                f"Expected a single ContentFile but got a list for path {change.path}"
+            )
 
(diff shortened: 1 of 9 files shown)


src/seer/automation/utils.py
```diff
@@ -23,3 +25,73 @@ def get_embedding_model():
     model.max_seq_length = 4096
 
     return model
+
+
+def parse_json_with_keys(json_str: str, valid_keys: list[str]) -> dict[str, Any]:
+    """
+    NOTE: This function will only work if the json is not nested, and also doesn't work on arrays.
```
jennmueng, line 32: The llm tool calls will be single-layer objects anyways so we conserve energy by not designing for this

src/seer/automation/codebase/repo_client.py
```diff
@@ -215,27 +215,43 @@ def _create_branch(self, branch_name):
...
+            if contents is None:
+                raise FileNotFoundError(f"File {change.path} does not exist in the repository.")
+
             self.repo.update_file(
                 change.path,
                 change.description or "File change",
-                new_contents,
-                contents.sha,
+                new_contents or "",
+                contents.sha,  # FYI: It wants the sha of the content blob here, not a commit sha.
```
corps, line 254: Correct, to make the operation idempotent, the existing sha is used to ensure that two writes don't compete.

corps (approved):

PR #240, opened 2024-02-22
fix(autofix): Fix execution agent context retriever

jennmueng:
The execution agent context retriever often returned a response that was invalid for the json parser and was extremely flaky.

- Have it return a json array inside a clear xml tag for ease of parsing. This way we specifically are able to extract it and ignore noise the llm says
- Provide multi-shot examples for the prompt for better queries + semantic queries.

### DIFF
--- a/src/seer/automation/autofix/autofix.py
+++ b/src/seer/automation/autofix/autofix.py
@@ -418,8 +418,17 @@ class Autofix:
     def _get_plan_step_context(self, plan_item: PlanStep):
         logger.debug(f"Getting context for plan item: {plan_item}")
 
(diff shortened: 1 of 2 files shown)


src/seer/automation/autofix/autofix.py
```diff
@@ -467,8 +476,6 @@ def run_execution_agent(self, plan_item: PlanStep):
             sentry_sdk.capture_exception(e)
             context_dump = ""
 
-        context_dump = ""
```
jennmueng, line 470: oops, forgot to remove this in a previous pr

corps (approved):

PR #241, opened 2024-02-23
Adding tasks framework for background processing

corps:
Cleaning up mypy and some failing tests (minor).  This is the starting place.

Next step will be integrating, slowly, the existing work of @jennmueng to make use of the async background process.

This framework supports a few things.
1.  Optional durable work queueing.  For work that should be retried indefinitely until it works, but coalesced so that multiple requests to do the work become one unit of that work (indexing).
2.  Ability to schedule work into an async processor from any context.  This can be remote (schedule work to be processed by any worker, durability preferred) or local (schedule work to be processed on this pod, latency preferred).
3.  Ability to even put GPU or CPU work off the celery queue.

Follow up will include using shared memory so that python services don't all load different copies of the same model on a pod.

### DIFF
--- a/src/seer/db.py
+++ b/src/seer/db.py
@@ -1,3 +1,5 @@
+import asyncio
+import contextlib
 import datetime
 from typing import Collection
 
@@ -5,7 +7,7 @@ import sqlalchemy
 from flask_migrate import Migrate
 from flask_sqlalchemy import SQLAlchemy
 from pgvector.sqlalchemy import Vector  # type: ignore
-from sqlalchemy import JSON, DateTime, ForeignKey, Integer, String, delete
+from sqlalchemy import JSON, DateTime, ForeignKey, Integer, String, delete, func, select
 from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column, sessionmaker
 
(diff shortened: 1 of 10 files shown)


src/seer/generator.py
```diff
@@ -391,7 +436,54 @@ def generate_any(context: GeneratorContext) -> typing.Iterator[Any] | None:
     return None
 
 
-def generate(context: GeneratorContext | Any) -> typing.Iterator[Any]:
+def generate_sqlalchemy_instance(context: GeneratorContext) -> typing.Iterator[Any] | None:
```
corps, line 439: Can generate Sql alchemy models for tests now for free, too!  They aren't saved by default, so you still need to save them by hand.

src/seer/db.py
```diff
@@ -42,7 +44,7 @@ class ProcessRequest(Base):
     id: Mapped[int] = mapped_column(Integer, primary_key=True)
     name: Mapped[str] = mapped_column(String(128), index=True, unique=True, nullable=False)
     scheduled_for: Mapped[datetime.datetime] = mapped_column(
-        DateTime, default=datetime.datetime.utcnow, index=True, nullable=False
+        DateTime, default=datetime.datetime(2020, 1, 1), index=True, nullable=False
```
jennmueng, line 47: by default scheduled to 2020? I assume this will cause the process the trigger right away when committed to the db right

jennmueng (approved):

PR #249, opened 2024-02-26
feat(grouping): switch from FAISS to pgvector

trillville:
- replace FAISS with pgvector
- add index for `project_id` and `stacktrace_embeddings`
- switch from `cosine_similarity` to `cosine_distance` (which is used by pgvector)
- Rather than always return k nearest neighbors, return k nearest neighbors **with cosine similarity >= 0.85**


jangjodi:
the logic lgtm, but you'll probably want someone with more experience with migrations to take a look at this too. also let me know before you want to merge this, we'll need to change how this api is called on the sentry side as well

src/migrations/versions/49e9388dfd85_migration.py
```diff
@@ -0,0 +1,55 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "grouping_records",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("group_id", sa.Integer(), nullable=False),
+        sa.Column("project_id", sa.Integer(), nullable=False),
+        sa.Column("message", sa.String(), nullable=False),
+        sa.Column("stacktrace_embedding", pgvector.sqlalchemy.Vector(dim=768), nullable=False),
```
jennmueng, line 26: You'll need to import pgvector in this file manually

jennmueng (approved):
just a couple comments otherwise lgtm

src/seer/grouping/grouping.py
```diff
@@ -51,143 +57,93 @@ class SimilarityResponse(BaseModel):
...
+                    DbGroupingRecord.stacktrace_embedding.cosine_distance(embedding).label(
+                        "distance"
+                    ),
+                )
+                .filter(
+                    DbGroupingRecord.project_id == issue.project_id,
+                    DbGroupingRecord.stacktrace_embedding.cosine_distance(embedding) <= 0.15,
+                    DbGroupingRecord.group_id != issue.group_id,
+                )
+                .order_by(DbGroupingRecord.stacktrace_embedding.nearest(embedding))
```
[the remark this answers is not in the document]
trillville: oh good question, i actually dont know - i just copied that from the pgvector docs - i'll use `.cosine_distance` (even if it does, better to be eexplicit)

src/seer/grouping/grouping.py
```diff
@@ -51,143 +57,93 @@ class SimilarityResponse(BaseModel):
...
             ).ratio()
-            should_group = stacktrace_similarity_score >= issue.threshold
+            should_group = distance <= issue.threshold
 
             similarity_response.responses.append(
                 GroupingResponse(
-                    parent_group_id=group_id,
-                    stacktrace_similarity=stacktrace_similarity_score,
+                    parent_group_id=record.group_id,
+                    stacktrace_similarity=distance,
```
jangjodi, line 127: should we rename this to distance?

src/seer/grouping/grouping.py
```diff
@@ -28,21 +29,28 @@ def check_field_is_not_empty(cls, v, info: ValidationInfo):
...
+            group_id=self.group_id,
+            project_id=self.project_id,
+            message=self.message,
+            stacktrace_embedding=self.stacktrace_embedding,
+        )
+
+    class Config:
+        arbitrary_types_allowed = True
+        json_encoders = {
+            np.ndarray: lambda x: x.tolist()  # Convert ndarray to list for serialization
```
corps, line 46: 👍 
In the future, I may want to create a common subclass of `BaseModel` that has this config by default, so we don't have to copy this multiple places, but not necessary for now.

src/seer/grouping/grouping.py
```diff
@@ -28,21 +29,28 @@ def check_field_is_not_empty(cls, v, info: ValidationInfo):
...
+        }
 
 
 class GroupingResponse(BaseModel):
     parent_group_id: Optional[int]
-    stacktrace_similarity: float
-    message_similarity: float
+    stacktrace_distance: float
```
corps, line 52: In the future, we'll need to be mindful of backwards compatibility.  Once there are live consumers of our data models via API, we cannot simply remove keys from responses.  We'll need a multi-step migration where

1.  We continue to support old keys, while adding new keys.
2.  Change sentry side to consume new keys
3.  Remove old keys.

This is totally fine for now, but we should be mindful of this in the future to prevent an accidental outage.

src/seer/grouping/grouping.py
```diff
@@ -51,143 +59,135 @@ class SimilarityResponse(BaseModel):
...
-            data_path (str): Path to the preprocessed data with stacktrace embeddings.
+        :param model_path: Path to the sentence transformer model.
         """
-        self.model = SentenceTransformer(model_path, trust_remote_code=True)
-        with open(data_path, "rb") as file:
-            self.data = pickle.load(file)
-        self.indexes = self.create_indexes()
-
-    def create_indexes(self) -> Dict[int, faiss.IndexFlatIP]:
+        self.model = SentenceTransformer(
```
corps, line 75: I'd love to move all model sinto the `inference_models` file behind a @functools.cache like we are doing with the others.

The reason is that I want to use shared memory spaces so that multiple python processes can share these once they are loaded.  It's not a big deal now, but once we have say, grouping in a celery job, we might accidentally load the model in two processes on the same pod.  That's wasted memory that could be spent making better doing heavier computation! :P

corps (approved):

src/seer/grouping/grouping.py
trillville: hm isnt that the current behavior? from what i can see the loading/caching behavior is identical between the models used for severity and grouping

corps: oh okay, I'm focusing on the SentenceTRansformer part and missed bigger picture.

PR #263, opened 2024-02-29
fix(ai-autofix): Fix stacktrace to_str logic

jennmueng:
Forgot to reverse the stacktrace before truncating it.....we were using the first 4 most irrelevant frames the whole time...

- Correctly reverses
- Default truncates to 16


trillville (approved):
🚢

PR #270, opened 2024-03-01
build: Use cuda-supported docker image

jennmueng:
Running `make dev-cuda` will run docker-compose with gpus enabled

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN chmod +x ./celeryworker.sh
 
 # Install dependencies
 RUN pip install --upgrade pip==23.0.1
-RUN pip install -r requirements.txt --ignore-installed blinker
+RUN pip install -r requirements.txt
 
 # Then install psycopg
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -26,6 +36,9 @@ RUN chmod +x ./celeryworker.sh
 RUN pip install --upgrade pip==23.0.1
 RUN pip install -r requirements.txt
 
+# Then install psycopg
+RUN pip install psycopg[binary]
```
corps, line 40: ?  Is there a reason we put this here and not in requirements.txt?  My concern is that pip install raw here will not lock down versions -- we'll get an incidental version upgrade during image build and deploy, which could be problematic.

corps:
Do we need to change our ops k8s to enable this as well?

Dockerfile
jennmueng: Was having issues earlier without this, let me remove it I kinda think it's not needed

corps (approved):

PR #271, opened 2024-03-01
chore: add sentry logs to json_api

jennmueng:
Log when we get a bad request


src/seer/json_api.py
```diff
@@ -29,11 +30,13 @@ def decorator(implementation: _F) -> _F:
         def wrapper() -> Any:
             data = request.get_json()
             if not isinstance(data, dict):
+                sentry_sdk.capture_message(f"Data is not an object: {data}")
```
corps, line 33: Hmm, let's not put the data into the message itself here, how about just the `type(data)`?  A simple bug on the sentry side could cause very strange inputs and spam us with weird messages for errors.

corps (approved):

PR #276, opened 2024-03-03
ref(ai-autofix): Better design for document chunk models

jennmueng:
Attempts to fix [TIMESERIES-ANALYSIS-SERVICE-2S](https://sentry.sentry.io/issues/5029600041/)

Refactors the document chunk models to not need a `repo_id` in the base model (now renamed `BaseDocumentChunk`) while ones with embeddings are now `EmbeddedDocumentChunk` and the one that has been retrieved from db is `StoredDocumentChunk`.

This way it is clearer that you're working with `StoredDocumentChunk` when retrieving from the database and the former two models are transitional and only used when processing chunks.

### DIFF
--- a/src/seer/automation/codebase/codebase_index.py
+++ b/src/seer/automation/codebase/codebase_index.py
@@ -2,16 +2,16 @@ import logging
 import uuid
 
 import numpy as np
-from langsmith import RunTree, traceable
+from langsmith import traceable
 from tqdm import tqdm
 
 from seer.automation.autofix.models import FileChange, RepoDefinition, Stacktrace
 from seer.automation.codebase.models import (
+    BaseDocumentChunk,
     Document,
-    DocumentChunk,
-    DocumentChunkWithEmbedding,
-    DocumentChunkWithEmbeddingAndId,
+    EmbeddedDocumentChunk,
     RepositoryInfo,
+    StoredDocumentChunk,
 )
 from seer.automation.codebase.parser import DocumentParser
 from seer.automation.codebase.repo_client import RepoClient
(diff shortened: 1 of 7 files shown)


src/seer/automation/codebase/codebase_index.py
```diff
@@ -124,26 +136,16 @@ def create(cls, organization: int, project: int, repo: RepoDefinition, run_id: u
...
 
-                doc_parser = DocumentParser(get_embedding_model())
-                chunks = doc_parser.process_documents(documents)
-                embedded_chunks = cls._embed_chunks(chunks)
-                logger.debug(f"Processed {len(chunks)} chunks")
-
-                repo_info = RepositoryInfo.from_db(db_repo_info)
-
-                db_chunks = [chunk.to_db_model() for chunk in embedded_chunks]
                 session.add_all(db_chunks)
```
trillville, line 144: i'm wondering if this should be batched? not sure how large of an operation it has to be before we care about that, but i can imagine this potentially being a very large transaction in the case of a large repo

trillville (approved):
LGTM overlal, had one question

src/seer/automation/codebase/codebase_index.py
jennmueng: @corps today said it shouldn't be an issue at all; he worked with millions of inserts before. And it's actually good in the case that 1 fails the whole transaction fails rather than a single batch. We don't want repos with partially missing data

trillville: FWIW i'm pretty sure there is a way to batch the operation while still treating it as an atomic transaction (e.g. if it fails halfway through, everything is rolled back). At the very least it might be worth adding a TODO here to revisit this (also worth noting that each row here is very large relative to a typical pg operation, so it may take significantly fewer rows than usual before its problematic)

PR #277, opened 2024-03-03
chore(autofix): Revert the embedding batch size to 4

jennmueng:
We reverted to 1 when we were cpu/memory constrained, now we aren't

(4 worked well on L4 in my experiments)


trillville (approved):
👍 definitely think we should experiemnt with this when we have a better validation framework setup, but we can revisit that later

PR #283, opened 2024-03-03
fix(autofix): Handle more tool usage edge cases

jennmueng:
Fixes multiple edge cases seen with agent tool use in latest runs:

1. Newlines in strings not being newline characters but \n. This was both in snippet matching and in the snippet replacement.
-> Handled by replacing `\n` matches **outside** of quotes in the value of json strings
2. `...` matching didn't work if the `...` was the only thing in a line as it didn't end with a `... ` as original code looked for.

- Also adds support for our JSON parser to handle other non string values when they are valid. Invalid parsing will still just make everything a string but that is a loss we can take.

Should also have extensive test cases to test a lot of possible edge cases for this fragile parsing stuff!


trillville (approved):
lgtm, love the test coverage

PR #295, opened 2024-03-04
fix(autofix): Batching chunk inserts

jennmueng:
Should fix [TIMESERIES-ANALYSIS-SERVICE-2S](https://sentry.sentry.io/issues/5029600041/)

### DIFF
--- a/src/seer/utils.py
+++ b/src/seer/utils.py
@@ -42,6 +42,7 @@ def json_dumps(data, **kwargs) -> str:
 def batch_save_to_db(session: Session, data: Sequence[DeclarativeBase], batch_size: int = 512):
     """
     Save a list of data to the database in batches. Flushes the session after each batch.
+    NOTE: Needs to be called inside a session/transaction.
     """
     for i in range(0, len(data), batch_size):
         session.bulk_save_objects(data[i : i + batch_size])
--- a/src/seer/utils.py
+++ b/src/seer/utils.py
@@ -2,6 +2,9 @@ import functools
 import json
 import weakref
 from enum import Enum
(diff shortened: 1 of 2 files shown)


src/seer/utils.py
```diff
@@ -2,6 +2,9 @@
 import json
 import weakref
 from enum import Enum
+from typing import Sequence
+
+from sqlalchemy.orm import DeclarativeBase, Session
```
trillville, line 7: shouldnt this be from `seer.db`?

src/seer/utils.py
```diff
@@ -34,3 +37,14 @@ def default(self, obj):
 
 def json_dumps(data, **kwargs) -> str:
     return json.dumps(data, cls=SeerJSONEncoder, **kwargs)
+
+
+def batch_save_to_db(session: Session, data: Sequence[DeclarativeBase], batch_size: int = 512):
+    """
+    Save a list of data to the database in batches. Flushes the session after each batch.
+    """
+    for i in range(0, len(data), batch_size):
```
trillville, line 46: i think you can wrap this entire operation in a transaction like this:

```
with session.begin():
            for i in range(0, len(data), batch_size):
                session.bulk_save_objects(data[i : i + batch_size])

                session.flush()
```

jennmueng: it's supposed to be in the same transaction as the parent not its own

jennmueng: lemme add comment

trillville: ah perfect i see

trillville (approved):

PR #306, opened 2024-03-05
fix(autofix): Fix codebase index updating logic

jennmueng:
Resolves [TIMESERIES-ANALYSIS-SERVICE-35](https://sentry.sentry.io/issues/5037525890/)

Introduces logic to correctly remove chunks that no longer exist in a document, embed only the new chunks, and update the indices of neighboring chunks during the codebase indexing step. All done by chunk hash comparison.

- Adds unit tests to test the update/delete cases
- Adds span instrumentation to chunking and embedding steps in both codebase index creation and update

### DIFF
--- a/src/seer/automation/autofix/autofix.py
+++ b/src/seer/automation/autofix/autofix.py
@@ -131,8 +131,8 @@ class Autofix:
                     else:
                         update_codebase_index.apply_async(
                             (UpdateCodebaseTaskRequest(repo_id=repo_id).model_dump(),),
-                            countdown=3 * 60,
-                        )  # 3 minutes
+                            countdown=10 * 60,
+                        )  # 10 minutes
                         logger.info(f"Codebase indexing scheduled for later")
                 else:
                     logger.debug(f"Codebase is up to date")
--- a/src/seer/automation/codebase/codebase_index.py
+++ b/src/seer/automation/codebase/codebase_index.py
@@ -197,7 +197,7 @@ class CodebaseIndex:
                     .all()
                 )
 
-            db_chunk_hashes = set([db_chunk.hash for db_chunk in existing_chunks])
+            db_chunk_hash_map = {chunk.hash: chunk for chunk in existing_chunks}
             new_chunk_hashes = set([chunk.hash for chunk in chunks])
 
             chunks_that_no_longer_exist: list[DbDocumentChunk] = []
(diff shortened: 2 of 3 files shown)


src/seer/automation/autofix/autofix.py
```diff
@@ -131,8 +131,8 @@ def run(self, run_tree: RunTree):
                     else:
                         update_codebase_index.apply_async(
                             (UpdateCodebaseTaskRequest(repo_id=repo_id).model_dump(),),
-                            countdown=3 * 60,
-                        )  # 3 minutes
+                            countdown=10 * 60,
+                        )  # 10 minutes
```
jennmueng, line 135: This runs in the background anyways, so it punts it back further to not eat up resources of the current run

src/seer/automation/codebase/codebase_index.py
```diff
@@ -122,8 +123,10 @@ def create(cls, organization: int, project: int, repo: RepoDefinition, run_id: u
                 logger.debug(f"  {language}: {len(docs)}")
 
             doc_parser = DocumentParser(get_embedding_model())
-            chunks = doc_parser.process_documents(documents)
-            embedded_chunks = cls._embed_chunks(chunks)
+            with sentry_sdk.start_span(op="seer.automation.codebase.create.process_documents"):
```
trillville, line 126: 👏

src/seer/automation/codebase/codebase_index.py
```diff
@@ -180,15 +183,67 @@ def update(self):
...
-                db_chunks = [
-                    chunk.to_db_model(repo_id=self.repo_info.id) for chunk in embedded_chunks
+                for chunk in chunks_that_no_longer_exist:
+                    session.delete(chunk)
+
+                session.flush()
+
+                # Update the indices of the chunks that already exist
+                for chunk in chunks_to_update:
+                    db_chunk = (
```
trillville, line 228: hmm this seems like a classic n+1 query - can this be done in one query since we know all of the hashes ahead of time?

trillville (approved):

src/seer/automation/codebase/codebase_index.py
corps: Yup, let's restructure.  I have some ideas for this I'll get back to you tomorrow.

jennmueng: oh yeah we already have the `existing_chunks` from above, can just use those

PR #328, opened 2024-03-07
ref(autofix): Refactor all the prompts to be more robust

jennmueng:
Moves instructions/information on the current topic at hand to the user prompt in all the prompts along with minor changes to the prompts that should make the agents a little more robust.

## Planning Agent
- Prevents unit test generation
- More focused, actionable output

## Execution Agent
- Should perform the task in less steps and get to a solution faster

[Example run of all these prompts and tracing in action](https://smith.langchain.com/public/e0bcfaa6-6b69-49bc-a35a-e73a804be1cb/r)

### DIFF
--- a/src/seer/automation/autofix/tools.py
+++ b/src/seer/automation/autofix/tools.py
@@ -11,6 +11,7 @@ from seer.automation.autofix.autofix_context import AutofixContext
 from seer.automation.autofix.models import FileChange
 from seer.automation.autofix.prompts import ExecutionPrompts
 from seer.automation.autofix.utils import find_original_snippet
+from seer.automation.codebase.codebase_index import CodebaseIndex
 
 logger = logging.getLogger("autofix")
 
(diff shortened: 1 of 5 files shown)


src/seer/automation/autofix/tools.py
```diff
@@ -136,13 +136,12 @@ def replace_snippet_with(
...
-                    len(lines), snippet_end_line + self.chunk_padding
-                )
-            ]
-        )
+        chunk_lines = lines[
+            max(0, snippet_start_line - self.chunk_padding) : min(
+                len(lines), snippet_end_line + self.chunk_padding
+            )
+        ]
+        chunk = "\n".join(chunk_lines).strip("\n")
```
jennmueng, line 144: This should strip unnecessary newlines from the start/end of the chunk

trillville (approved):

PR #340, opened 2024-03-07
feat(autofix): Output a diff with autofix result

jennmueng:
Computes a diff of the combined file changes of an autofix run. Does not use the github API, this way we can use this to generate a diff for a set of file changes without needing to interact with github at all.

In conjunction with https://github.com/getsentry/sentry/pull/66522

Screenshot 2024-03-08 at 1 27 29 PM

### DIFF
--- a/src/seer/automation/codebase/codebase_index.py
+++ b/src/seer/automation/codebase/codebase_index.py
@@ -4,12 +4,19 @@ import uuid
 
 import numpy as np
 import sentry_sdk
+import tree_sitter_languages
 from langsmith import traceable
 from sqlalchemy.orm import class_mapper
 from tqdm import tqdm
+from tree_sitter import Tree
 from unidiff import PatchSet
 
 from seer.automation.autofix.models import RepoDefinition, Stacktrace
+from seer.automation.codebase.ast import (
+    extract_declaration,
+    find_first_parent_declaration,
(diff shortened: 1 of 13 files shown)


src/seer/automation/codebase/codebase_index.py
```diff
@@ -465,3 +471,87 @@ def _populate_chunks(self, chunks: list[DbDocumentChunk]) -> list[StoredDocument
...
+                            line_type=line.line_type,
+                        )
+                    )
+                hunks.append(
+                    Hunk(
+                        source_start=hunk.source_start,
+                        source_length=hunk.source_length,
+                        target_start=hunk.target_start,
+                        target_length=hunk.target_length,
+                        section_header=hunk.section_header,
```
malwilley, line 537: I wonder why the section header isn't showing up since this looks correct to me

malwilley: Is it because of the way you created `combined_diff`? I tested with `PatchSet.from_str(git_diff_content)` and that gave me section headers

jennmueng: hmm could also be because unified_diff isn't outputting section headers. will look into this further

trillville (approved):

PR #341, opened 2024-03-07
Add durable processing head to Autofix

corps:
Uses the `ProcessRequest` model and async backend to

1.  Store the request and intent to process with a unique key.
2. Dedupe by that key, while updating the request if parameterizations change
3. Lock processing based on that key.
4. Handle state management by allowing celery jobs to communicate progress back to the async worker
5. Recovers steps state from the source (sentry api).
6. Does not yet establish other state sinks, focused on just having abstraction.

Gotta repair merge issues and test.   There may not be an easier way to make this change.


src/seer/automation/autofix/event_manager.py
```diff
@@ -1,318 +1,274 @@
...
             )
-        )
 
-        self._send_steps_update(AutofixStatus.PROCESSING)
+            cur.status = AutofixStatus.PROCESSING
 
     def send_codebase_indexing_result(
         self, status: Literal[AutofixStatus.COMPLETED, AutofixStatus.ERROR, AutofixStatus.CANCELLED]
     ):
         # Update the status of step 2 to COMPLETED and step 3 to PROCESSING
```
jennmueng, line 198:
```suggestion
```
Comment no longer relevant

src/seer/automation/autofix/event_manager.py
```diff
@@ -1,318 +1,274 @@
...
+        for step in state.steps:
             step.status = AutofixStatus.COMPLETED
 
-    def mark_running_steps_errored(self):
-        for step in self.steps:
+    def mark_running_steps_errored(self, state: AutofixContinuation):
+        for step in state.steps:
             if step.status == AutofixStatus.PROCESSING:
                 step.status = AutofixStatus.ERROR
                 for substep in step.progress:
```
jennmueng, line 274: we could move these two inside `AutofixContinuation`

jennmueng (approved):
lgtm

PR #401, opened 2024-03-13
hotfix(autofix): Force new state for now

jennmueng:


corps (approved):
Don't you already do this on the sentry side?

PR #414, opened 2024-03-13
chore(seer): switch from pre-push to push

trillville:


.pre-commit-config.yaml
```diff
@@ -42,7 +42,7 @@ repos:
       additional_dependencies: [packaging==21.3]
     - id: prevent-push
       name: prevent pushing master
-      stages: [pre-push]
+      stages: [push]
```
asottile-sentry, line 45: this is the deprecated name -- `pre-push` is the non-deprecated name

PR #441, opened 2024-03-15
feat(autofix): Get autofix to be more responsive to instructions

jennmueng:
Makes autofix more receptive to instructions from the user. Changes the concept of "additional context" to "instruction"

### DIFF
--- a/src/seer/automation/autofix/models.py
+++ b/src/seer/automation/autofix/models.py
@@ -31,7 +31,7 @@ class StacktraceFrame(BaseModel):
         )
     )
 
-    function: Optional[Annotated[str, Examples(generator.ascii_words)]] = 'unknown_function'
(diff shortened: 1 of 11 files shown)


src/seer/automation/autofix/models.py
```diff
@@ -258,7 +258,7 @@ class AutofixRequest(BaseModel):
     invoking_user: Optional[AutofixUserDetails] = None
 
     base_commit_sha: Optional[Annotated[str, Examples(generator.shas)]] = None
-    additional_context: Optional[str] = None
+    instruction: Optional[str] = Field(default=None, validation_alias="additional_context")
```
jennmueng, line 261: Corresponding PR on the sentry side coming up,  but this will still allow any requests that come in between this being deployed and then to still work.

trillville (approved):

PR #443, opened 2024-03-16
feat(autofix): Support dry runs + evaluation

jennmueng:
Autofix now supports dry runs
- Can update codebase index to a specific git sha
- Can make the codebase index update temporary
- `DummyRpcClient` has a `dry_run` flag

These flags are designed to be set when running autofix from inside a notebook, you can't use them via API request.

Adds evaluation set creation + running notebooks to the `notebooks/` directory


notebooks/run_eval.ipynb
```diff
@@ -0,0 +1,347 @@
+{
```
trillville, line 1: i dont love the idea of putting these notebooks in the repo, but i see the reason why - fine for now but let's revisit whether or not there's a cleaner way of handling this

trillville (approved):

notebooks/run_eval.ipynb
jennmueng: I don't either but as of now can't think of another clean way to get it to run code from inside the repo consistently

PR #461, opened 2024-03-21
Storage abstraction with tests

corps:
Storage abstraction for codebase with tests.  Not fully integrated, but a great place to start.
Adding support for branches shouldn't require too much work.  "Canonical" is just a point of reference, you can add another column and update the existing unique indexes to get there.


src/seer/automation/codebase/models.py
```diff
@@ -13,13 +17,17 @@ class Document(BaseModel):
     language: str
 
 
+lorem_ipsum_parts = "lLorem ipsum dolor sit amet, consectetur adipiscing elit.".split()
+lorem_ipsum = (" ".join(r.choice(lorem_ipsum_parts) for _ in range(r.randint(15, 90))) for r in gen)
```
jennmueng, line 21: I love this so much

jennmueng (approved):
lgtm, I can take it from here to use it in our codebase index class

PR #480, opened 2024-03-26
feat(similarity-embedding): Create bulk insert record endpoint

jangjodi:


src/seer/grouping/grouping.py
```diff
@@ -130,6 +145,16 @@ def encode_text(self, stacktrace: str) -> np.ndarray:
...
 
+    def encode_multiple_texts(self, stacktraces: List[str], batch_size: int = 1000) -> np.array:
+        """
+        Encodes multiple stacktraces in batches using the sentence transformer model.
+
+        :param stacktraces: The list of stacktracse to encode.
+        :param batch_size: The batch size used for the computation.
+        :return: The embeddings of the stacktraces.
+        """
+        return self.model.encode(sentences=stacktraces, batch_size=batch_size)
```
jangjodi, line 156: Attempt at bulk encoding stacktraces

src/seer/grouping/grouping.py
```diff
@@ -197,3 +254,13 @@ def insert_new_grouping_record(self, session, issue: GroupingRequest, embedding:
...
             session.add(new_record)
+
+    def bulk_insert_new_grouping_records(self, records: List[DbGroupingRecord]):
+        """
+        Bulk inserts new GroupingRecord into the database.
+
+        :param records: List of records to be inserted
+        """
+        with Session() as session:
+            session.bulk_save_objects(records)
```
jangjodi, line 265: do you happen to know if/how this might fail?

src/seer/grouping/grouping.py
```diff
@@ -30,6 +30,17 @@ def check_field_is_not_empty(cls, v, info: ValidationInfo):
         return v
 
 
+class CreateGroupingRecordData(BaseModel):
+    group_id: int
```
trillville, line 34: should this be rebased with the PR adding grouphash merged?

src/seer/grouping/grouping.py
trillville: yeah this should work fine for single GPU - we potentially want to modify this to [this](https://www.sbert.net/docs/package_reference/SentenceTransformer.html#sentence_transformers.SentenceTransformer.encode_multi_process) but let me sync with SRE first

src/seer/grouping/grouping.py
trillville: LGTM @corps any thoughts/concerns?

src/seer/app.py
```diff
@@ -77,6 +83,17 @@ def similarity_endpoint(data: GroupingRequest) -> SimilarityResponse:
     return similar_issues
 
 
+@json_api("/v0/issues/similar-issues/grouping-record")
+def similarity_grouping_record_endpoint(
+    data: CreateGroupingRecordsRequest,
+) -> BulkCreateGroupingRecordsResponse:
+    with sentry_sdk.start_span(
+        op="seer.grouping-record", description="grouping record bulk insert"
```
corps, line 91: random thought: should we just put the sentry_sdk.start_span logic into `json_api` and derive an op from the url?  Tiny savings of effort.

Not a PR blocker, just a thought.

src/seer/grouping/grouping.py
corps: The most likely failure would be a timeout if the bulk is too big.  One alternative is to attempt to write in a discretely sized chunks.  It's a halting problem though so there is never any guarantee that you won't hit a timeout, but as long as we have a lever to adjust (batch size) we'd be good.

To do batch processing you'd essentially break your input up into sized chunks do something like this:

```
for chunk in chunks:
  session.bulk_save_objects(chunk)
  session.flush() # writes to network but doesn't commit

session.commit() # still atomic with regards to full set of records
```

corps (approved):
No blockers, just nits.

ram-senth:
I have a high level question and a potential concern. With this current implementation, both the long running bulk insert calls and the shorter online transactional calls are routed to the same system. Should we be concerned about one interfering with other? Another point from Tillman was that this current implementation is done synchronously and can take about 30 seconds for a batch of 1000 events.

src/seer/grouping/grouping.py
ram-senth: I had a discussion with Tillman about this (after he posted his comment). It probably will be better to either make the bulk insert an asynchronous operation or move it our of this service into the component that handles doc indexing.

jangjodi: Thanks for the feedback @ram-senth! Which component handles doc indexing?

jangjodi: Thanks @corps! I'll have a process to break the input into chunks on the sentry side that calls this

ram-senth: @jangjodi I'm setting up a call with Jenn for later today to ramp up on the indexing component. Will add you as optional. If you can join that will be great.

PR #510, opened 2024-03-29
feat(autofix): Lookup org slug using Seer RPC endpoint

ram-senth:


ram-senth:
PR https://github.com/getsentry/sentry/pull/67961 needs to land first.

src/seer/automation/autofix/autofix.py
```diff
@@ -333,6 +340,20 @@ def _create_prs(self, title: str, description: str, steps: list[Any], request: A
 
         return prs
 
+    def _get_slug_id(self, organization_id: int) -> str:
```
jennmueng, line 343:
```suggestion
    def _get_slug_id(self, organization_id: int) -> str | None:
```
Hmm I think we should handle this more gracefully. Instead of returning the string `"UNKNOWN"` let's have it return None and skip linking to the issue in the PR, we can use the logic that checks for the issue short id

ram-senth: Sounds good.

ram-senth: Have made the changes.

README.md
```diff
@@ -76,3 +76,9 @@ If you run into any data issue or connection issue and want to start from scratc
 docker-compose down --volumes
 make update && make dev
 ```
+
+### Reset Mock Projects
+Running the following command in the `sentry` folder will ensure that the pre loaded test projects internal, earth, fire etc are reset to original state.
+```
+sentry cleanup --days 0 && bin/load-mocks
```
jennmueng, line 83: small nit but hmm is it the best idea to put a sentry side dev command on seer's readme? might confuse people especially because we're open source

jennmueng (approved):
Note on the readme addition but otherwise LGTM

README.md
ram-senth: That is a good point. I was adding information based on my on-boarding but that is not universally useful.

ram-senth: Have removed this.

PR #597, opened 2024-04-04
chore: Use all CUDA devices

ram-senth:
Assign one Cuda device to one celery process. Tested on I4 VM. Below are screen shots of memory and GPU usage when I submitted 100 autofix requests to Seer.
>Baseline - Memory and CPU before submitting 100 requests
Before Starting - baseline

>Snapshot 1 during processing
During - 1

>Snapshot 2 during processing
During - 2

>Snapshot 3 during processing (memory drops after completion of a batch)
During - 3 - shows memory going back to baseline

>After 100 requests -  memory back to baseline
After 100 requests - memory back to baseline

### DIFF
--- a/src/seer/automation/utils.py
+++ b/src/seer/automation/utils.py
@@ -1,26 +1,58 @@
 import functools
+import logging
 import os
 from typing import Any
 
+import billiard  # type: ignore[import-untyped]
 import torch
 from sentence_transformers import SentenceTransformer
 
+# ALERT: Using magic number four. This is temporary code that ensures that AutopFix uses all 4
+# cuda devices available. This "4" should match the number of celery sub-processes configured in celeryworker.sh.
+EXPECTED_CUDA_DEVICES = 4
+logger = logging.getLogger("autofix")
 
-def get_torch_device():
+
+def _get_torch_device_name():
+    device: str = "cpu"
     if torch.cuda.is_available():
-        return torch.device("cuda")
+        if torch.cuda.device_count() >= EXPECTED_CUDA_DEVICES:
+            try:
+                index: int = billiard.process.current_process().index
+                if index < 0 or index >= EXPECTED_CUDA_DEVICES:
+                    logger.warn(
+                        f"CUDA device selection: invalid process index {index}. Defaulting to active GPU."
+                    )
+                    device = "cuda"
+                else:
+                    device = f"cuda:{index}"
+            except Exception as e:
+                logger.warn(
+                    "CUDA device selection: unable to look up celery process index. Defaulting to active GPU.",
+                    exc_info=e,
+                )
+                device = "cuda"
+        else:
+            logger.info(
+                f"CUDA device selection: found {torch.cuda.device_count()} CUDA devices which is less than the required {EXPECTED_CUDA_DEVICES}. Defaulting to active GPU"
+            )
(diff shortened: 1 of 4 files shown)


src/seer/automation/utils.py
```diff
@@ -1,26 +1,58 @@
...
 import os
 from typing import Any
 
+import billiard  # type: ignore[import-untyped]
 import torch
 from sentence_transformers import SentenceTransformer
 
+# ALERT: Using magic number four. This is temporary code that ensures that AutopFix uses all 4
+# cuda devices available. This "4" should match the number of celery sub-processes configured in celeryworker.sh.
+EXPECTED_CUDA_DEVICES = 4
```
trillville, line 12: if we _want_ to make this dynamic (maybe its better to just be explicit) we could use `torch.cuda.device_count()`

src/seer/automation/utils.py
```diff
@@ -1,26 +1,58 @@
 import functools
+import logging
 import os
 from typing import Any
 
+import billiard  # type: ignore[import-untyped]
```
trillville, line 6: i'm not familiar with this library - doesn't it need to be added to `requirements.txt`? 

Relatedtly, we might want to consider separating out requirements between autofix, severity, and grouping

trillville (commented):
Overall LGTM as a way of assigning workers to GPUs (though I am not familiar with this `billiard` library). 

Correct me if I'm missing something, but to the extent there is some sort of memory leak/fragmentation occurring, this will not address that root cause, right? Just reduce the spam to one particular GPU when multiple concurrent requests are being made.

src/seer/automation/utils.py
ram-senth: This number needs to match the number of celery processors set in [celeryworker.sh](https://github.com/getsentry/seer/pull/597/files#diff-f22bf07bec9fbc42f06910e933c57c162fbbe995b60db915f5415b7e78047b12). Since I couldn't figure a way to use device_count there, I decided not to use it here as well.

src/seer/automation/utils.py
ram-senth: Per the [project page](https://pypi.org/project/billiard/), Billiard is used in and is a dependency for [Celery](http://celeryproject.org/) and is maintained by the Celery team. So I did not add it to requirements.txt.

ram-senth:
> Overall LGTM as a way of assigning workers to GPUs (though I am not familiar with this `billiard` library).
> 
> Correct me if I'm missing something, but to the extent there is some sort of memory leak/fragmentation occurring, this will not address that root cause, right? Just reduce the spam to one particular GPU when multiple concurrent requests are being made.

Yes that is correct it does not address any memory issue directly. However, testing with this change seems to indicate that we do not have a memory leak anymore (at least not on the VM). Most likely Jenn's theory is right - that the memory issue was a side effect of multiple processes accessing the same model. I think we should deploy this and monitor the behavior.

src/seer/automation/utils.py
trillville: ah got it!

src/seer/automation/utils.py
jennmueng: In the future we can just set an environment variable that's both used in the celery command and here

PR #598, opened 2024-04-05
Models are loaded in a thread

corps:
When starting the web server, load models in a thread rather than blocking, and have the liveness and readiness checks report that back.

1.  When models fail to load, live will return 500 and we'll have the pods restart, along with sentry errors.
2.  When models are slow to reload, live returns 200 but ready immediately returns 503, ensuring k8s does not restart the pods directly, but merely waits until the ready hits 200.
3.  Tested.

### DIFF
--- a/src/seer/inference_models.py
+++ b/src/seer/inference_models.py
@@ -1,6 +1,10 @@
+import contextlib
 import functools
 import os
-from typing import Any, Callable
+import threading
+from typing import Any, Callable, Literal, TypeVar
+
+import sentry_sdk
 
 from seer.grouping.grouping import GroupingLookup
 from seer.severity.severity_inference import SeverityInference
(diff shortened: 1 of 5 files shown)


src/seer/inference_models.py
```diff
@@ -12,29 +16,98 @@ def model_path(subpath: str) -> str:
...
         model_path=model_path("issue_grouping_v0/embeddings"),
         data_path=model_path("issue_grouping_v0/data.pkl"),
     )
 
 
-function_env_config = {
-    "embeddings_model": "SEVERITY_ENABLED",
-    "grouping_lookup": "GROUPING_ENABLED",
-}
+LoadingResult = Literal["pending"] | Literal["loading"] | Literal["done"] | Literal["failed"]
```
jennmueng, line 50:
```suggestion
LoadingResult = Literal["pending", "loading", "done", "failed"]
```
why are the literals separated out?

jennmueng (approved):
lgtm, very minor nit

src/seer/inference_models.py
corps: Oh, dude, I literally didn't know about that syntax!!!

jennmueng: Oh, LOL

src/seer/inference_models.py
```diff
@@ -12,29 +16,98 @@ def model_path(subpath: str) -> str:
...
+
+@deferred_loading("SEVERITY_ENABLED")
 def embeddings_model() -> SeverityInference:
     return SeverityInference(
         model_path("issue_severity_v0/embeddings"), model_path("issue_severity_v0/classifier")
     )
 
 
-@functools.cache
+@deferred_loading("GROUPING_ENABLED")
```
ram-senth, line 42: Very nice. Solves the issue of the model being loaded into the cuda device:0 even for auto fix deployment.

PR #625, opened 2024-04-18
ref(similarity-embeddings): Add group hash column, defer hash column

jangjodi:
Add non-optional `group_hash` column, give it a default value of "00000000000000000000000000000000"
Defer usage of optional `stacktrace_hash` column (to be dropped in a follow up PR)
Remove usage of `stacktrace_hash` column

Return the `parent_group_hash` from` get_nearest_neighbor`
Allow nearest neighbors endpoint to be called without the `group_id`
Remove the token from the nearest neighbors endpoint return (we aren't calling an update endpoint anymore, so we don't need it)


src/seer/grouping/grouping.py
```diff
@@ -145,25 +146,6 @@ def get_nearest_neighbors(self, issue: GroupingRequest) -> SimilarityResponse:
                  stacktrace similarity scores, message similarity scores, and grouping flags.
         """
         with Session() as session:
-            # If an exact match of the stacktrace hash is found, return this record
```
jangjodi, line 148: Remove this logic since we're using the group hash instead of group id

trillville (approved):

src/seer/grouping/grouping.py
```diff
@@ -180,22 +162,24 @@ def get_nearest_neighbors(self, issue: GroupingRequest) -> SimilarityResponse:
                 .filter(
                     DbGroupingRecord.project_id == issue.project_id,
                     DbGroupingRecord.stacktrace_embedding.cosine_distance(embedding) <= 0.15,
-                    DbGroupingRecord.group_id != issue.group_id,
-                    DbGroupingRecord.group_id != None,
+                    or_(
+                        DbGroupingRecord.group_id != issue.group_id,
+                        DbGroupingRecord.group_id == None,
+                    ),
+                    # TODO We will return the group itself as a similar group if it exists in the old table with no group_hash
```
lobsterkatie, line 169:
```suggestion
                    # TODO We will return the group id as a similar group if it exists in the old table with no hash
```

src/seer/grouping/grouping.py
```diff
@@ -204,7 +188,8 @@ def get_nearest_neighbors(self, issue: GroupingRequest) -> SimilarityResponse:
 
             similarity_response.responses.append(
                 GroupingResponse(
-                    parent_group_id=record.group_id,
+                    parent_group_id=record.group_id if hasattr(record, "group_id") else None,
```
lobsterkatie, line 191: So we can't stop sending this back until we remake the database with hash data (using the backfill script), correct?

src/seer/grouping/grouping.py
```diff
@@ -224,22 +209,14 @@ def insert_new_grouping_record(
...
-                group_id=issue.group_id,
+                group_id=issue.group_id if hasattr(issue, "group_id") else None,
                 project_id=issue.project_id,
                 message=issue.message,
                 stacktrace_embedding=embedding,
-                stacktrace_hash=issue.stacktrace_hash,
+                hash=issue.hash,
             ).to_db_model()
             session.add(new_record)
-            session.commit()
```
lobsterkatie, line 242: Does the session automatically get committed somewhere/somehow else? (This may just be my ignorance of flask/sqlalchemy.)

src/seer/schemas/seer_api.json
```diff
@@ -429,6 +440,10 @@
                         ],
                         "title": "Parent Group Id"
                     },
+                    "parent_hash": {
+                        "type": "string",
+                        "title": "Group Hash"
```
lobsterkatie, line 445:
```suggestion
                        "title": "Parent Hash"
```

tests/seer/grouping/test_grouping.py
```diff
@@ -6,59 +6,67 @@
 
 
 class TestGrouping(unittest.TestCase):
-    def test_get_nearest_neighbors_no_hash_has_neighbor(self):
+    def test_get_nearest_neighbors_has_neighbor(self):
         """
-        Tests get_nearest_neighbors when the request has no stacktrace hash and a neighbor exists
-        within the threshold.
+        Tests get_nearest_neighbors when the request has a hash and this hash matches
+        an existing record (ie. the neighbor exists within the threshold)
```
lobsterkatie, line 12:
```suggestion
        Tests get_nearest_neighbors when the request has a hash and group id and the request's data matches
        an existing record (ie. a neighbor exists within the threshold)
```

tests/seer/grouping/test_grouping.py
```diff
@@ -6,59 +6,67 @@
...
+                hash="QYK7aNYNnp5FgSev9Np1soqb1SdtyahD",
             )
             grouping_lookup().insert_new_grouping_record(session, grouping_request, embedding)
             session.commit()
 
         grouping_request = GroupingRequest(
-            group_id=2, project_id=1, stacktrace=b"stacktrace", message="message"
+            group_id=2,
+            project_id=1,
+            stacktrace=b"stacktrace",
```
lobsterkatie, line 29: Why is this stacktrace value bytes? (The previous one is not.)

tests/seer/grouping/test_grouping.py
```diff
@@ -6,59 +6,67 @@
...
             ],
-            token=None,
         )
 
-    def test_get_nearest_neighbors_has_hash_has_neighbor(self):
+    def test_get_nearest_neighbors_has_neighbor_no_group_id(self):
         """
-        Tests get_nearest_neighbors when the request has a stacktrace hash and this hash matches
+        Tests get_nearest_neighbors when the request has a group hash and this hash matches
         an existing record (ie. the neighbor exists within the threshold)
```
lobsterkatie, line 52:
```suggestion
        Tests get_nearest_neighbors when the request has only a hash and the request's data matches
        an existing record (ie. a neighbor exists within the threshold)
```

tests/seer/grouping/test_grouping.py
```diff
@@ -154,14 +120,17 @@ def test_insert_new_grouping_record_group_record_exists(self):
...
-                session, grouping_request, embedding
-            )
+            grouping_lookup().insert_new_grouping_record(session, grouping_request, embedding)
+            session.commit()
             # Re-insert the grouping record
-            existing_id = grouping_lookup().insert_new_grouping_record(
-                session, grouping_request, embedding
+            grouping_lookup().insert_new_grouping_record(session, grouping_request, embedding)
+            session.commit()
+            new_record = (
```
lobsterkatie, line 131:
```suggestion
            matching_records = (
```

lobsterkatie (approved):
A few questions and a few small suggestions/nits, but overall looks good!  🚀

src/seer/grouping/grouping.py
jangjodi: Yes, that's correct

src/seer/grouping/grouping.py
jangjodi: Yes, it's done outside the function [here](https://github.com/getsentry/seer/pull/625/files#diff-1e79eedfebb1556ea67699fd4535d2ab852d264bbf2fc6515e06fdbf357b5d7bR180)

tests/seer/grouping/test_grouping.py
jangjodi: That was a mistake, I've removed it

src/seer/grouping/grouping.py
lobsterkatie: Ah, got it. Didn't see that, thanks!

PR #652, opened 2024-04-25
feat(autofix): Mutex for read/write to codebase namespace storage

jennmueng:
Introduces a mutex stored in postgres for reading/writing to a codebase namespace storage. This is for both filesystem & gcs. This is due to the read/write for chroma db dumps going across multiple files to prevent half the files being from different revisions


src/seer/automation/codebase/namespace.py
```diff
@@ -35,6 +36,8 @@ class CodebaseNamespaceManager:
     client: ChromaClient
     storage_adapter: StorageAdapter
 
+    NAMESPACE_MUTEX_TIMEOUT_MINUTES = 2
```
trillville, line 39: whats the intuition for this timeout?

src/seer/automation/codebase/namespace.py
```diff
@@ -53,6 +56,104 @@ def __init__(
     def __del__(self):
         self.cleanup()
 
+    @staticmethod
+    def get_mutex(namespace_id: int) -> datetime.datetime | None:
```
trillville, line 60: is there a way to make this more granular than namespace_id? e.g. at the file level or something? i can see his being problematic for a large org/project

trillville (approved):
have some questions about this, but not blockers

src/seer/automation/codebase/namespace.py
jennmueng: gcs upload/download times for a 50-500mb file

PR #654, opened 2024-04-29
feat(codebase): Status endpoint for codebase index

jennmueng:
Introduces the endpoint `/v1/automation/codebase/index/status` to get the current status of the codebase. Returns the following statuses:
```py
class CodebaseIndexStatus(str, enum.Enum):
    UP_TO_DATE = "up_to_date"
    INDEXING = "indexing"
    OUT_OF_DATE = "out_of_date"
    NOT_INDEXED = "not_indexed"
```

Now a namespace has the field `status` which can be 
```py
class CodebaseNamespaceStatus(str, enum.Enum):
    CREATED = "created"
    PENDING = "pending"
    UPDATING = "updating"
```
to do this, we change the initial codebase index creation step to create the repo and namespace db rows before the indexing is complete. Because of this in the case of a failure during embedding etc it will need to delete the namespace


jennmueng:
re-enabled the codebase index create & patch tests... for the life of me I cannot get the codebase update tests to not freeze...

tests/automation/codebase/test_codebase_index.py
```diff
@@ -1,688 +1,774 @@
...
+                workspace = CodebaseNamespaceManager.load_workspace(namespace.id)
+                self.assertIsNotNone(workspace)
+                if workspace:
+                    self.assertEqual(workspace.namespace.id, namespace.id)
+
+    @patch("seer.automation.codebase.codebase_index.CodebaseIndex.embed_chunks")
+    @patch("seer.automation.codebase.codebase_index.read_directory")
+    @patch("seer.automation.codebase.codebase_index.RepoClient")
+    @patch("seer.automation.codebase.codebase_index.cleanup_dir")
+    def test_failing_create(
```
jennmueng, line 110: this is the new test case fyi

src/seer/automation/codebase/storage_adapters.py
```diff
@@ -85,6 +89,12 @@ def save_to_storage(self):
 
         return True
 
+    def delete_from_storage(self):
+        storage_path = self.get_storage_location(self.repo_id, self.namespace_slug)
+        shutil.rmtree(storage_path, ignore_errors=True)
```
ram-senth, line 94: I think we should avoid silent failures. ignore_errors should be False? Also, will ignoring errors cause unwanted side effects due to left over files?

ram-senth: Maybe we need to add a check for existence to avoid failure if there is nothing to delete?

src/seer/automation/codebase/models.py
```diff
@@ -176,12 +177,20 @@ def to_repo_definition(self) -> RepoDefinition:
         )
 
 
+class CodebaseNamespaceStatus(str, enum.Enum):
+    CREATED = "created"
+    PENDING = "pending"
+    UPDATING = "updating"
```
ram-senth, line 183: Do we need an error status when indexing fails?

jennmueng: Ideally we just delete it

src/seer/automation/codebase/storage_adapters.py
jennmueng: Hmm in this case that the filesystem storage adapter is only used for local dev for now we can just have a check and surface any errors

src/seer/automation/codebase/models.py
ram-senth: At some point we will have to keep track of repos for which we see consistent failures and avoid re-indexing until is forced, right? One of the items in the list we discussed during standup.

jennmueng: Oh I thought you meant that in context of the evals, yeah it would make sense in production too to keep track...maybe in a follow up PR because by doing so it will greatly expand the scope of code changes needed. Right now we assume if a namespace exists then it's correctly indexed

ram-senth (approved):

PR #656, opened 2024-04-29
feat(autofix): Expand validation dataset

ram-senth:


src/seer/automation/models.py
```diff
@@ -160,7 +160,7 @@ def validate_stacktrace(cls, sentry_stacktrace: SentryStacktrace | Stacktrace):
 
 class EventDetails(BaseModel):
     title: str
-    exceptions: list[ExceptionDetails] = Field(default_factory=list, exclude=True)
+    exceptions: list[ExceptionDetails] = Field(default_factory=list, exclude=False)
```
ram-senth, line 163: @jennmueng mind confirming if this change is ok? I needed this change to store intermediate data while creating dataset.

.gitignore
```diff
@@ -140,3 +140,4 @@ models
 
 # Evaluation Data
 data/
+val_100_data/
```
jennmueng, line 143: let's keep it organized and make it a subfolder inside of `data` instead of another root level folder

src/seer/automation/models.py
jennmueng: hmm this should've been here because it might have not been JSON  serializable but if you see no problems after changing it it should be fine, I don't see it being seralized afaik anywhere anymore

jennmueng (approved):
lgtm, little nit, also you probably want to clear the notebook outputs

ram-senth:
> lgtm, little nit, also you probably want to clear the notebook outputs

I have cleared most of the outputs. Just kept a few that have context about my run.

.gitignore
ram-senth: Have removed this line.

PR #663, opened 2024-05-01
ref(codebase): Split codebase create and indexing

jennmueng:
Splits out the codebase index creation into a separate creation of the db entries and the index/embedding step. This was needed to be done to return the indexing status to Sentry immediately. In the previous version, it would've needed to wait for a celery worker to pick up the job before the indexing status is present.

This actually plays nicely into splitting up logic that doesn't need the GPU power out (creating the db entries) from the work that does, which is sent to the celery worker.

### DIFF
--- a/src/seer/automation/codebase/codebase_index.py
+++ b/src/seer/automation/codebase/codebase_index.py
@@ -41,7 +41,15 @@ from seer.automation.codebase.utils import (
     read_directory,
     read_specific_files,
 )
-from seer.automation.models import FileChange, FilePatch, Hunk, Line, RepoDefinition, Stacktrace
+from seer.automation.models import (
+    FileChange,
+    FilePatch,
+    Hunk,
+    InitializationError,
+    Line,
+    RepoDefinition,
+    Stacktrace,
+)
 from seer.automation.state import State
 from seer.db import DbRepositoryInfo, Session
 from seer.utils import class_method_lru_cache
@@ -221,7 +229,7 @@ class CodebaseIndex:
         workspace = CodebaseNamespaceManager.load_workspace(namespace_id, skip_copy=True)
 
         if not workspace:
-            raise ValueError("Failed to load workspace")
+            raise InitializationError("Failed to load workspace for namespace_id")
 
         try:
             repo_client = RepoClient.from_repo_info(workspace.repo_info)
--- a/src/seer/automation/codebase/codebase_index.py
+++ b/src/seer/automation/codebase/codebase_index.py
@@ -172,20 +172,14 @@ class CodebaseIndex:
 
         raise ValueError(f"Repository with id {repo_id} not found")
 
-    @classmethod
-    @traceable(name="Creating codebase index")
-    @ai_track(description="Creating codebase index")
+    @staticmethod
     def create(
-        cls,
         organization: int,
         project: int,
         repo: RepoDefinition,
-        embedding_model: SentenceTransformer,
         tracking_branch: str | None = None,
         sha: str | None = None,
-        state: State | None = None,
-        state_manager_class: Type[CodebaseStateManager] | None = None,
-    ):
+    ) -> int:
         repo_client = RepoClient(repo)
 
         head_sha = sha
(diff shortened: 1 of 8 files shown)


src/seer/automation/codebase/codebase_index.py
```diff
@@ -233,32 +247,32 @@ def create(
...
+            finally:
+                cleanup_dir(tmp_dir)
+        except Exception as e:
+            logger.error(f"Failed to create codebase index: {e}")
 
-            logger.debug(f"Create Step: Inserted {len(chunks)} chunks into the database")
+            try:
+                workspace.delete()
+            except Exception as ex:
+                sentry_sdk.capture_exception(ex)
```
ram-senth, line 273: Curious, why do we have `sentry_sdk.capture_exception` call here but not in the `except` above or in other exception handlers?

src/seer/automation/codebase/codebase_index.py
```diff
@@ -203,17 +197,37 @@ def create(
...
+        cls,
+        namespace_id: int,
+        embedding_model: SentenceTransformer,
+        state: State | None = None,
+        state_manager_class: Type[CodebaseStateManager] | None = None,
+    ):
+        workspace = CodebaseNamespaceManager.load_workspace(namespace_id, skip_copy=True)
+
+        if not workspace:
+            raise ValueError("Failed to load workspace")
```
ram-senth, line 224: Is there a way to identify why it failed to load workspace here? Will help if we have to troubleshooting this error in production. Also, see my comment for `load_workspace` method for changing this.

src/seer/automation/codebase/namespace.py
```diff
@@ -172,14 +173,16 @@ def load_workspace(cls, namespace_id: int):
...
-        did_copy = storage_adapter.copy_to_workspace()
+            did_copy = storage_adapter.copy_to_workspace()
 
-        cls._clear_mutex(namespace.id)
+            cls._clear_mutex(namespace.id)
 
-        if did_copy:
+        if skip_copy or did_copy:
             return cls(repo_info, namespace, storage_adapter)
         return None
```
ram-senth, line 187: We are returning None in this failure case but raising exception for other error cases above (line 164 and 169). Should we standardize and raise exception here too? I do realize that these lines are only indirectly related to current change for error handling in caller.

src/seer/automation/codebase/codebase_index.py
jennmueng: We want to raise and fail with the original exception still even if the delete call errors, and capture it so we don't lose it

src/seer/automation/codebase/namespace.py
jennmueng: The namespace handlers return none so we can gracefully handle the error when manipulating alot of the file movement. The codebase namespace raising is so it fails the celery worker instead of having it be a successful run. However, I think this is basically just down to preference and it works either way.

src/seer/automation/codebase/codebase_index.py
ram-senth: Does the other logging that you have added surface the reason why storage adapter failed to copy?

src/seer/automation/codebase/namespace.py
ram-senth: Correct me if I'm wrong - `index` method in codebase_index.py is the only place this method is invoked and it is raising an exception when None is returned, right? Where else is this method called where return of None is treated as not an error?

src/seer/automation/codebase/codebase_index.py
jennmueng: Yep it should log the exception from the storage adapter

src/seer/automation/codebase/namespace.py
jennmueng: `CodebaseIndex.from_repo_id` is the other place that calls it and also just raises an error in that case too

jennmueng: I'll actually make it return None, it makes more sense in that case

jennmueng: done

ram-senth (approved):

PR #682, opened 2024-05-06
fix(autofix-fs): Make sure storage location is clear

jennmueng:
Local runs were getting mixed up filesystem storage dumps, we already are deleting before saving to gcs in prod but good to add a test for it too.

### DIFF
--- a/src/seer/automation/codebase/storage_adapters.py
+++ b/src/seer/automation/codebase/storage_adapters.py
@@ -85,6 +85,7 @@ class FilesystemStorageAdapter(StorageAdapter):
                 shutil.rmtree(storage_path, ignore_errors=False)
             except Exception as e:
                 autofix_logger.exception(e)
+                return False
 
         shutil.copytree(self.tmpdir, storage_path, dirs_exist_ok=True)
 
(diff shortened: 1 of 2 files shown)


src/seer/automation/codebase/storage_adapters.py
```diff
@@ -79,6 +79,13 @@ def copy_to_workspace(self):
 
     def save_to_storage(self):
         storage_path = self.get_storage_location(self.repo_id, self.namespace_slug)
+
+        if os.path.exists(storage_path):
+            try:
+                shutil.rmtree(storage_path, ignore_errors=False)
+            except Exception as e:
+                autofix_logger.exception(e)
```
ram-senth, line 87: If `rmtree` fails then shouldn't we fail as the folder is probably corrupt?

src/seer/automation/codebase/storage_adapters.py
```diff
@@ -79,6 +79,14 @@ def copy_to_workspace(self):
...
+                return False
+
```
ram-senth, line 89: Do we need any additional test coverage for this?

ram-senth (approved):

PR #690, opened 2024-05-09
ref(similarity-embeddings): Reset grouping_records table in s4s

jangjodi:
Reset the grouping_records table if SENTRY_REGION is s4s
Remove janky logic to initialize the grouping_records table using a static pickle file

- this logic is incorrect for s4s
- this logic has already been applied for sentry and we will eventually want to correctly backfill later

### DIFF
--- a/src/migrations/versions/704569328ec7_delete_s4s_group_records.py
+++ b/src/migrations/versions/704569328ec7_delete_s4s_group_records.py
@@ -1,0 +1,62 @@
+"""delete s4s group records
+
+Revision ID: 704569328ec7
+Revises: b6cca7c6d99c
+Create Date: 2024-05-09 17:12:28.920636
+
+"""
+import os
+
+import sqlalchemy as sa
+from alembic import op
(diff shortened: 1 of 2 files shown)


src/migrations/versions/704569328ec7_delete_s4s_group_records.py
```diff
@@ -0,0 +1,62 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "704569328ec7"
+down_revision = "b6cca7c6d99c"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    if os.getenv("SENTRY_REGION", None) == "s4s":
```
JoshFerge, line 22: nit: could do return early to avoid [code arrowing ](https://blog.codinghorror.com/flattening-arrow-code/). but not a big deal in this case

JoshFerge (approved):

trillville (approved):

PR #695, opened 2024-05-10
feat(similarity_embedding): Create bulk insert endpoint

jangjodi:
Re-doing this [PR](https://github.com/getsentry/seer/pull/480/) using hash column (got the original PR in a bad state locally)

### DIFF
--- a/src/seer/grouping/grouping.py
+++ b/src/seer/grouping/grouping.py
@@ -158,7 +158,7 @@ class GroupingLookup:
     def encode_multiple_texts(self, stacktraces: List[str], batch_size: int = 1000) -> np.ndarray:
         """
         Encodes multiple stacktraces in batches using the sentence transformer model.
-        :param stacktraces: The list of stacktracse to encode.
+        :param stacktraces: The list of stacktraces to encode.
         :param batch_size: The batch size used for the computation.
         :return: The embeddings of the stacktraces.
         """
--- a/src/seer/grouping/grouping.py
+++ b/src/seer/grouping/grouping.py
@@ -32,6 +32,17 @@ class GroupingRequest(BaseModel):
         return v
 
 
+class CreateGroupingRecordData(BaseModel):
+    hash: str
+    project_id: int
+    message: str
+
+
+class CreateGroupingRecordsRequest(BaseModel):
+    data: List[CreateGroupingRecordData]
+    stacktrace_list: List[str]
+
+
 class GroupingRecord(BaseModel):
     group_id: Optional[int]
     project_id: int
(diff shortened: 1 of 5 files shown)


src/seer/grouping/grouping.py
```diff
@@ -140,6 +155,15 @@ def encode_text(self, stacktrace: str) -> np.ndarray:
         """
         return self.model.encode(stacktrace)
 
+    def encode_multiple_texts(self, stacktraces: List[str], batch_size: int = 1000) -> np.ndarray:
+        """
+        Encodes multiple stacktraces in batches using the sentence transformer model.
+        :param stacktraces: The list of stacktracse to encode.
```
JoshFerge, line 161:
```suggestion
        :param stacktraces: The list of stacktraces to encode.
```

tests/test_seer.py
```diff
@@ -303,6 +303,69 @@ def test_no_data_after_request_start(self):
...
+
+        response = app.test_client().post(
+            "/v0/issues/similar-issues/grouping-record",
+            data=json.dumps(record_requests),
+            content_type="application/json",
+        )
+        output = json.loads(response.get_data(as_text=True))
+        assert output == {"success": False}
+        for i in range(2, 4):
+            with Session() as session:
```
JoshFerge, line 361: 1) its not ideal to initialize a database session in a loop like this
2) is there a world whee we ever return `False` and make database calls? if not, then no need to check the database here

JoshFerge: 3. if we are going to check the database, can we do it in one query?

JoshFerge: (granted this is all in a test, but still good to be prudent about this)

JoshFerge (approved):

src/seer/grouping/grouping.py
```diff
@@ -198,6 +222,39 @@ def get_nearest_neighbors(self, issue: GroupingRequest) -> SimilarityResponse:
 
         return similarity_response
 
+    def bulk_create_and_insert_grouping_records(
```
trillville, line 225: based on the conversation @lobsterkatie, @JoshFerge, and i had last friday - we'll want to update this logic so that the pg inserts are done sequentially. 

Assuming input records are sorted by times_seen or first_seen (or similar) - for each record:
1. check if an existing record is more than 0.985 similar
2. If YES -> do not insert. if NO -> insert record

^ can be done in a separate PR if we want to just test the plumbing asap

jangjodi: yeah, I'm going to do this in a follow up PR

trillville (approved):

PR #697, opened 2024-05-10
fix(autofix): Increase pipeline step time limits

jennmueng:
Fixes [SEER-55](https://sentry.sentry.io/issues/5332811854/)

Loading from GCS takes longer, p75 of 7 seconds which would easily make it hit the 15 second soft limit. Increase it to 50 seconds because this should be killing overrun steps not just steps that take long.


trillville (approved):
lgtm 😎

PR #698, opened 2024-05-10
fix(autofix): model dump when calling apply_async

jennmueng:
Fixes [SEER-56](https://sentry.sentry.io/issues/5332812124/)


trillville (approved):
LGTM 🚀

PR #969, opened 2024-07-29
feat(dynamic alert thresholds): Streaming detection

ram-senth:
Implementation of stream anomaly detection. 
There are some TODOs and unit tests still pending.


src/seer/anomaly_detection/accessors.py
```diff
@@ -0,0 +1,133 @@
+import abc
+import datetime
+import logging
+from typing import List
+
+# import numpy.typing as npt
```
trillville, line 6: remove?

ram-senth: Yes, will do

src/seer/anomaly_detection/accessors.py
```diff
@@ -0,0 +1,133 @@
...
+                        "external_alert_id": external_alert_id,
+                    },
+                )
+                return None
+
+            timeseries: TimeSeries = TimeSeries(
+                timestamps=np.empty([len(alert_info.timeseries)]),
+                values=np.empty([len(alert_info.timeseries)]),
+            )
+            for i, point in enumerate(alert_info.timeseries):
```
trillville, line 59: nit: can we do this more efficiently than a loop? arent we effectively going from one array to another

src/seer/anomaly_detection/accessors.py
```diff
@@ -0,0 +1,133 @@
...
+        config: AnomalyDetectionConfig,
+        timeseries: List[TimeSeriesPoint],
+    ):
+        with Session() as session:
+            existing_records = (
+                session.query(DbDynamicAlert).filter_by(external_alert_id=external_alert_id).count()
+            )
+
+            if existing_records > 0:
+                # This logic assumes that alert id is unique across organizations and projects.
```
trillville, line 85: can we just filter by a combination of alert and project to be safer?

trillville: may want to check with foundations on this assumption

ram-senth: I have confirmed with them already. Our anomaly detection API accepts just the alert id as of now. In order to lookup using project we will need to update the API. Will check with them again to see if that is required.

src/seer/anomaly_detection/accessors.py
ram-senth: Do you mean using list comprehension instead?

trillville: no since tht would be a loop as well - can't we just directly reference `alert_info.timeseries`?

ram-senth: We need to convert it to a Numpy array for matrix profile computation. Hence this loop. One thing I can do is change the representation in TimeSeries to store two arrays instead of an array of objects. That would avoid the looping. However, this looping is not that costly as we are talking about a max of 2500 items in the list.

src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -28,39 +29,85 @@
...
+        historic = self.alert_data_accessor.query(alert.id)
+        if historic is None:
+            raise Exception(f"Invalid alert id {alert.id}")
+
+        # TODO: Need to check the time gap between historic data and the new datapoint against the alert configuration
+
+        # Run batch detect on history data
+        # TODO: This step can be optimized further by caching the matrix profile in the database
+        mp_config = MPConfig(ignore_trivial=True, normalize_mp=False)
+        batch_detector = MPBatchAnomalyDetector(
```
trillville, line 69: this part is confusing to me - why are we doing both batch and stream here?

src/seer/anomaly_detection/models/external.py
```diff
@@ -1,14 +1,12 @@
...
 from pydantic import BaseModel, ConfigDict, Field
 
-from seer.anomaly_detection.models.internal import (
-    AnomalyFlags,
-    Directions,
-    Seasonalities,
-    Sensitivities,
-    TimePeriods,
-)
+AnomalyFlags = Literal["none", "anomaly_low", "anomaly_high", "no_data"]
```
trillville, line 5: do you think we can make these a bit more clear to avoid any potential confusion about whether low/high refers to the direction vs the magnitude? i know it will require some coordination with foundations

src/seer/anomaly_detection/anomaly_detection.py
ram-senth: This is the pending optimization step that the comment in line 67 refers to. We should compute the matrix profile of historic data once using batch anomaly detection and store that in DB. However, there are some complexities around making it generic enough to support future extensions to use non-matrix profile based detection. So, for now I'm not doing that, especially since the performance of this seems to be pretty good (~80 ms on local). Flip side is that we batch compute the matrix profile for historic data in each call.

src/seer/anomaly_detection/models/external.py
ram-senth: Yes will check with them about changing this. Do you have any recommendations for names? `anomaly_lower_confidence` and `anomaly_higher_confidence` work better?

ram-senth: Have made the change to the strings

trillville (approved):

PR #972, opened 2024-07-29
fix(autofix): Fix embedding model loading condition

jennmueng:
Embedding model's loading condition should be based off of `no_codebase_indexes` which includes the request param not the `skip_loading_codebase` argument, which is only passed on the PR creation.


trillville (approved):
good catch - hopefully we see memory use drop now :)

PR #974, opened 2024-07-29
ref(autofix): Stop logging github key issues in dev + use AppConfig

jennmueng:
Fix the annoying github key logs from showing up in dev, also refactor to use the `AppConfig` from dependency injection

### DIFF
--- a/src/seer/automation/codebase/repo_client.py
+++ b/src/seer/automation/codebase/repo_client.py
@@ -16,6 +16,8 @@ from seer.automation.autofix.utils import generate_random_string, sanitize_branc
 from seer.automation.codebase.models import RepositoryInfo
 from seer.automation.codebase.utils import get_language_from_path
 from seer.automation.models import FileChange, InitializationError, RepoDefinition
+from seer.configuration import AppConfig
+from seer.dependency_injection import inject, injected
 from seer.utils import class_method_lru_cache
 
 logger = logging.getLogger("autofix")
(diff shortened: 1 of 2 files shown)


src/seer/automation/codebase/repo_client.py
```diff
@@ -45,41 +47,46 @@ def get_repo_app_permissions(
...
+@inject
+def get_write_app_credentials(config: AppConfig = injected) -> tuple[int | str | None, str | None]:
+    app_id = config.GITHUB_APP_ID
+    private_key = config.GITHUB_PRIVATE_KEY
 
     if not app_id or not private_key:
-        logger.exception(
-            InitializationError(
-                "GITHUB_APP_ID and GITHUB_PRIVATE_KEY environment variables must be set for app authentication."
+        if not config.DEV:
```
corps, line 66: Might be worth adding this directly to `do_validation`: https://github.com/getsentry/seer/blob/main/src/seer/configuration.py#L72

Reason being, explicitly throwing during startup would still hit sentry_sdk (model_validation happens after sdk setup), but it would also cause the deploy to fail (which is good, the rollout would fail and the older version would then be retained).  Rollout in GKE watches for health checks before continuing the replication, reverting if something goes wrong.

But I leave it to your discretion.

corps (approved):

src/seer/automation/codebase/repo_client.py
jennmueng: Great solution!

PR #975, opened 2024-07-29
feat(autofix-evals): Run random or specific test

jennmueng:
Can now run a random eval item from a set, or a specific one. This is to aid in debugging and R&D.


corps (approved):
This is fine for now, but any reason we don't just expose a CLI That allows invoking this evaluation directly from the CLI?  the python click library provides a convenient way of setting up scripts with a CLI.

jennmueng:
> This is fine for now, but any reason we don't just expose a CLI That allows invoking this evaluation directly from the CLI? the python click library provides a convenient way of setting up scripts with a CLI.

Good point, and make it a `make` command, will follow up...

PR #977, opened 2024-07-30
feat(grouping): add reranking logic

trillville:
quickly tested and it takes ~1ms for 100 records on my laptop, we can probably bump this up but lets confirm performance in prod first. i also added some sentry metrics dogfooding stuff - lmk if this seems appropriate

logic:

- take first 100 records from pg, sorted by HNSW index implied distance - no longer include "distance" calculation from pg, require at least 0.05 distance for candidates
- manually calculate cosine distance for all 100 candidates
- rerank based on distance and then return top k

### DIFF
--- a/src/seer/app.py
+++ b/src/seer/app.py
@@ -110,8 +110,9 @@ def breakpoint_trends_endpoint(data: BreakpointRequest) -> BreakpointResponse:
 def similarity_endpoint(data: GroupingRequest) -> SimilarityResponse:
     with sentry_sdk.start_span(op="seer.grouping", description="grouping lookup"):
         sentry_sdk.set_tag("read_only", data.read_only)
-        sentry_sdk.set_tag("stacktrace_len", len(data.stacktrace))
-        sentry_sdk.set_tag("request_hash", data.hash)
+        sentry_sdk.metrics.distribution(
+            key="stacktrace_len", value=len(data.stacktrace), unit="none"
+        )
         similar_issues = grouping_lookup().get_nearest_neighbors(data)
     return similar_issues
 
--- a/src/seer/app.py
+++ b/src/seer/app.py
@@ -110,6 +110,8 @@ def breakpoint_trends_endpoint(data: BreakpointRequest) -> BreakpointResponse:
 def similarity_endpoint(data: GroupingRequest) -> SimilarityResponse:
     with sentry_sdk.start_span(op="seer.grouping", description="grouping lookup"):
         sentry_sdk.set_tag("read_only", data.read_only)
+        sentry_sdk.set_tag("stacktrace_len", len(data.stacktrace))
+        sentry_sdk.set_tag("request_hash", data.hash)
         sentry_sdk.metrics.distribution(
             key="stacktrace_len", value=len(data.stacktrace), unit="none"
         )
--- a/src/seer/grouping/grouping.py
+++ b/src/seer/grouping/grouping.py
@@ -178,48 +178,44 @@ class GroupingLookup:
         distance: float,
         k: int,
         hnsw_candidates: int = 100,
-        hnsw_distance: float = 0.95,
+        hnsw_distance: float = 0.05,
     ) -> List[tuple[DbGroupingRecord, float]]:
         custom_options = {"postgresql_execute_before": "SET LOCAL hnsw.ef_search = 100"}
 
         candidates = (
-            session.query(
-                DbGroupingRecord,
-                DbGroupingRecord.stacktrace_embedding.cosine_distance(embedding).label("distance"),
-            )
+            session.query(DbGroupingRecord)
             .filter(
                 DbGroupingRecord.project_id == project_id,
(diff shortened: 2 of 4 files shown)


src/seer/app.py
```diff
@@ -110,8 +110,9 @@ def breakpoint_trends_endpoint(data: BreakpointRequest) -> BreakpointResponse:
 def similarity_endpoint(data: GroupingRequest) -> SimilarityResponse:
     with sentry_sdk.start_span(op="seer.grouping", description="grouping lookup"):
         sentry_sdk.set_tag("read_only", data.read_only)
-        sentry_sdk.set_tag("stacktrace_len", len(data.stacktrace))
-        sentry_sdk.set_tag("request_hash", data.hash)
```
trillville, line 114: based on https://docs.sentry.io/product/explore/metrics/#limits-and-restrictions - not sure if this is important to keep though for debugging

src/seer/grouping/grouping.py
```diff
@@ -19,6 +19,8 @@
 logger = logging.getLogger(__name__)
 
 NN_GROUPING_DISTANCE = 0.01
+NN_GROUPING_HNSW_DISTANCE = 0.05
+NN_GROUPING_HNSW_CANDIDATES = 100
```
jangjodi, line 23: I'm wondering if we want to start this off smaller, maybe 50, and then have sentry control this value with an option as we see how long this takes

src/seer/grouping/grouping.py
```diff
@@ -177,23 +181,53 @@ def query_nearest_k_neighbors(
...
+            .limit(max(k, hnsw_candidates))
             .execution_options(**custom_options)
+            .all()
         )
-        return query.all()
+
+        reranked = self.rerank_candidates(candidates, embedding, distance)
+
+        if candidates and reranked and candidates[0].hash != reranked[0][0].hash:
+            sentry_sdk.metrics.incr(
```
jangjodi, line 206: can we also add a log here of the project id and two hashes so we can analyze them?

src/seer/grouping/grouping.py
trillville: this should be a very safe value - let's just merge this when we arent doing a backfill - we can see pretty quickly from the span how long its taking

src/seer/grouping/grouping.py
```diff
@@ -177,23 +181,58 @@ def query_nearest_k_neighbors(
...
+    @sentry_sdk.tracing.trace
+    def rerank_candidates(
+        candidates: List[DbGroupingRecord], embedding: np.ndarray, distance: float
+    ) -> List[tuple[DbGroupingRecord, float]]:
+        embedding_norm = np.linalg.norm(embedding)
+
+        def cosine_distance(candidate_embedding):
+            candidate_norm = np.linalg.norm(candidate_embedding)
+            dot_product = np.dot(candidate_embedding, embedding)
+            return 1 - dot_product / (candidate_norm * embedding_norm)
```
JoshFerge, line 228: why not just use `from sklearn.metrics.pairwise import cosine_similarity`?

JoshFerge (dismissed):
could we add this behind a parameter so that we can control the rollout of this via option? like `use_reranking`.

src/seer/grouping/grouping.py
trillville: this implementation allows me to precompute `embedding_norm` rather than recalculate it every time we loop through our candidates (tbf not a big deal either way)

src/seer/grouping/grouping.py
```diff
@@ -177,23 +187,138 @@ def query_nearest_k_neighbors(
...
+
+        candidates = (
+            session.query(DbGroupingRecord)
             .filter(
                 DbGroupingRecord.project_id == project_id,
                 DbGroupingRecord.stacktrace_embedding.cosine_distance(embedding) <= distance,
                 DbGroupingRecord.hash != hash,
             )
-            .order_by("distance")
+            .order_by(DbGroupingRecord.stacktrace_embedding.cosine_distance(embedding))
```
JoshFerge, line 241: what exactly does this change do? and is it any different from existing behavior?

src/seer/grouping/grouping.py
```diff
@@ -177,23 +187,138 @@ def query_nearest_k_neighbors(
...
             .limit(k)
             .execution_options(**custom_options)
+            .all()
+        )
+
+        return [
```
JoshFerge, line 247: this is just to keep the interface between the two functions the same i assume?

src/seer/grouping/grouping.py
trillville: yeah, i got rid of `"distance"` since we are manually calculating it now (seemed weird to pull it and then overwrite it) - `DbGroupingRecord.stacktrace_embedding.cosine_distance(embedding)` is just how `"distance"` is defined in the old code

JoshFerge (approved):

PR #981, opened 2024-07-30
feat(autofix): New Plan+Code Agent to replace planner & executor

jennmueng:
Introduces a new **Plan+Code Agent** with two stages to replace the combined planning + execution agent.

Functionality as detailed in the below diagram:
image

Overall Scores:
```
Run: fix-plan-code-6
--Scores--
Mean: 0.56
Median: 0.60
Max: 1.00
Min: 0.00
Num of >=0.9 scores: 8
--Error Weighted Scores--
Mean: 0.40
Median: 0.42
Max: 1.00
Min: 0.00
--Errors--
Errored: 22, rate: 0.29
Errored in Root Cause: 10, rate: 0.13
Errored in Planning: 12, rate: 0.16
```

### Included in this PR:
- Fix for #997 where we were not correctly escaping the code snippet in the root cause, causing XML issues. CDATA blocks are not needed as they could confuse the LLM even further.
- Fix for brittle celery test case, use set comparison instead


src/seer/automation/autofix/components/coding/models.py
```diff
@@ -19,7 +20,7 @@ class SnippetXml(PromptXmlModel, tag="snippet"):
     snippet: Annotated[str, StringConstraints(strip_whitespace=True)]
 
 
-class RootCausePlanTaskPromptXml(PromptXmlModel, tag="task", skip_empty=True):
+class RootCausePlanTaskPromptXml(PromptXmlModel, tag="root_cause", skip_empty=True):
     title: str = element()
     description: str = element()
     fix_title: Optional[str] = element()
```
roaga, line 26: Can we remove/rename references to suggested fixes since we no longer have those?

src/seer/automation/autofix/components/coding/models.py
```diff
@@ -52,70 +53,79 @@ class PlanStep(BaseModel):
...
+        if lines and lines[-1].strip() == "```":
+            lines = lines[:-1]
+        return "\n".join(lines).strip()
 
     @classmethod
     def get_example(cls):
         return cls(
             file_path="path/to/file.py",
             repo_name="owner/repo",
+            type="Either 'file_change', 'file_create', or 'file_delete'",
```
roaga, line 79: Grammar nit: "Either" is only for lists of 2 items. Idk if it matters to the model though.

src/seer/automation/autofix/components/coding/models.py
```diff
@@ -52,70 +53,79 @@ class PlanStep(BaseModel):
...
 
     @classmethod
     def get_example(cls):
         return cls(
             tasks=[
-                ReplaceCodePromptXml.get_example(),
-                CreateFilePromptXml.get_example(),
+                PlanTaskPromptXml.get_example(),
+                PlanTaskPromptXml.get_example(),
             ]
```
roaga, line 110: We're giving it two of the same example? Shouldn't we give it two different examples?

src/seer/automation/autofix/components/coding/prompts.py
```diff
@@ -33,12 +29,25 @@ def format_default_msg(event: EventDetails, task_str: str, instruction: str | No
             Given the issue:
             {event_str}
 
-            And some relevant context in the codebase:
+
+
+            The root cause of the issue has been identified and a fix has been suggested. The fix is as follows:
             {task_str}
```
roaga: We shouldn't tell the model that we have a suggested fix. We're giving it code context instead. You can probably look back in the main branch to see what prompt I had here and modify that.

src/seer/automation/autofix/components/coding/utils.py
```diff
@@ -0,0 +1,157 @@
...
+    changes = []
+    diff_chunks = extract_diff_chunks(task.diff)
+
+    for chunk in diff_chunks:
+        result = find_original_snippet(
+            chunk.original_chunk,
+            file_content,
+            threshold=0.75,
+            initial_line_threshold=0.95,
+        )
```
roaga, line 140: How did we decide these numbers? Is it worth an effort to find "optimal" numbers?

src/seer/automation/autofix/components/coding/utils.py
```diff
@@ -0,0 +1,157 @@
...
+                    change_type="edit",
+                    path=task.file_path,
+                    reference_snippet=original_snippet,
+                    new_snippet=chunk.new_chunk,
+                    description=task.description,
+                )
+            )
+        else:
+            logger.info(f"Original snippet not found in file {task.file_path}")
+            append_langfuse_trace_tags(["skipped_file_change:snippet_not_found"])
```
roaga, line 155: Do we give that feedback to the model so it can try again?

src/seer/automation/autofix/components/coding/models.py
jennmueng: It's actually less of an example and more prompts for the diff now, examples seemed to confuse it in the past

src/seer/automation/autofix/components/coding/utils.py
jennmueng: The numbers were selected to allow the new snippet-wide fuzzy matcher to be a bit more lenient on the entire snippet comparison, and the initial line threshold higher to make sure the initial line is exactly the same. It can definitely be improved upon but the effort will be high as we'll need to compile a ton of examples.

src/seer/automation/autofix/components/coding/utils.py
jennmueng: No, not yet, will be done in a followup to this.

roaga (approved):

PR #982, opened 2024-07-30
chore(dynamic alert thresholds) Moving to dependency injection for inversion of control

ram-senth:
Migrating to our DI framework


corps (approved):
Yup, this looks good.  With my PR in feel free to inject as attributes now by decorating the class directly.

PR #987, opened 2024-07-31
feat: Langfuse in DI + disable in tests

jennmueng:
1. Langfuse env vars in `AppConfig`
2. Dependency injected langfuse 
3. Util method to add tags to a currently running trace

### DIFF
--- a/src/seer/langfuse.py
+++ b/src/seer/langfuse.py
@@ -1,0 +1,51 @@
+import logging
+
+from langfuse import Langfuse
+from langfuse.decorators import langfuse_context
+
+from seer.configuration import AppConfig
+from seer.dependency_injection import Module, inject, injected
+
+logger = logging.getLogger(__name__)
+
+langfuse_module = Module()
+
+
+@inject
+def initialize_langfuse_context(config: AppConfig = injected):
+    langfuse_context.configure(
+        public_key=config.LANGFUSE_PUBLIC_KEY,
+        secret_key=config.LANGFUSE_SECRET_KEY,
+        host=config.LANGFUSE_HOST,
+        enabled=bool(config.LANGFUSE_HOST),
+    )
+
+
+@langfuse_module.provider
+def provide_langfuse(config: AppConfig = injected) -> Langfuse:
+    return Langfuse(
+        public_key=config.LANGFUSE_PUBLIC_KEY,
+        secret_key=config.LANGFUSE_SECRET_KEY,
+        host=config.LANGFUSE_HOST,
+        enabled=bool(config.LANGFUSE_HOST),
+    )
+
+
+@inject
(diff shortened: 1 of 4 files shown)


src/seer/langfuse.py
```diff
@@ -0,0 +1,51 @@
...
+    langfuse_context.configure(
+        public_key=config.LANGFUSE_PUBLIC_KEY,
+        secret_key=config.LANGFUSE_SECRET_KEY,
+        host=config.LANGFUSE_HOST,
+        enabled=bool(config.LANGFUSE_HOST),
+    )
+
+
+@langfuse_module.provider
+def provide_langfuse(config: AppConfig = injected) -> Langfuse:
```
jennmueng, line 25: @corps If I want to test this method, and override the app config with the one from `langfuse_configuration_test_module`, how would I do that? I tried it and I assume because we don't have @inject here it's not consuming from that module?

corps (approved):
Looks great!  Eventually, if you see that making the test langfuse module is useful in other tests (or should be the test default), you can add it to `conftest.py`.  But that's at your discretion if it makes sense.

src/seer/langfuse.py
jennmueng: nvm, looks like it already works huh, maybe I messed something up earlier

corps: Simple, try this:

```
@test_langfuse_module.provider
def test_config() -> AppConfig:
    base = invoke_original_config_method()
    ...
    return base
```

In other words, if you provide the same type of value in a different module, that module overrides proceeding module's definition of that type.

You can even create a module _within a specific test_.

```
def test_thing():
    override = Module()
    @override.provider
    ....
    with override:
        ....
```

jennmueng:
> Looks great! Eventually, if you see that making the test langfuse module is useful in other tests (or should be the test default), you can add it to `conftest.py`. But that's at your discretion if it makes sense.

For other tests it has the default empty string to disable langfuse, so this module should only be for here to test the config/provider methods

ram-senth:
@corps I think we should standardize how we define the providers. Current setup has providers in multiple files. Will be very hard to reason about them. Not directly related to this PR but need to follow-up.

corps:
> @corps I think we should standardize how we define the providers. Current setup has providers in multiple files. Will be very hard to reason about them. Not directly related to this PR but need to follow-up.

Fair.  Part of the weakness has to do with python's module import system.
The simplest answer, that we share one `module`, and we put providers in a separate file, has problems with import circles and the fact that the module itself must be guaranteed to be imported to count.

The other alternative is that types have providers next to them in the same file, with their own local module, works, until it doesn't work due to, again, import circles from base python.

This is probably part of the reason the DI isn't yet widely adopted in python.  I'm pondering solutions but I don't definitively have on at the moment.

ram-senth:
Let us add this as a discussion topic for today's standup.

PR #991, opened 2024-07-31
fix: Fix langfuse init

jennmueng:
looks like we have to call `langfuse_context.configure` before any of the decorators are used, otherwise it messes up the langfuse singleton...we can't do that and use the AppConfig, so we revert to just using the env variables for langfuse but keep the checks in AppConfig


corps (approved):
That's unfortunate, but it definitely happens from time to time.  Let's not force injection when it gets in the way.   It's a tool to be used when it provides value, not a goal in of itself.  Good call!

PR #1008, opened 2024-08-05
feat(evals): Add description and multiple runs per item

jennmueng:
Also just passes the pydantic model to the eval method


roaga (approved):
LGTM

Issue #1004, opened 2024-08-02
(autofix): Root cause should give hyperlinks to code


trillville:
https://www.notion.so/sentry/root-cause-analysis-should-give-hyperlinks-to-code-d05ac92eacf246b6a784e69138ffe4f0?pvs=4

roaga:
Done, but line numbers not supported because LLMs are bad at counting.

PR #1010, opened 2024-08-05
feat(autofix): Add repo name to root cause code context snippets

roaga:
Adds repository name alongside file path in code snippets. This is to support code hyperlinks on the frontend. Issue #1004 

Sentry PR: https://github.com/getsentry/sentry/pull/75610

### DIFF
--- a/src/seer/automation/autofix/components/root_cause/models.py
+++ b/src/seer/automation/autofix/components/root_cause/models.py
@@ -11,7 +11,7 @@ from seer.automation.models import EventDetails, PromptXmlModel
 
 class SnippetPromptXml(PromptXmlModel, tag="code"):
     file_path: str = attr()
-    repo_name: str = attr()
+    repo_name: Optional[str] = attr()
     snippet: Annotated[str, StringConstraints(strip_whitespace=True)]
 
     @classmethod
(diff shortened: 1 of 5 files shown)


src/seer/automation/autofix/components/root_cause/models.py
```diff
@@ -11,18 +11,21 @@
 
 class SnippetPromptXml(PromptXmlModel, tag="code"):
     file_path: str = attr()
+    repo_name: str = attr()
```
jennmueng, line 14: Hmm I think we should let `repo_name` be optional in the cases where there's only one repo, we don't enforce it on the tools in that case

jennmueng: If we're not making it optional, we should make sure we're giving the llm the repo name in the prompts

roaga: We do need it sent to the frontend no matter what, but I made the field `Optional` for flexibility within Seer

jennmueng: If there's only one repo then on the frontend we can just make it implied that it's that one whenever it's null here

jennmueng (approved):

PR #1016, opened 2024-08-07
(autofix): Prompt engineering to make root cause output descriptions more detailed

roaga:
An example below of how it's more informative to the user now.
Also results in a small eval bump from 0.43/0.53 to 0.44/0.54 (error weighted/normal).

Screenshot 2024-08-07 at 7 35 54 AM


jennmueng (approved):
LGTM

PR #1017, opened 2024-08-07
ci: combine the test workflows

jennmueng:
Should only run the models in the branches, skip on PRs


.github/workflows/tests.yml
```diff
@@ -31,11 +32,20 @@ jobs:
           id_token_audience: "610575311308-9bsjtgqg4jm01mt058rncpopujgk3627.apps.googleusercontent.com"
           id_token_include_email: true
       - name: Set up Cloud SDK
+        if: github.event_name == 'push'
         uses: google-github-actions/setup-gcloud@v1
-      - name: Fetch models + Build image
+      - name: Fetch models
+        if: github.event_name == 'push'
```
corps, line 38: Do we also need to filter on branch name here?  In other words, the I'm curious what the interaction of the `on` does here, do pushes to all PRs also trigger?

jennmueng: It should only trigger push events on those branches, on PRs the event name is still pull_request it looks like, just tried w/ empty commit and it still runs as expected:

Logged as 
```
if [[ "pull_request" == "pull_request" ]]; then
```
in the compose set step

corps (approved):

PR #1024, opened 2024-08-08
feat(autofix): Allow repeating steps + backwards compatibility

jennmueng:
Allows multiple of the same step inside an autofix run.
Each step in the autofix step state will now have a unique UUID, with the unique strings now in the `key` field.

Backwards compatible and will migrate existing runs via `event_manager.migrate_step_keys()`

Screenshot 2024-08-07 at 5 08 02 PM

### DIFF
--- a/src/seer/automation/autofix/event_manager.py
+++ b/src/seer/automation/autofix/event_manager.py
@@ -1,6 +1,5 @@
 import dataclasses
 import logging
-from typing import Literal
 
 from seer.automation.autofix.components.coding.models import CodingOutput
 from seer.automation.autofix.components.root_cause.models import RootCauseAnalysisOutput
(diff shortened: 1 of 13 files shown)


src/seer/automation/autofix/models.py
```diff
@@ -343,13 +371,10 @@ def mark_triggered(self):
...
         self.updated_at = datetime.datetime.now()
 
-    def delete_steps_after(self, step: Step):
-        steps_to_keep = []
-        for cur_step in self.steps:
-            steps_to_keep.append(cur_step)
-            if step.id == cur_step.id:
-                break
-        self.steps = steps_to_keep
+    def delete_steps(self, step: Step, include_current: bool = False):
```
roaga, line 374: This new name doesn't communicate that we're deleting specifically the steps after the provided step.

src/seer/automation/autofix/event_manager.py
```diff
@@ -29,42 +28,55 @@ class AutofixEventManager:
...
+            for step in cur.steps:
+                step.ensure_uuid_id()
 
-            cur.status = AutofixStatus.PROCESSING
+    def send_root_cause_analysis_pending(self):
+        with self.state.update() as cur:
+            step = cur.add_step(self.root_cause_analysis_processing_step)
+            step.status = (
+                AutofixStatus.PROCESSING
+            )  # We want it to be spinning on the UI, not a grey pending, so we just make it processing.
```
roaga, line 68: What's even the point of the PENDING status anymore? We're not using it here, we don't have execution sub-steps anymore...what's left?

roaga: And here specifically, why do we need both `send_root_cause_analysis_pending` and `send_root_cause_analysis_start` - isn't it a bit redundant?

src/seer/automation/autofix/models.py
```diff
@@ -107,7 +108,8 @@ class StepType(str, enum.Enum):
 
 
 class BaseStep(BaseModel):
-    id: str
+    id: str = Field(default_factory=lambda: str(uuid.uuid4()))
```
roaga, line 111: What's the difference between `id` and `key`? It's kind of confusing at first glance

src/seer/automation/autofix/event_manager.py
jennmueng: `send_root_cause_analysis_pending` is sent before the celery task is accepted, this way we get the UI to start showing "Analyzing Issue" asap, the celery task takes a sec to start

jennmueng: I agree, we can remove pending

src/seer/automation/autofix/models.py
jennmueng: Will add comment

src/seer/automation/autofix/models.py
jennmueng: actually the old name worked anyways, `delete_steps_after`

roaga (approved):

PR #1025, opened 2024-08-08
feat(autofix): Allow steps to retry on an exception

jennmueng:
Allow steps to retry on an exception, this is like a high-level catch all for issues that we don't specifically handle.

Current retry settings:
1. Root cause -> maximum 2 retires or 3 runs max
2. Plan+Code-> maximum 2 retries or 3 runs max
3. Change Describer -> maximum 1 retries or 2 runs max

Screenshot 2024-08-07 at 5 08 02 PM

### DIFF
--- a/src/seer/automation/autofix/components/coding/utils.py
+++ b/src/seer/automation/autofix/components/coding/utils.py
@@ -132,6 +132,11 @@ def task_to_file_change(task: PlanTaskPromptXml, file_content: str) -> list[File
     diff_chunks = extract_diff_chunks(task.diff)
 
     for chunk in diff_chunks:
+        if chunk.original_chunk == chunk.new_chunk:
+            continue
+        if chunk.original_chunk.strip() == "":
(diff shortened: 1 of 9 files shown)


src/seer/automation/autofix/components/coding/utils.py
```diff
@@ -132,6 +132,11 @@ def task_to_file_change(task: PlanTaskPromptXml, file_content: str) -> list[File
     diff_chunks = extract_diff_chunks(task.diff)
 
     for chunk in diff_chunks:
+        if chunk.original_chunk == chunk.new_chunk:
+            continue
+        if chunk.original_chunk.strip() == "":
```
jennmueng, line 137: just tacking this on but it was something I ran into while trying to evaluate this PR, when the original chunk is an empty string, it selects the whole file...

roaga (approved):
Code looks good but I really think we should revisit the UX

PR #1032, opened 2024-08-12
feat(autofix): Autofix run timeout checks

jennmueng:
On a call to get the state of an Autofix, we will check if it hasn't been updated in the last 90 seconds or been running for > 15 minutes. If yes we will mark it errored and log the error to sentry so we can be alerted.

Includes a migration to pull out the `last_triggered_at` and `updated_at` fields from the JSON (will default to current timestamp if those values don't exist) + another migration to index those columns.

This PR also sets up celerybeat and prepares a scheduled task that runs every hour to run the autofix check on runs started within the last 1 hour and 15 minutes. This will catch any runs that fall through and aren't caught by the state call. **The scheduled task is not yet enabled** by this PR, in case that there are runs that default to the current timestamp in the migration, we can enable it later as a followup.

# Next Steps
1. Followup PR to uncomment the celerybeat task, should be deployed at least 1h 15m after this
2. Set up an alert on seer's sentry for this error


corps (approved):
Looks good, solid approach.

PR #1035, opened 2024-08-13
fix the run state migration

jennmueng:


corps (approved):
And just to sanity check -- have you validated this works locally against some actual data?

Issue #998, opened 2024-08-01
(autofix): Make use of more Sentry context


roaga:
Make use of more Sentry context.

Very useful:
     - **Issue summary** or at least components like chain of thought (DONE)
      - **Trace connected issues**: magically fix the correct issue
    - **Breadcrumbs** (DONE)
    - Suspect commit, if we can make it really smart and accurate
    
Marginal value:
    - Tags (these cover all events in an issue)
    - Spans from the trace

Minimal value:
    - Key value pairs at bottom (”contexts”)
    - Multiple events

PR #1036, opened 2024-08-13
(autofix): Add breadcrumbs support, pre-filter issue context

roaga:
Start leveraging breadcrumbs context in Autofix in addition to stack traces (part of #998 ).

Add a gpt-4o-mini call before the root cause step and before the coding step to filter out context that is probably useless for debugging.

Eval score: 0.46/0.55, which is comparable to current main branch


jennmueng (approved):

jennmueng:
Probably want to add/fix test cases but otherwise LGTM

PR #1037, opened 2024-08-13
Fix celery bootup and configuration

corps:
Fixes several celery bootup and configuration issues by using the right signal.  Adds test to validate.


src/migrations/versions/0226c3793a61_migration.py
```diff
@@ -5,6 +5,7 @@
 Create Date: 2024-02-09 14:01:48.518223
 
 """
+
```
corps, line 8: This whitespaces are an incidental `black` change of opinion on whitespace.  🤷

docker-compose.yml
```diff
@@ -37,6 +37,11 @@ services:
             - ~/.config/gcloud:/root/.config/gcloud
             # Codecov test artifacts
             - ./.artifacts:/app/.artifacts
+        healthcheck:
```
corps, line 40: You now can validate the smoke test locally by running `docker compose up -d` followed by calls to `docker compose ps` and observing changes to the healthy state.

jennmueng (approved):

ram-senth:
Wanted to confirm if my understanding is right. The basic idea is to have one `smoke_test_id` per instance and have the status for that test stored in the db. This id gets regenerated whenever the instance restarts? Do we need to worry about clearing out old data in the `smoke_tests` table?

corps:
> Wanted to confirm if my understanding is right. The basic idea is to have one `smoke_test_id` per instance and have the status for test against each instance stored in the db. This id gets regenerated whenever the instance restarts? Do we need to worry about clearing out old data in the `smoke_tests` table?

I intend to add a TTL job to clean up the smoke tests, as it's cheap and useful to do it that way.

The intention is to run the smoke test *just* on the canary, as it depends upon the environment variable which I will modify our k8s canary config to perform.  Thus, we get one smoke test per canary.

The smoke test id is a uuid4 attached to the `AppConfig` whenever AppConfig is generated.

corps:
Thank you guys!

Issue #1044, opened 2024-08-14
(autofix): 90 day TTL for autofix state


trillville:
- 90 day for autofix state

PR #1049, opened 2024-08-15
(autofix) Add celery job to delete Autofix runs over 90 days old

roaga:
To meet our 90 day data retention policy, add a celery job that runs once a day to delete any Autofix runs over 90 days old.

Issue #1044


jennmueng (approved):
lgtm

src/seer/automation/autofix/tasks.py
```diff
@@ -97,6 +97,15 @@ def get_all_autofix_runs_after(after: datetime.datetime):
         return [ContinuationState.from_id(run.id, AutofixContinuation) for run in runs]
 
 
+def delete_all_runs_before(before: datetime.datetime):
+    with Session() as session:
+        deleted_count = (
+            session.query(DbRunState).filter(DbRunState.last_triggered_at < before).delete()
```
corps, line 103: A few things.

1.  Let's send up a sentry metric associated with the count.  https://docs.sentry.io/product/explore/metrics/. You can count using these parts of the sdk.

2.  Can we batch this and put an upper bound on each delete?  To do this, try something like this:

```
while True:
   with Session() as session:
        deleted_count = (
            session.query(DbRunState).filter(DbRunState.last_triggered_at < before).limit(BATCH_SIZE).delete() 
        if deleted_count == 0:
          break
```

corps: Reasons for batching: when a delete operation occurs on the database it must hold a wide lock based on the filter that ca be contentious.  Furthermore, we have limits on the time any one query can run (solution to the halting problem of a non halting query), so if a batch is too big it will fail (and not commit) due to timeout. 

Breaking down the work to fixed sizes gives us a bit more consistent behavior.
Also please writes some tests that involve > BATCH_SIZE data to validate the looping logic!

corps: Also hint: make `BATCH_SIZE` an optional argument to the function.  The default can be like, 500 or 1000, but in a test, you can pass a smaller number (2 or 3) to make the test faster and easier.

tests/automation/autofix/test_autofix_tasks.py
```diff
@@ -215,3 +218,53 @@ def test_check_and_mark_recent_autofix_runs(
         mock_logger.info.assert_any_call("Got 2 runs")
         mock_check_and_mark.assert_has_calls([call(mock_run1), call(mock_run2)])
         assert mock_check_and_mark.call_count == 2
+
+
+class TestDeleteOldAutofixRuns:
+    @patch("seer.automation.autofix.tasks.datetime")
```
corps, line 224: These tests are VERY Suspect.

What, is the semantic here?  There's so many mocks, it's hard to tell what's really going on.  I'm convinced most of these assertions are meaningless and fragile (who cares if you change the logger message?)

Feedback: Cursor is great, but if you don't constrain what you point cursor at, you get lots of text that doesn't contribute much.  Constrain it.  Focus it on *specific* semantic concerns, such as.

1.  Is actual data deleted?  The right data?
2.  Is data that should NOT be deleted left alone (for instance, create some data that occurs within the TTL and verify it doesn't go away)
3.  Does the batching logic execute but not perform an infinite loop?

I think.... both these tests are completely worthless.  They are technically a test, but they aren't covering meaningful failure modes.  Add more constraint, and use your brain :P

src/seer/automation/autofix/tasks.py
roaga: Added batching and a count metric

tests/automation/autofix/test_autofix_tasks.py
roaga: Re-worked the tests with my new knowledge of our testing environment, hopefully they're more useful now :)

tests/automation/autofix/test_autofix_tasks.py
```diff
@@ -215,3 +217,77 @@ def test_check_and_mark_recent_autofix_runs(
...
+        with Session() as session:
+            session.add(recent_run)
+            session.commit()
+
+        before_date = datetime.datetime.now() - datetime.timedelta(days=90)
+        deleted_count = delete_all_runs_before(before_date)
+
+        with Session() as session:
+            remaining_run = session.query(DbRunState).filter(DbRunState.id == recent_run.id).first()
+        assert remaining_run is not None
```
corps, line 254: Nit: can we assert that the remaining_run 's date is in fact within ttl?  Just to really lock in the semantic of the test.

tests/automation/autofix/test_autofix_tasks.py
corps: Looking much better!

tests/automation/autofix/test_autofix_tasks.py
roaga: Done in next commit

corps: Thank you!

corps (approved):

PR #1051, opened 2024-08-15
(issue summary) Improve issue summary prompt

roaga:
From limited testing:
1. Now we don't just reiterate minified JS errors
2. We don't just restate the error message in general
3. Added a new "headline" in the response that we can make use of
4. Verbosity and fluffy speech is reduced
5. Reduced overlap between headline and summary/impact so that they are all more valuable together


jennmueng (approved):

jennmueng:
An extra field on `SummarizeIssueResponse` shouldn't break the sentry side, :ship: it

PR #1057, opened 2024-08-16
(autofix/issue summary) Trim stacktrace frames

roaga:
For both Autofix and Issue Summary, limit the number of stacktrace frames we use to 16.
Prioritizes cutting out frames that aren't in-app.

Issue #1056

### DIFF
--- a/src/seer/automation/models.py
+++ b/src/seer/automation/models.py
@@ -333,7 +333,7 @@ class EventDetails(BaseModel):
                     breadcrumbs.append(crumb_details)
 
         return cls(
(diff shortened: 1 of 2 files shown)


roaga:
> I like prioritizing app frames, and LGTM, but could this logic be redundant with the `max_frames` we have in the `to_str` method?

For the way we use it now, it is redundant, but I think it's nice to maintain the distinction between the frames we actually store in our `Stacktrace` object and the ones we'd show as a string, as they could be different one day.

PR #1061, opened 2024-08-16
(issue summary) Remove confusing unknown_function value

roaga:
Issue #1060 

Just default to None if we don't have a function name, instead of making one up that confused the summarizer.


jennmueng (approved):
LGTM I'll leave you with the green check, but given how core this model is getting to both autofix and issue summary, should we add test coverage here?

PR #1063, opened 2024-08-16
feat(security): Add API bearer token auth via JWT

jennmueng:
Mainly for the staging deployment in `ml-ai`

Introduces an `ENFORCE_API_AUTH` env variable that will enforce either the rpc secret signing or the bearer token.

### DIFF
--- a/src/seer/configuration.py
+++ b/src/seer/configuration.py
@@ -54,7 +54,12 @@ class AppConfig(BaseModel):
     LANGFUSE_SECRET_KEY: str = ""
     LANGFUSE_HOST: str = ""
 
+    API_PUBLIC_KEY_SECRET_ID: str = ""
     JSON_API_SHARED_SECRETS: ParseList = Field(default_factory=list)
+    IGNORE_API_AUTH: ParseBool = False  # Used for both API Tokens and RPC Secrets
+
+    GOOGLE_CLOUD_PROJECT_ID: str = ""
+
     TORCH_NUM_THREADS: ParseInt = 0
     NO_SENTRY_INTEGRATION: ParseBool = False
     DEV: ParseBool = False
@@ -74,6 +79,11 @@ class AppConfig(BaseModel):
         return not self.NO_SENTRY_INTEGRATION
 
     def do_validation(self):
+        if self.IGNORE_API_AUTH:
+            assert (
+                self.JSON_API_SHARED_SECRETS or self.API_PUBLIC_KEY_SECRET_ID
+            ), "JSON_API_SHARED_SECRETS or API_PUBLIC_KEY_SECRET_ID required if IGNORE_API_AUTH is true!"
+
         if self.is_production:
             assert self.has_sentry_integration, "Sentry integration required for production mode."
             assert self.SENTRY_DSN, "SENTRY_DSN required for production!"
--- a/src/seer/configuration.py
+++ b/src/seer/configuration.py
@@ -79,10 +79,10 @@ class AppConfig(BaseModel):
         return not self.NO_SENTRY_INTEGRATION
 
     def do_validation(self):
-        if self.IGNORE_API_AUTH:
+        if not self.IGNORE_API_AUTH:
             assert (
                 self.JSON_API_SHARED_SECRETS or self.API_PUBLIC_KEY_SECRET_ID
-            ), "JSON_API_SHARED_SECRETS or API_PUBLIC_KEY_SECRET_ID required if IGNORE_API_AUTH is true!"
+            ), "JSON_API_SHARED_SECRETS or API_PUBLIC_KEY_SECRET_ID required if IGNORE_API_AUTH is false!"
 
         if self.is_production:
             assert self.has_sentry_integration, "Sentry integration required for production mode."
--- a/src/seer/json_api.py
+++ b/src/seer/json_api.py
@@ -5,21 +5,70 @@ import inspect
 import logging
 from typing import Any, Callable, Type, TypeVar, get_type_hints
 
+import jwt
 import sentry_sdk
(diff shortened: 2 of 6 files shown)


mdtro:
What is the plan for generating the JWT, just locally with some sort of management command?

src/seer/json_api.py
```diff
@@ -37,6 +41,28 @@ def wrapper() -> Any:
...
+
+                        # If the token is successfully decoded and verified,
+                        # the function will continue execution
+                    except jwt.ExpiredSignatureError:
+                        raise Unauthorized("Token has expired")
+                    except jwt.InvalidTokenError:
+                        raise Unauthorized("Invalid token")
+                except Exception as e:
+                    sentry_sdk.capture_exception(e)
+                    raise Unauthorized("Invalid Bearer token")
```
mdtro, line 63: It might be a good idea to handle the [`jwt.exceptions.InvalidSignatureError`](https://pyjwt.readthedocs.io/en/stable/api.html#jwt.exceptions.InvalidSignatureError) explicitly instead of catching it in the general Exception. 

I suspect we'll get a lot of garbage in the header and it'll make it difficult to differentiate between invalid JWTs and said junk.

src/seer/configuration.py
```diff
@@ -52,7 +52,10 @@ class AppConfig(BaseModel):
     LANGFUSE_SECRET_KEY: str = ""
     LANGFUSE_HOST: str = ""
 
+    API_PUBLIC_KEY: str = ""
```
mdtro, line 55: I'm not fully certain how you'll be loading the public key at runtime, but in general, I would recommend [using the cryptography package's `load_pem_public_key` function](https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/#cryptography.hazmat.primitives.serialization.load_pem_public_key).

It's easy to run into encoding/parsing errors when calling the decode function if the key isn't formatted just right.

Might save some headache in the future, especially if it's being uploaded to GCP's Secret Manager as a file.

src/seer/configuration.py
```diff
@@ -54,7 +54,12 @@ class AppConfig(BaseModel):
     LANGFUSE_SECRET_KEY: str = ""
     LANGFUSE_HOST: str = ""
 
+    API_PUBLIC_KEY_SECRET_ID: str = ""
     JSON_API_SHARED_SECRETS: ParseList = Field(default_factory=list)
+    ENFORCE_API_AUTH: ParseBool = False  # Used for both API Tokens and RPC Secrets
```
corps, line 59: I have a request on this: Can we invert this into `IGNORE_API_AUTH` and make it clear that we enforce API by default, excepting only when one explicitly ignores the api auth?

Basically, security is the default.  The exception is ignoring that security.  It also means far less changes in environment cariables.

src/seer/json_api.py
```diff
@@ -37,6 +87,33 @@ def wrapper() -> Any:
...
+                        # the function will continue execution
+                    except jwt.ExpiredSignatureError:
+                        raise Unauthorized("Token has expired")
+                    except jwt.InvalidSignatureError:
+                        raise Unauthorized("Invalid signature")
+                    except jwt.InvalidTokenError:
+                        raise Unauthorized("Invalid token")
+
+                except Exception as e:
+                    sentry_sdk.capture_exception(e)
```
corps, line 113: Hmm, I don't think this is wise.  If there is a wide exception, we should report that as a 500 instead of covering it with a `Unauthorized` response.

src/seer/json_api.py
```diff
@@ -61,8 +138,9 @@ def wrapper() -> Any:
     return decorator
 
 
-def get_json_api_shared_secrets() -> list[str]:
-    result = os.environ.get("JSON_API_SHARED_SECRETS", "").split()
+@inject
+def get_json_api_shared_secrets(config: AppConfig = injected) -> list[str]:
```
corps, line 142: I also am addressing this change in my PR https://github.com/getsentry/seer/pull/1074
Would it make sense for me to merge my change first, validate it, and then we can merge it into your PR and validate it?

tests/test_json_api.py
```diff
@@ -67,6 +74,76 @@ def my_endpoint(request: DummyRequest) -> DummyResponse:
...
+
+    app.register_blueprint(blueprint)
+
+    app_config = resolve(AppConfig)
+    app_config.ENFORCE_API_AUTH = True
+
+    pk = resolve(PublicKeyBytes)
+    pk.bytes = b"mock_public_key"
+
+    with patch("seer.json_api.jwt.decode") as mock_jwt_decode:
```
corps, line 94: This is fine for the unit testing case, can we also have atleast one test where we don't use the mock, we actually validate the jwt decoding will work with the right public key in place?  For instance, different jwt implementations can exist, and if we changed the jwt implementation, this test would continue to pass even if it could accidentally change how production tokens were handled.

src/seer/json_api.py
```diff
@@ -2,19 +2,69 @@
...
+
+    return public_key_bytes
+
+
+class PublicKeyBytes(BaseModel):
+    bytes: bytes | None
+
+
+@module.provider
+def provide_public_key(config: AppConfig = injected) -> PublicKeyBytes:
```
corps, line 46: 👍

src/seer/configuration.py
```diff
@@ -74,6 +79,11 @@ def has_sentry_integration(self) -> bool:
         return not self.NO_SENTRY_INTEGRATION
 
     def do_validation(self):
+        if self.IGNORE_API_AUTH:
```
corps, line 82: should this be `if not self.IGNORE_API_AUTH`?

corps (approved):

PR #1067, opened 2024-08-21
feat(autofix): Reproduction in root causes

jennmueng:
1. Introduces reproduction step in root cause output.
2. Uses gpt4o w/ structured output for the root cause for increased reliability
3. (Tacking on) Adds claude agent back into the plan+code agent

Screenshot 2024-09-10 at 10 45 47 AM

### Eval results:
(Before is above, after is below)
Screenshot 2024-09-10 at 10 48 23 AM

### DIFF
--- a/src/seer/automation/agent/client.py
+++ b/src/seer/automation/agent/client.py
@@ -138,6 +138,19 @@ class GptClient(LlmClient):
     ) -> tuple[dict[str, Any] | None, Message, Usage]:
         return super().json_completion(messages, model, system_prompt)
 
+    def clean_tool_call_assistant_messages(self, messages: list[Message]) -> list[Message]:
+        new_messages = []
+        for message in messages:
+            if message.role == "assistant" and message.tool_calls:
+                new_messages.append(
+                    Message(role="assistant", content=message.content, tool_calls=[])
+                )
+            elif message.role == "tool":
+                new_messages.append(Message(role="user", content=message.content, tool_calls=[]))
+            else:
+                new_messages.append(message)
+        return new_messages
+
 
 class ClaudeClient(LlmClient):
     @inject
--- a/src/seer/automation/autofix/components/root_cause/component.py
+++ b/src/seer/automation/autofix/components/root_cause/component.py
@@ -1,19 +1,21 @@
 import logging
+import textwrap
 
 from langfuse.decorators import observe
 from sentry_sdk.ai.monitoring import ai_track
 
 from seer.automation.agent.agent import AgentConfig, GptAgent
+from seer.automation.agent.client import GptClient
+from seer.automation.agent.models import Message
 from seer.automation.autofix.autofix_context import AutofixContext
 from seer.automation.autofix.components.root_cause.models import (
+    MultipleRootCauseAnalysisOutputPrompt,
     RootCauseAnalysisOutput,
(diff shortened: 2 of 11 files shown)


src/seer/automation/autofix/components/root_cause/component.py
```diff
@@ -35,56 +38,81 @@ def invoke(self, request: RootCauseAnalysisRequest) -> RootCauseAnalysisOutput |
...
+                textwrap.dedent(
+                    """\
+                    Given all the above potential root causes you just gave, please provide a 1-2 sentence concise instruction on how to reproduce the issue for each root cause.
+                    - Assume the user is an experienced developer well-versed in the codebase, simply give the reproduction steps.
+                    - You must use the local variables provided to you in the stacktrace to give your reproduction steps.
+                    - Try to be open ended to allow for the most flexibility in reproducing the issue. Avoid being too confident.
+                    - This step is optional, if you're not sure about the reproduction steps for a root cause, just skip it."""
+                )
+            )
+
```
roaga, line 70: Wanna make this a helper in a different file to keep it clean and consistent with our other prompts/steps?

src/seer/automation/autofix/components/root_cause/component.py
```diff
@@ -35,56 +38,81 @@ def invoke(self, request: RootCauseAnalysisRequest) -> RootCauseAnalysisOutput |
...
+                        new_messages.append(
+                            Message(role="assistant", content=message.content, tool_calls=[])
+                        )
+                    elif message.role == "tool":
+                        new_messages.append(
+                            Message(role="user", content=message.content, tool_calls=[])
+                        )
+                    else:
+                        new_messages.append(message)
+                return new_messages
```
roaga, line 84: Couldn't this be cleanly integrated into the client classes instead? That way the client class always cleans up its messages and any other part of the code doesn't have to think about it

roaga (commented):
Just some thoughts on code organization/cleanliness

src/seer/automation/autofix/components/root_cause/component.py
jennmueng: Yeah originally had it here cuz thought it was gonna be temporary and fixed by now lol

src/seer/automation/agent/client.py
```diff
@@ -138,6 +138,19 @@ def json_completion(
...
+            if message.role == "assistant" and message.tool_calls:
+                new_messages.append(
+                    Message(role="assistant", content=message.content, tool_calls=[])
+                )
+            elif message.role == "tool":
+                new_messages.append(Message(role="user", content=message.content, tool_calls=[]))
+            else:
+                new_messages.append(message)
+        return new_messages
+
```
roaga, line 153: On this I meant not just as a helper function, but if any messages passed into the client for any completion can be cleaned behind-the-scenes to prevent the errors. That way external code doesn't have to choose whether or not to clean.

roaga (approved):

src/seer/automation/agent/client.py
jennmueng: I don't think that's a good idea, given that openai can expect a tool call and its response to be used in a specific way for the LLM input. We should only use this workaround sparingly, in this case, only for the formatter.

PR #1069, opened 2024-08-23
Minor setup fixes

armenzg:


.envrc
```diff
@@ -7,6 +7,7 @@ if ! command -v "devenv" >/dev/null; then
 Try out with an install from:
 https://github.com/getsentry/devenv#install
 '
+  exit 1
```
armenzg, line 10: If `devenv` is not installed we should abort the execution.

README.md
```diff
@@ -23,7 +23,7 @@ These instructions require access to internal Sentry resources, and are intended
 1. Clone the repository and navigate to the project root
 2. Run `direnv allow` to set up the Python environment
 3. Create a `.env` file based on `.env.example` and set the required values
```
armenzg, line 25: There should be a section explaining how to set up all of the values in the file.

README.md
```diff
@@ -23,7 +23,7 @@ These instructions require access to internal Sentry resources, and are intended
...
-4. (Optional) Add `export SENTRY_AUTH_TOKEN=<your token>` to your shell's RC file
+4. (Optional) Add `SENTRY_AUTH_TOKEN=<your token>` to your `.env` file
```
armenzg, line 26: `direnv` reads values from `.env` and exports them.

README.md
jennmueng: Now that we have more and more people setting up seer, seems like we should get this done soon

jennmueng (approved):

PR #1074, opened 2024-08-26
Require api signatures for json_api

corps:
In production, ensure that api signatures are provided, and reject requests that lack it.


jennmueng (approved):
LGTM :shipit:

PR #1081, opened 2024-08-28
Anomaly detection - peak/trough detection and caching of matrix profle

ram-senth:
This PR has multiple changes:
- Trough and peak detection using interquartile range
- Caching of matrix profile
- Changes to the detect end point to support historic time steps separated from current time steps of a time series


src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -60,27 +59,70 @@ def _online_detect(
...
+            anomaly_algo_data=streamed_anomalies.get_anomaly_algo_data(len(ts_external))[0],
+        )
         # TODO: Clean up old data
-        return ts_external
+        return ts_external, streamed_anomalies
+
+    @inject
+    @sentry_sdk.trace
+    def _combo_detect(
+        self, ts_with_history: TimeSeriesWithHistory, config: AnomalyDetectionConfig
```
jennmueng, line 92: Missing a `= injected` for DI?

ram-senth: No, the config param is part of request payload.

ram-senth: I do need to remove the `@inject` above.

src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -92,30 +134,60 @@ def _update_anomalies(self, ts_external: List[TimeSeriesPoint], anomalies: TimeS
...
-            self._online_detect(request.context)
+        transaction_name = (
+            "Stream AD for alert"
             if isinstance(request.context, AlertInSeer)
-            else self._batch_detect(request.context)
+            else (
+                "Stream AD for timeseries with history"
+                if isinstance(request.context, TimeSeriesWithHistory)
+                else "Batch AD for timeseries"
+            )
```
aayush-se, line 144: Nit: This would probably be more clear as an elif...else rather than a ternary

aayush-se, line 144: nit: nested ternary makes this section somewhat hard to follow. Could this be done using a switch or an if/else?

ram-senth: Sure, will change it.

src/seer/anomaly_detection/anomaly_detection.py
ram-senth: Done

src/seer/anomaly_detection/anomaly_detection.py
ram-senth: Done

tests/seer/anomaly_detection/test_accessors.py
trillville, line 1: will there be a follow up PR for the peak/trough logic?

src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -29,22 +27,23 @@
...
-        anomalies = batch_detector.detect(convert_external_ts_to_internal(timeseries))
-        self._update_anomalies(timeseries, anomalies)
-        return timeseries
+        batch_detector = MPBatchAnomalyDetector()
+        anomalies = batch_detector.detect(convert_external_ts_to_internal(timeseries), config)
+        return timeseries, anomalies
 
     @inject
     @sentry_sdk.trace
     def _online_detect(
```
trillville, line 41: i think it would be good to add some brief docstrings to these to clarify the context in which each is used

tests/seer/anomaly_detection/test_accessors.py
ram-senth: Yes I will tweak the peak/trough logic some more based on your experiments in a follow-up PR. Wanted to get this out as frontend work is dependent on the API change.

src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -92,30 +137,60 @@ def _update_anomalies(self, ts_external: List[TimeSeriesPoint], anomalies: TimeS
...
-        return DetectAnomaliesResponse(timeseries=ts)
+
+        if isinstance(request.context, AlertInSeer):
+            transaction_name = "Stream AD for alert"
+        elif isinstance(request.context, TimeSeriesWithHistory):
+            transaction_name = "Stream AD for timeseries with history"
+        else:
+            transaction_name = "Batch AD for timeseries"
+
+        with sentry_sdk.start_transaction(op="task", name=transaction_name):
```
trillville, line 148: any particular reason why we want a separate transaction here?

src/seer/anomaly_detection/detectors/mp_scorers.py
```diff
@@ -62,17 +86,89 @@ def score(
...
         if np.isnan(mp_dist):
             return "none"
         if mp_dist < threshold_lower:
             return "none"
         if mp_dist < threshold_upper:
             return "anomaly_lower_confidence"
         return "anomaly_higher_confidence"
+
+    def _adjust_flag_for_vicinity(
+        self, flag: AnomalyFlags, ts_value: float, context: npt.NDArray[np.float64]
```
trillville, line 162: this logic is pretty difficult to follow without a docstring - specifically what is `context` and how should i think about it? (i think i know the answer from talking to you but will probably be confused looking at this in 2 months)

trillville (approved):
code LGTM - do we feel confident about the peak/trough logic?

src/seer/anomaly_detection/anomaly_detection.py
ram-senth: Done

src/seer/anomaly_detection/anomaly_detection.py
ram-senth: This one endpoint caters to three different types of anomaly detection - stateless batch, stateless online and stateful alert based. Each serve a different use case and will have different performance characteristics. So I think it is worth tracking them as different transactions with separate names. Any potential problem with this approach?

src/seer/anomaly_detection/detectors/mp_scorers.py
ram-senth: Done

ram-senth:
> code LGTM - do we feel confident about the peak/trough logic?

The logic that is going in is incrementally better than what we had previously. It does need more tweaking. I will be following up on it immediately and have an updated version out before the internal release.

Issue #1071, opened 2024-08-26
(issue summary) Make use of traces


trillville:
Really we mean _trace connected issues_. Send their event data to Seer too so we can summarize it.
- summarize the original issue
- summarize the trace relationship as a 3rd bucket under the summary output. Tell them if they need to look at another issue and link them there.

PR #1088, opened 2024-08-28
feat(issue summary): Add support for trace connected issues

roaga:
Included in this PR:
1. Optionally takes in trace connected issues for an issue to inform the summary output. (#1071 )
2. Attempts to trim `vars` in a stacktrace frame if they get lengthy.
3. Fixes `ValidationError` for exceptions without a stacktrace by making the stacktrace optional. (#1084 )

### DIFF
--- a/src/seer/automation/models.py
+++ b/src/seer/automation/models.py
@@ -14,6 +14,7 @@ from pydantic import (
     ConfigDict,
     Field,
     ValidationError,
+    ValidationInfo,
     field_validator,
 )
 from pydantic.alias_generators import to_camel, to_snake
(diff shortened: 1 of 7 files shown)


src/seer/automation/models.py
```diff
@@ -42,6 +42,25 @@ class StacktraceFrame(BaseModel):
...
 
+    @field_validator("vars", mode="before")
+    @classmethod
+    def validate_vars(cls, vars: Optional[dict[str, Any]]):
+        if not vars or len(vars.keys()) <= 4:
+            return vars
+        return cls._trim_vars(vars)
+
+    @staticmethod
+    def _trim_vars(vars: dict[str, Any]):
```
jennmueng, line 53: Would trimming these potentially remove valuable context from a bug? Or are these types of vars useless?

src/seer/automation/models.py
```diff
@@ -256,7 +275,7 @@ class SentryEventData(TypedDict):
 class ExceptionDetails(BaseModel):
     type: str
     value: Optional[str] = ""
-    stacktrace: Stacktrace
+    stacktrace: Optional[Stacktrace] = None
```
jennmueng, line 278: 👍

jennmueng (approved):

src/seer/automation/models.py
roaga: I haven't seen an example where they are useful, and they are usually useless, but as always, any trimming runs that risk. Same thing with trimming stacktrace frames.

jennmueng (approved):

PR #1118, opened 2024-09-04
fix(anomaly detection): handle overflow for SuSS algorithm

aayush-se:
- Set up upper bound to avoid `window_size` overflow when hitting `_score()` (upper limit is np.integer max size which is 2**63 - 1)
- Write basic unit tests to confirm functionality

### DIFF
--- a/src/seer/anomaly_detection/detectors/window_size_selectors.py
+++ b/src/seer/anomaly_detection/detectors/window_size_selectors.py
@@ -99,6 +99,7 @@ class SuSSWindowSizeSelector(WindowSizeSelector):
         exp = 0
         found_window = False
 
+        # exponential search (to find window size interval) with limit to avoid overflow
         while 2**exp < np.iinfo(np.integer).max:
             window_size = 2**exp
 
--- a/src/seer/anomaly_detection/detectors/window_size_selectors.py
+++ b/src/seer/anomaly_detection/detectors/window_size_selectors.py
@@ -90,7 +90,7 @@ class SuSSWindowSizeSelector(WindowSizeSelector):
         ts_min_max = np.max(time_series) - np.min(time_series)
 
         lbound = self.lbound
-        ubound = len(time_series) // 3 + 1  # Limit to 1/3 of total data
+        ubound = len(time_series)
         threshold = self.threshold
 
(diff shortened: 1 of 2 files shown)


src/seer/anomaly_detection/detectors/window_size_selectors.py
```diff
@@ -97,10 +97,10 @@ def optimal_window_size(self, time_series: npt.NDArray[np.float64]) -> int:
         min_score = self._score(time_series, time_series.shape[0] - 1, stats)
 
         exp = 0
-
-        # exponential search (to find window size interval)
         found_window = False
-        for _ in range(10000):
+
+        # exponential search (to find window size interval) with limit to avoid overflow
+        while 2**exp < np.iinfo(np.integer).max:
```
ram-senth, line 103: It maybe worth keeping both checks - value as well as iteration count. What do you think?

aayush-se: I think that is probably a good idea was just worried about hardcoding values.

Also, what are your thoughts on limiting the window size to the max number of samples possible? (24 * 30 * 60) = 43200 as the upper bound for the window_size?

ram-senth: To be frank, we can limit it to 1/3 of the length of the input time series a we already have a check for at least seven days of data.

aayush-se: Agreed with the 1/3 of timeseries length will set that.

aayush-se: Done

ram-senth: I'm thinking about it a bit more. If the window size goes above the range temporarily then we are ok with that, right? Having the loop terminate if window hits a high value may be an issue.

aayush-se: I think in this case if the window size goes above the range, it should be okay as it would still be under the limit of the length of the input time series (I can change my code to implement that as the true upper bound). I added a case on line 124 to set the upper bound to limit it before it proceeds into the binary search section

PR #1123, opened 2024-09-05
use gcp project env vars

jennmueng:


roaga (approved):
If it works locally, LGTM

PR #1133, opened 2024-09-06
feat(codegen): Support different types of runs in run state

jennmueng:
1. Adds a type field to the `DbRunState` to allow for different types of runs other than Autofix
2. Moves the TTL logic out to `/automation` to account for this.

This PR is in preparation for a new `codegen` folder with unit test generation.

Migration & state setting/calling tested locally.


roaga (commented):
Why put two different features/products in the same table?

jennmueng:
@roaga They use the same state-machine-like functionality with the steps, making it less code to manage when they're in the same table, can easily be told apart via the `type` key

roaga (approved):

PR #1135, opened 2024-09-06
feat(codegen): Unit test generation init

jennmueng:
Introduces a new `codegen` folder with a unit test generation agent that was made for the codecov hack week project

Uses a new `CodegenContext` similar to the autofix context but new

image

Lack of unit tests is due to this being a WIP agent, will follow up & harden


roaga (approved):
Let's make sure to add test coverage soon.

Also I'm wondering if anything can be abstracted out to be shared between Autofix and Unit Test Gen (without coupling the two together), since both agents look very similar in a lot of ways.

jennmueng:
@roaga Agreed on all the points, just wanted this code in so it doesn't become stale

PR #1136, opened 2024-09-06
fix(anomaly-detection): Handle illegal padding amount

aayush-se:
- Util raises exception if requested length for padding is less than current mp_dist array

Note: is there a more graceful way to handle this?

### DIFF
--- a/src/seer/anomaly_detection/detectors/mp_utils.py
+++ b/src/seer/anomaly_detection/detectors/mp_utils.py
@@ -42,6 +42,8 @@ class MPUtils(BaseModel):
             mp_dist = normalizer.normalize(mp_dist)
 
         if pad_to_len is not None:
+            if pad_to_len - len(mp_dist) < 0:
+                raise Exception("Requested length should be greater than current mp_dist")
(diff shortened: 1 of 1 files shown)


src/seer/anomaly_detection/detectors/mp_utils.py
```diff
@@ -42,6 +42,8 @@ def get_mp_dist_from_mp(
             mp_dist = normalizer.normalize(mp_dist)
 
         if pad_to_len is not None:
+            if pad_to_len - len(mp_dist) < 0:
```
ram-senth, line 45: Minor nit:  `len(mp_dist) <= pad_to_len`  would be a little bit more readable?

aayush-se: I dont think it should be equal since that just means no padding is added

ram-senth: Yes you are right. So we should raise error if pad_to_len is less than or equal to len(mp_len).

ram-senth (approved):

PR #1140, opened 2024-09-07
(dynamic alert thresholds) Force recompilation of Stumpy library during bootup

ram-senth:
Use mode preloading mechanism to force numba compilation for Stumpy library

### DIFF
--- a/src/seer/anomaly_detection/anomaly_detection.py
+++ b/src/seer/anomaly_detection/anomaly_detection.py
@@ -1,7 +1,9 @@
 import logging
 from typing import List, Tuple
 
+import numpy as np
 import sentry_sdk
+import stumpy  # type: ignore # mypy throws "missing library stubs"
 from pydantic import BaseModel
 
 from seer.anomaly_detection.accessors import AlertDataAccessor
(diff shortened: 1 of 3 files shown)


src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -27,6 +33,29 @@
...
+    @sentry_sdk.trace
+    def __init__(self):
+        """
+        Force Stumpy compilation by making a dummy call to the library.
+
+        Note: compilation triggered here is very specific to the exact values of parameters ignore_trivial, normalize etc. A
+            future call with different values for one or more parameter will still trigger a recompilation.
+        """
+        data = np.arange(10.0)
+        try:
```
trillville, line 45: does this windowsizeselector logic need to be here?

trillville (approved):
lgtm - how long does this take? do we need to change anything on the ops side WRT bootup logic to accommodate?

src/seer/anomaly_detection/anomaly_detection.py
ram-senth: That is a good point, it doesn't have to be. I was experimenting and forgot to remove that.

ram-senth:
> lgtm - how long does this take? do we need to change anything on the ops side WRT bootup logic to accommodate?

Interestingly, it doesn't take much time actually, less than 100 ms. Seems like it is a function of the size of the time series used in the first call.

PR #1142, opened 2024-09-09
chore(auth): Exclude URL from auth check as it can differ between environments

ram-senth:
By using URL as part of the payload that is signed we are assuming that URL used by client will be same as the URL seen in the request payload on Seer end. This is not true as requests from Subscription processor has full UTL including server name while those from the UI do not.

This PR makes URL optional. This will be step 1 of the fix. We need to follow this up with two more changes:
* remove URL from all locations on the client end where we are signing Seer requests
* once we confirm that Seer is not getting requests with URL anymore, then remove the code that checks for URL + payload.

### DIFF
--- a/src/seer/json_api.py
+++ b/src/seer/json_api.py
@@ -88,6 +88,7 @@ def json_api(blueprint: Blueprint, url_rule: str) -> Callable[[_F], _F]:
             # Optional for now during rollout, make this required after rollout.
             if auth_header.startswith("Rpcsignature "):
                 parts = auth_header.split()
+                sentry_sdk.metrics.incr(key="rpc_signature_auth_attempt", value=1)
                 if len(parts) != 2 or not compare_signature(request.url, raw_data, parts[1]):
                     raise Unauthorized(
                         f"Rpcsignature did not match for given url {request.url} and data"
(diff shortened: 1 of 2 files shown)


src/seer/json_api.py
```diff
@@ -163,18 +173,28 @@ def compare_signature(url: str, body: bytes, signature: str, config: AppConfig =
...
-    )
-
     for key in secrets:
-        computed = hmac.new(key.encode(), signature_input, hashlib.sha256).hexdigest()
-        is_valid = hmac.compare_digest(computed.encode(), signature_data.encode())
-        if is_valid:
+        if nonce:
+            if is_valid(b"%s:%s" % (nonce.encode(), body), key, signature_data):
+                sentry_sdk.metrics.incr(
+                    key="rpc_signature_auth_success", value=1, tags={"with_url": False}
```
ram-senth, line 180: Change the key and value to `{"additional_payload": "nonce"}`? Similarly change line 193 to `{"additional_payload": "url"}`?

ram-senth: I can also make the change if you are ok with it.

src/seer/json_api.py
```diff
@@ -88,7 +88,10 @@ def wrapper(config: AppConfig = injected) -> Any:
             # Optional for now during rollout, make this required after rollout.
             if auth_header.startswith("Rpcsignature "):
                 parts = auth_header.split()
-                if len(parts) != 2 or not compare_signature(request.url, raw_data, parts[1]):
+                sentry_sdk.metrics.incr(key="rpc_signature_auth_attempt", value=1)
```
trillville, line 91: metrics API is being deprecated, so you would want to switch it out for a span + span metric (e.g. https://github.com/getsentry/seer/blob/f393fef55f0446a98e16e081239459998aef0ec4/src/seer/app.py#L108-L109)

you can also do this to reference the current span https://github.com/getsentry/seer/blob/f393fef55f0446a98e16e081239459998aef0ec4/src/seer/grouping/grouping.py#L312

trillville (approved):

src/seer/json_api.py
ram-senth: Have removed use of metrics API and changed the key-value pairs.

src/seer/json_api.py
ram-senth: Used the second option to update existing span.

PR #1146, opened 2024-09-10
Anomaly Detection: Unit and Integration Tests

aayush-se:
- Generated Synthetic timeseries (cyclical, cyclical with anomaly, noisy, noisy with anomaly) to use for unit and integration tests
  - Timeseries generated in`data-analysis` GitHub repo

- Wrote unit tests for the following modules in anomaly_detection:
  - Detectors
    - `anomaly_detectors.py`
    - `mp_config.py`
    - `mp_scorers.py`
    - `mp_utils.py`
    - `normalizers.py`
    - `window_size_selctors.py`
  - Models
    - `converters.py`
    - `timeseries_anomalies.py`
    - `timeseries.py`

Wrote integration tests for the `store` and `detect` endpoints for batch, stream, and combo pipelines

**NOTE:** Wanted to submit this PR to start getting some feedback as there are a lot of changes. As of submitting this PR, the following unit/integration tests do not work and I wanted to get some feedback:
- `test_detect_anomalies_online` in `test_anomaly_detection.py`: Was unsure how to handle database changes during integration test
- `test_detect` in `test_anomaly_detectors.py`: was running into shape errors with the matrix profile
- `test_incorrect_padding` in `test_mp_utils.py`: this one is fixed as of [this PR](https://github.com/getsentry/seer/pull/1136) so needs to be merged into main to reflect changes


tests/seer/anomaly_detection/detectors/test_anomaly_detectors.py
```diff
@@ -0,0 +1,110 @@
...
+        )
+        self.config = AnomalyDetectionConfig(
+            time_period=15, sensitivity="low", direction="up", expected_seasonality="auto"
+        )  # TODO: Placeholder values as not used in detection yet
+
+    @patch("stumpy.stumpi")
+    @patch("seer.anomaly_detection.detectors.MPScorer")
+    @patch("seer.anomaly_detection.detectors.MPUtils")
+    def test_detect(self, MockMPUtils, MockMPScorer, MockStumpi):
+        mock_stream = MagicMock()
```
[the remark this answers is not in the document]
aayush-se: Good call out as magic mock (and the mock/patch module in general) doesn't have those default checks.

tests/seer/anomaly_detection/detectors/test_anomaly_detectors.py
```diff
@@ -0,0 +1,123 @@
...
+        mock_stream.left_I_ = [[0], [1], [2]]
+        mock_stream.T_ = np.array([1.1, 2.1, 3.1])
+
+        mock_utils.get_mp_dist_from_mp.return_value = np.array([0.1, 0.2])
+
+        mock_scorer.stream_score.return_value = ([0.5], ["none"])
+
+        anomalies = self.detector.detect(self.timeseries, self.config, mock_scorer, mock_utils)
+
+        self.assertIsInstance(anomalies, MPTimeSeriesAnomalies)
```
corps, line 113: Nit: Use `assert isinstance(anomaies, ...)` as opposed to using any of the `self.assert` methods.  The reason is that modern `pytest` has much better output, debugging support for `assert x` than the old unittest `self.assertX` methods.

tests/seer/anomaly_detection/detectors/test_anomaly_detectors.py
```diff
@@ -0,0 +1,123 @@
...
+        self.assertIsInstance(anomalies.flags, list)
+        self.assertIsInstance(anomalies.scores, list)
+        self.assertIsInstance(anomalies.matrix_profile, np.ndarray)
+        self.assertIsInstance(anomalies.window_size, int)
+
+        self.assertEqual(len(anomalies.scores), 3)
```
corps, line 119: Same, convert these to the pytest equivalents (`assert len(anomalies.scores)) == 3`)

tests/seer/anomaly_detection/detectors/test_anomaly_detectors.py
```diff
@@ -0,0 +1,123 @@
...
+        self.assertEqual(len(anomalies.flags), 3)
+        self.assertEqual(len(anomalies.matrix_profile), 3)
+        mock_scorer.stream_score.assert_called_once()
```
corps, line 122: The mock methods can still use the `assert_called_x` variety since there is no equivalent `assert x` pytest method.

tests/seer/anomaly_detection/detectors/test_mp_utils.py
```diff
@@ -0,0 +1,89 @@
...
+    def test_no_normalizer_with_normalization(self):
+        mp = np.array([[1.0], [2.0], [3.0], [4.0]])
+        self.mock_mp_config.normalize_mp = True
+
+        with self.assertRaises(Exception) as context:
+            self.utils.get_mp_dist_from_mp(
+                mp, pad_to_len=None, mp_config=self.mock_mp_config, normalizer=None
+            )
+
+        self.assertTrue("Need normalizer to normalize MP" in str(context.exception))
```
corps, line 74: this too can become `assert "Need normalizer" in str(context.exception)`

tests/seer/anomaly_detection/test_anomaly_detection.py
```diff
@@ -0,0 +1,139 @@
...
+        timeseries, _, _ = convert_synthetic_ts(
+            "tests/seer/anomaly_detection/test_data/synthetic_series", as_ts_datatype=True
+        )
+
+        for i, ts in enumerate(timeseries):
+
+            alert = AlertInSeer(id=i)
+
+            request = StoreDataRequest(
+                organization_id=i,
```
corps, line 36: Do these values matter in any way?  Is it reasonable to just organization_id and project_id to 0?

It's just nice to know which parts of the input data matter to the test assertion.  If these values are "throw away" for the purposes of our assertions, preferably use a static value to signal that.  But if they need to be different for some reason, this is fine.

aayush-se: My thought process for including it was to have the `store` test run with the corresponding `alert.id` first so that when the online detection test runs, it could pull from those corresponding ids in the db.

Probably have to pull @ram-senth into this since I think he has better context on how to test online detection and if these are necessary for these integration tests.

aayush-se:
Reverted changes -- improperly merged from main. Will updated changes from comments again.

tests/seer/anomaly_detection/test_anomaly_detection.py
aayush-se: Update: I'll replace it with a static value since these values will end up being thrown away.

ram-senth (approved):

PR #1148, opened 2024-09-10
Disable signature auth

corps:


src/seer/json_api.py
```diff
@@ -82,19 +82,21 @@ def decorator(
...
-            elif auth_header.startswith("Bearer "):
+                #     parts = auth_header.split()
+                #     if len(parts) != 2 or not compare_signature(
+                #         request.url, request.args.get("nonce", ""), raw_data, parts[1]
+                #     ):
+                #         raise Unauthorized(
+                #             f"Rpcsignature did not match for given url {request.url} and data"
+                #         )
+                pass
+            if auth_header.startswith("Bearer "):
```
corps, line 99: Will remove this later once we determine our plan for codecov stuff, but didn't see any point removing it atm.

jennmueng: we should change the condition with `config.IGNORE_API_AUTH` too, cuz right now if we merge this into prod it will hit that case and error

corps: done

ram-senth (approved):

jennmueng (approved):

PR #1157, opened 2024-09-12
Adding ability to push self hoisting staging docker compose serice

corps:
Creates a self hoisting docker-compose image and enables command to push that to a target sandbox for staging deployment.


Dockerfile
```diff
@@ -31,18 +31,17 @@ RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.11 1 &
...
-COPY setup.py requirements.txt celeryworker.sh celerybeat.sh gunicorn.sh ./
-
-RUN chmod +x ./celeryworker.sh ./celerybeat.sh ./gunicorn.sh
-
 # Install dependencies
+COPY setup.py requirements.txt ./
 RUN pip install --upgrade pip==24.0
 RUN pip install -r requirements.txt --no-cache-dir
 
+# Copy model files (assuming they are in the 'models' directory)
```
corps, line 39: Moving these makes rebuilding images faster, since changes to models or the worker files doesn't require reinstalling pip

Dockerfile
```diff
@@ -55,8 +54,11 @@ COPY supervisord.conf /etc/supervisord.conf
 RUN pip install --default-timeout=120 -e . --no-cache-dir --no-deps
 
 ENV FLASK_APP=src.seer.app:start_app()
-# Set in cloudbuild.yaml for production images
+
+# Supports sentry releases
 ARG SEER_VERSION_SHA
 ENV SEER_VERSION_SHA ${SEER_VERSION_SHA}
+ARG SENTRY_ENVIRONMENT=production
```
corps, line 61: Used to adjust the environment flag of sentry sent errors, allowing us to distinguish stuff coming from staging in the same seer project.

ram-senth: Do we need to add SENTRY_ENVIRONMENT to .env which will be different in staging and local dev?

ram-senth (approved):

Dockerfile
corps: We shouldn't have to change anything locally.  When you want to run a staging environment, the `push-staging` pushes this value into the image with `staging` set, but for everyone else it is "production".  Production is fine because locally we won't all share the same sentry project, only when we're sharing DSNs do we need to differentiate.

PR #1168, opened 2024-09-16
feat(autofix): Add first pass at interactive flow

roaga:
This supports insight card generation, user interjection, and a root cause flow that makes sense for the new design.

### DIFF
--- a/src/seer/automation/agent/agent.py
+++ b/src/seer/automation/agent/agent.py
@@ -63,11 +63,25 @@ class LlmAgent(ABC):
             system_prompt=self.config.system_prompt if self.config.system_prompt else None,
             tools=(self.tools if len(self.tools) > 0 else None),
         )
+    
+    def use_user_messages(self, context: AutofixContext):
+        # adds any queued user messages to the memory
+        user_msgs = context.state.get().steps[-1].queued_user_messages
+        if user_msgs:
+            self.memory.append(Message(content="\n".join(user_msgs), role="user"))
+            with context.state.update() as cur:
+                cur.steps[-1].queued_user_messages = []
+            context.event_manager.add_log("Thanks for the input. I'm thinking through it now...")
 
     def run_iteration(self, context: Optional[AutofixContext] = None):
         logger.debug(f"----[{self.name}] Running Iteration {self.iterations}----")
-
+        
         message, usage = self.get_completion()
+
+        # interrupt if user message is queued and awaiting handling
+        if context and context.state.get().steps[-1].queued_user_messages:
+            self.use_user_messages(context)
+            return
 
         self.memory.append(message)
 
@@ -75,7 +89,6 @@ class LlmAgent(ABC):
         if message.content and context:
             text_before_tag = message.content.split("<")[0]
             text = text_before_tag
-            print("Asking for insight" if text else "Not asking for insight")
             if text:
                 # call LLM separately with the same memory to generate structured output insight cards 
                 insight_sharing = InsightSharingComponent(context)
@@ -88,13 +101,8 @@ class LlmAgent(ABC):
                         past_insights=past_insights
                     )
                 )
-                print("insight step run")
                 if insight_card:
-                    print("insight card generated")
-                    if insight_card.is_unimportant_insight or insight_card.repeats_existing_idea or insight_card.is_incomplete_idea:
-                        print(f"Shouldn't share: {insight_card.insight}")
-                    elif context.state.get().steps and isinstance(context.state.get().steps[-1], DefaultStep):
-                        print("updating step with new insight card")
+                    if context.state.get().steps and isinstance(context.state.get().steps[-1], DefaultStep):
                         step = cast(DefaultStep, context.state.get().steps[-1])
                         step.insights.append(insight_card)
                         with context.state.update() as cur:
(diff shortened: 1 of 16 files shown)


src/seer/automation/agent/agent.py
```diff
@@ -61,26 +63,51 @@ def get_completion(self):
...
+                    )
+                )
+                if insight_card:
+                    if context.state.get().steps and isinstance(
+                        context.state.get().steps[-1], DefaultStep
+                    ):
+                        step = cast(DefaultStep, context.state.get().steps[-1])
+                        step.insights.append(insight_card)
+                        with context.state.update() as cur:
+                            cur.steps[-1] = step
```
jennmueng, line 110: Hmm I think this is worth pulling out of the general agent logic and into a specific `AutofixAgent` that inherits from this generic `Agent` as codecov will be using this too and we don't need to generate insights there.

jennmueng (commented):
1 note, will take a second pass later after long interview

src/seer/automation/agent/agent.py
roaga: Since in many cases we'll want to disable it for Autofix too (e.g. evals, GH Copilot), I'm adding an `interactive` flag to the AgentConfig so that these kinds of features are easy to enable/disable whenever we want.

src/seer/automation/autofix/models.py
```diff
@@ -121,6 +122,13 @@ class BaseStep(BaseModel):
     progress: list["ProgressItem | Step"] = Field(default_factory=list)
     completedMessage: Optional[str] = None
 
+    queued_user_messages: list[str] = []
+
+    def receive_user_message(self, message: str):
+        self.queued_user_messages.append(message)
+        print("HELLO user message")
```
jennmueng, line 129: leftover log?

src/seer/automation/autofix/components/insight_sharing/component.py
```diff
@@ -0,0 +1,134 @@
...
+                total_tokens=completion.usage.total_tokens,
+            )
+            cur.usage += usage
+        structured_message = completion.choices[0].message
+        if structured_message.refusal:
+            raise RuntimeError(structured_message.refusal)
+        if not structured_message.parsed:
+            raise RuntimeError("Failed to parse message")
+
+        res = completion.choices[0].message.parsed
```
jennmueng, line 124: You can use `extract_parsed_model`: https://github.com/getsentry/seer/blob/23fa0e508c6306d5cc6126d943c991e585101457/src/seer/automation/utils.py#L236
to not have to repeat this logic

src/seer/automation/agent/client.py
```diff
@@ -147,6 +147,10 @@ def clean_tool_call_assistant_messages(self, messages: list[Message]) -> list[Me
                 )
             elif message.role == "tool":
                 new_messages.append(Message(role="user", content=message.content, tool_calls=[]))
+            elif message.role == "tool_use":
+                new_messages.append(
+                    Message(role="assistant", content=message.content, tool_calls=[])
+                )
```
jennmueng, line 153: `tool_use` is for anthropic right, we don't need to use this hack on anthropic calls right?

jennmueng (approved):
couple comments, nothing major, lgtm otherwise

src/seer/automation/agent/client.py
roaga: I believe I did need to use this hack for Claude, I can't recall the exact error that was occurring though

roaga:
👋 Hi there! Here is a root cause analysis of issue SEER-5Q automatically generated by Autofix 🤖

This analysis was triggered by Jenn Mueng

# Missing Validation for `namespace_id` Before Insert Operation

## Description
The `ForeignKeyViolation` error occurs because the `namespace_id` being inserted into the `codebase_namespace_mutex` table does not exist in the `codebase_namespaces` table. This issue arises due to the absence of validation or creation logic for `namespace_id` before the insert operation.

The `namespace_id` is part of the `CodebaseState` class, but it is not explicitly set or validated during the initialization of `CodebaseState` instances. Additionally, there is no direct assignment or validation logic for `namespace_id` in the codebase.

To resolve this issue, ensure that the `namespace_id` is validated to exist in the `codebase_namespaces` table before attempting the insert operation into the `codebase_namespace_mutex` table.

## Reproduction Steps
Attempt to save a `CodebaseNamespaceManager` instance with a `namespace_id` that does not exist in the `codebase_namespaces` table, using the provided `namespace_id` value of `88`.

## Relevant Code Context

### CodebaseState Class Definition
The `CodebaseState` class includes a `namespace_id` attribute but lacks explicit validation or assignment logic.

**File:** src/seer/automation/autofix/models.py
```
class CodebaseState(BaseModel):
    repo_id: int | None = None
    namespace_id: int | None = None
    repo_external_id: str | None = None
    file_changes: list[FileChange] = []
```

### CodebaseState Initialization
The `CodebaseState` class is instantiated without explicitly setting the `namespace_id`.

**File:** src/seer/automation/autofix/autofix_context.py
```
with state.update() as cur:
    for repo in request.repos:
        if repo.external_id not in cur.codebases:
            cur.codebases[repo.external_id] = CodebaseState(
                file_changes=[],
                repo_external_id=repo.external_id,
            )
```

## More Info

If you have any questions or feedback for the Sentry team about this fix, please email [autofix@sentry.io](mailto:autofix@sentry.io) with the Run ID (see below).

### 🤓 Stats for the nerds:
Run ID: **38**
Prompt tokens: **86864**
Completion tokens: **5848**
Total tokens: **92712**

PR #1169, opened 2024-09-17
[feat] Add codecov client 

rohitvinnakota-codecov:
This PR updates automated test generation by adding a couple of code coverage related pieces of information. 

We fetch this information by querying 2 public APIs

1.) https://docs.codecov.com/reference/repos_pulls_retrieve
2.) https://docs.codecov.com/reference/repos_test_results_retrieve

These queries are defined in a new class `CodecovClient`, and are called in `src/seer/automation/codegen/unit_test_coding_component.py` so that we can add that context to the prompt. 

We only send additional coverage data if they are relevant.


src/integrations/codecov/codecov_client.py
```diff
@@ -0,0 +1,37 @@
+import requests
+
+
+CODECOV_TOKEN = 'FETCH FROM ENV'
```
rohitvinnakota-codecov, line 4: This will eventually be an environment variable once it's configured.

src/integrations/codecov/codecov_client.py
```diff
@@ -0,0 +1,37 @@
+import requests
+
+
+CODECOV_TOKEN = "FETCH FROM ENV"
```
rohitvinnakota-codecov, line 4: This will eventually be an environment variable.

jennmueng: You can get it from the env via AppConfig, like how we get the github tokens here: https://github.com/getsentry/seer/blob/63347dcc12afb044af0f94058e6031851dc0ebad/src/seer/automation/codebase/repo_client.py#L51

Just need to decorate the method with `@inject`, set the default value of the argument to `= injected`, and define the property in `AppConfig` and it will just read those defined properties from the env

jennmueng (commented):
yep, this looks good to me, left a comment on how to set up dependency injection for the token

src/integrations/codecov/codecov_client.py
```diff
@@ -0,0 +1,36 @@
+import requests
+
+from seer.configuration import AppConfig
+from seer.dependency_injection import inject, injected
+
+class CodecovClient:
```
corps, line 6: This is totes fine for now, but in the future, using instance methods makes it easier to potentially subclass for test mock purposes.

```
@dataclasses.dataclass
class CodecovClient
    config: AppConfig = injected
    ...

class MockCodecovClient(CodecovClient):
    ...

```

This is a minor nit that doesn't need to be addresses here, but food for thought on future design.

src/seer/automation/codegen/models.py
```diff
@@ -45,7 +46,7 @@ def mark_updated(self):
 
 class CodeUnitTestRequest(BaseComponentRequest):
     diff: str
-
+    codecov_client_params: Dict[Any, Any]
```
corps, line 49: We should add a default here so that inflight requests from older clients don't fail.

As a general principle, all of our request objects should be backwards compatible -- that is to say, if we add new fields, it should have a reasonable default in all cases.

```
codecov_client_params: dict = Field(default_factory=dict)
```

jennmueng (approved):
LGTM, codecov is failing though 🤣

PR #1184, opened 2024-09-19
chore(similarity): Add try catch around insert grouping record

jangjodi:
Add `on_conflict_do_nothing` for inserting group records

fixes [SEER-9Y](https://sentry.sentry.io/issues/5754951463/)

### DIFF
--- a/src/seer/grouping/grouping.py
+++ b/src/seer/grouping/grouping.py
@@ -11,6 +11,7 @@ import torch
 from pydantic import BaseModel, ValidationInfo, field_validator
 from sentence_transformers import SentenceTransformer
 from sqlalchemy.dialects.postgresql import insert
+from sqlalchemy.exc import IntegrityError
 from torch.cuda import OutOfMemoryError
 
 from seer.db import DbGroupingRecord, Session
@@ -477,7 +478,6 @@ class GroupingLookup:
     ) -> None:
         """
         Inserts a new GroupingRecord into the database if the group_hash does not already exist.
-        If new grouping record was created, return the id.
 
         :param session: The database session.
         :param issue: The issue to insert as a new GroupingRecord.
@@ -489,6 +489,12 @@ class GroupingLookup:
             .first()
         )
 
+        extra = {
+            "project_id": issue.project_id,
+            "stacktrace_length": len(issue.stacktrace),
+            "input_hash": issue.hash,
+        }
+
         if existing_record is None:
             new_record = GroupingRecord(
                 project_id=issue.project_id,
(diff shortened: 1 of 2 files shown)


src/seer/grouping/grouping.py
```diff
@@ -498,15 +504,26 @@ def insert_new_grouping_record(
                 error_type=issue.exception_type,
             ).to_db_model()
             session.add(new_record)
+
+            try:
+                session.flush()
```
JoshFerge, line 509: why flush?

jangjodi: I initially thought we didn't want to rollback the entire session, but that was incorrect. I created a new session to rollback, but would it also work if I kept the session parameter and did rollback since there were no other adds?

src/seer/grouping/grouping.py
```diff
@@ -472,43 +472,58 @@ def insert_batch_grouping_records(
...
+                .first()
             )
 
+            extra = {
+                "project_id": issue.project_id,
+                "stacktrace_length": len(issue.stacktrace),
+                "input_hash": issue.hash,
+            }
+
+            if existing_record is None:
```
JoshFerge, line 496: can we modify this to return early if existing record is not none so we don't get so much arrowing? https://blog.codinghorror.com/flattening-arrow-code/

src/seer/grouping/grouping.py
```diff
@@ -472,43 +472,58 @@ def insert_batch_grouping_records(
...
+                    stacktrace_embedding=embedding,
+                    hash=issue.hash,
+                    error_type=issue.exception_type,
+                ).to_db_model()
+                session.add(new_record)
+
+                try:
+                    session.commit()
+                except IntegrityError:
+                    session.rollback()
```
JoshFerge, line 509: what does rollback do? do we need it here?

src/seer/grouping/grouping.py
```diff
@@ -472,43 +472,58 @@ def insert_batch_grouping_records(
...
+                    existing_record = (
+                        session.query(DbGroupingRecord)
+                        .filter_by(hash=issue.hash, project_id=issue.project_id)
+                        .first()
+                    )
+                    extra["existing_hash"] = existing_record.hash
+                    logger.info(
+                        "group_already_exists_in_seer_db",
+                        extra=extra,
+                    )
```
JoshFerge, line 519: i'm wondering if this is necessary or if we can skip not having this log -- in theory we should be able to see the group hashes linked on the sentry side, right?

JoshFerge (commented):
any way to add a test that enters this logic for this case?

jangjodi:
@JoshFerge I updated this PR to use on_conflict_do_nothing, instead of logging and doing the try catch. Let me know if that makes sense!

JoshFerge (approved):
makes sense to me!

PR #1191, opened 2024-09-23
feat(autofix): Store root cause memory

### LINKED ISSUE #1187: (autofix): Save agent memory and reuse

jennmueng:
Allows the memory to of an agent to easily be stored and read via the `AutofixContext`, right now we store only the root cause agent memory.

Validated working via running locally & checking db with datagrip:
CleanShot 2024-09-23 at 12 07 06@2x

Resolves #1187


roaga (approved):
LGTM

PR #1192, opened 2024-09-24
Adds per deployment beat configuration

corps:
1.  Abstract the notion of deployment questions into the `AppConfig` object with new properties.  NOTE:  Deployments are not considered discrete, they can overlap, as per local development.
2.  Check for deployment type either by explicit environment variable or by the hostname (confirmed the deployment name in our k8s config reflects in production host names).
3.  Conditionalize beat tasks by deployments.
4.  Tests.  Add your new beat job to the anomaly detect test to validate it becomes a condiational beat task.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -49,6 +49,7 @@ services:
             - .env
         environment:
             - DEV=1
+            - AUTOFIX_ENABLED=1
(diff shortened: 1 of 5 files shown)


aayush-se (approved):
Code is clear for the questions asked. Looks great!

PR #1193, opened 2024-09-24
Anomaly detection/data pruning

aayush-se:
- Create Celery task to delete data that is over 28 days old and update matrix profiles for remaining points
- Write unit tests for new accessor methods and cleanup task


tests/seer/anomaly_detection/test_cleanup_tasks.py
```diff
@@ -0,0 +1,62 @@
...
+            project_id=project_id,
+            external_alert_id=external_alert_id,
+            config=config,
+            timeseries=timeseries,
+            anomalies=anomalies,
+            anomaly_algo_data={"window_size": anomalies.window_size},
+            data_purge_flag="not_queued",
+        )
+
+        # Confirm if points are being deleted and matrix profile recalculated after cleanup task is called
```
corps, line 52: Solid test setup.
Could we also sample some amount of the resulting, remaining time series objects, validating that the cut (those deleted vs note) is shaped like we hope?

corps (approved):

corps:
Before we turn this on in production with ENABLE_CELERY_WORKERs, lets get https://github.com/getsentry/seer/pull/1192 in and add move this beat task to a conditional for anomaly detection.

tests/seer/anomaly_detection/test_cleanup_tasks.py
aayush-se: I agree on adding that.

@ram-senth do you have any suggestions about this since I think the shape should be standard (using the stumpy library for matrix profiling and a getter from MpTimeSeriesAnomalies to convert it to a dictionary), however, I was unsure on if and how I could make further checks for the remaining data since it didn't seem clear to me on how to compare the older and newer values in this case.

corps: Any reasonable signal is fine, it doesn't have to deep assertion.  You can even do a trivial comparison.  "new" = "not one of the old ones" by some signal.

corps: For instance, you can set just one data point up in such a way as that it produces a clear signal about it being either new or old, and then just test that.

src/seer/anomaly_detection/tasks.py
```diff
@@ -0,0 +1,128 @@
...
+                timestamp=datetime.fromtimestamp(point.timestamp),
+                value=point.value,
+                anomaly_type=anomalies.flags[i],
+                anomaly_score=anomalies.scores[i],
+                anomaly_algo_data=algo_data[i],
+            )
+            for i, point in enumerate(ts)
+        ]
+        dynamic_alert.anomaly_algo_data = {"window_size": anomalies.window_size}
+        session.commit()
```
aayush-se, line 106: Currently working on a fix for this as it is not functioning as intended

aayush-se: Updated!

src/seer/anomaly_detection/tasks.py
```diff
@@ -0,0 +1,134 @@
...
+        algo_data = anomalies.get_anomaly_algo_data(len(ts))
+
+        # Update mp values (algo_data) by iterating through respective points by timestamp
+        for timestamp, algo_data_point in zip(ts, algo_data):
+            point = (
+                session.query(DbDynamicAlertTimeSeries)
+                .filter_by(timestamp=datetime.fromtimestamp(timestamp.timestamp))
+                .first()
+            )
+            point.anomaly_algo_data = algo_data_point
```
aayush-se, line 103: Would like to hear suggestions to optimizing this since I couldn't perform a bulk/batch update on it since I don't have the indexes of the rows I wanted to update. SQLAlchemy only supports bulk update through primary key (https://docs.sqlalchemy.org/en/20/orm/queryguide/dml.html#orm-bulk-update-by-primary-key)

tests/seer/anomaly_detection/test_cleanup_tasks.py
aayush-se: Just added a few more tests for shape and some value comparisons. Great callout since the original work this was testing was incorrect and needed a fix.

src/seer/anomaly_detection/tasks.py
```diff
@@ -0,0 +1,97 @@
...
+from seer.anomaly_detection.models.external import AnomalyDetectionConfig
+from seer.anomaly_detection.models.timeseries import TimeSeries
+from seer.db import DbDynamicAlert, Session
+
+logger = logging.getLogger(__name__)
+
+
+@celery_app.task
+@sentry_sdk.trace
+def cleanup_timeseries(alert_id: int, date_threshold: float, anomaly_detection_config: dict):
```
ram-senth, line 18: Actually we do not need to pass anomaly_detection_config in as a parameter as it is already stored in the db in the `config` column.

tests/seer/anomaly_detection/test_cleanup_tasks.py
```diff
@@ -0,0 +1,137 @@
...
+
+from seer.anomaly_detection.accessors import DbAlertDataAccessor
+from seer.anomaly_detection.detectors.anomaly_detectors import MPBatchAnomalyDetector
+from seer.anomaly_detection.models.external import AnomalyDetectionConfig, TimeSeriesPoint
+from seer.anomaly_detection.models.timeseries import TimeSeries
+from seer.anomaly_detection.tasks import cleanup_timeseries
+from seer.db import DbDynamicAlert, DbDynamicAlertTimeSeries, Session
+
+
+class TestCleanupTasks(unittest.TestCase):
```
ram-senth, line 15: A little indirectly related - I think we need an unit test against anomaly detection to check what happens if a stream detection call is received for an alert for which we have deleted all time steps.

tests/seer/anomaly_detection/test_cleanup_tasks.py
```diff
@@ -28,38 +26,82 @@ def test_cleanup_timeseries(self):
...
+        external_alert_id, config, _, _ = self._save_alert(0, 0)
+        date_threshold = (datetime.now() - timedelta(days=28)).timestamp()
+        cleanup_timeseries(external_alert_id, date_threshold, config.serialize())
+
+    def test_only_old_points_deleted(self):
+        # Create and save alert with 1000 points (all old)
+        external_alert_id, config, points, anomalies = self._save_alert(1000, 0)
+        date_threshold = (datetime.now() - timedelta(days=28)).timestamp()
+        cleanup_timeseries(external_alert_id, date_threshold, config.serialize())
+
```
ram-senth, line 80: I missed adding some asserts on the alert data in table. Mind adding that in?

aayush-se:
> We also need some metrics published from the Celerey task so that we can keep track of:
> 
> * number of clean ups done in a time window
> * time duration taken for each cleanup so we can monitor p90 or p99
> * count of times when a cleanup results in the alert having less than 7 days of data
> * number of times cleanup task  encountered error
> 
> These are in addition to the regular queue related metrics like queue size, time spent waiting etc. We should check with Jenn if we need to do anything specific for those to work.

I did include Sentry tracing for the cleanup task so I think everything aside from the third point (less than 7 days of data) should be viewable. I'm unsure about the other queue related metrics though since Sentry doesn't support direct metrics anymore (https://docs.sentry.io/product/explore/metrics/).

src/seer/db.py
```diff
@@ -286,9 +289,13 @@ class DbDynamicAlert(Base):
         cascade="all, delete, delete-orphan",
         passive_deletes=True,
     )
-    data_purge_flag: Mapped[str] = mapped_column(String, nullable=False)
-    queued_at: Mapped[datetime.date] = mapped_column(
-        DateTime, nullable=False, default=datetime.datetime.utcnow
+    data_purge_flag: Mapped[TaskStatus] = mapped_column(
+        Enum("not_queued", "processing", "queued", name="data_purge_flag_enum"),
```
ram-senth, line 293: If we change TaskStatus  to an enum then we can use it here instead.

src/seer/db.py
```diff
@@ -58,6 +59,8 @@ class Base(DeclarativeBase):
 migrate = Migrate(directory="src/migrations")
 Session = sessionmaker(autoflush=False, expire_on_commit=False)
 
+TaskStatus = Literal["not_queued", "processing", "queued"]
```
ram-senth, line 62: We can convert this to string enum and use it in both the business logic as well as in the data layer.

src/seer/anomaly_detection/models/external.py
```diff
@@ -50,7 +50,6 @@ def serialize(self) -> dict:
             "sensitivity": self.sensitivity,
             "direction": self.direction,
             "expected_seasonality": self.expected_seasonality,
-            "model_config": self.model_config,
```
ram-senth: We can remove this whole serialize method as it is not needed.

ram-senth:
> > We also need some metrics published from the Celerey task so that we can keep track of:
> > 
> > * number of clean ups done in a time window
> > * time duration taken for each cleanup so we can monitor p90 or p99
> > * count of times when a cleanup results in the alert having less than 7 days of data
> > * number of times cleanup task  encountered error
> > 
> > These are in addition to the regular queue related metrics like queue size, time spent waiting etc. We should check with Jenn if we need to do anything specific for those to work.
> 
> I did include Sentry tracing for the cleanup task so I think everything aside from the third point (less than 7 days of data) should be viewable. I'm unsure about the other queue related metrics though since Sentry doesn't support direct metrics anymore (https://docs.sentry.io/product/explore/metrics/).

Wondering if we should add the alert id as a tag so that we can look at metrics at an alert level. Let us ask Zach and Jenn about the celery metrics in the next standup. Rest is good.

src/migrations/versions/62facff812cc_migration.py
```diff
@@ -0,0 +1,41 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("dynamic_alerts", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column(
+                "data_purge_flag",
+                sa.Enum("NOT_QUEUED", "PROCESSING", "QUEUED", name="taskstatus", native_enum=False),
```
ram-senth, line 25: It seems like python enums can be directly used as SqlAlchemy enums: 
```
class MyEnum(enum.Enum):
    one = 1
    two = 2
    three = 3

class MyClass(Base):
    __tablename__ = 'some_table'
    id = Column(Integer, primary_key=True)
    value = Column(Enum(MyEnum))
```

ram-senth (approved):

PR #1195, opened 2024-09-25
GRPC Server in Seer

corps:
Note, in this case I left the implementation "empty".  Goal here is to have all the infrastructure in place so that add services is implementing the service interface + adding to server.

There's an unfortunate amount of scaffolding to get a grpc server + health + reflection running, but afterwards each new service is a much smaller footprint (just one line in the setup servers).

Also notably, a big part of this experiment is the XDS credentials line, which may give us the authentication and tls like behavior via XDS hooks in gcloud.

Next goal is to add some tests.


supervisord.conf
```diff
@@ -16,6 +16,18 @@ stdout_logfile_maxbytes=0
 stderr_logfile=/dev/stderr
 stderr_logfile_maxbytes=0
 
+; The grpc server is disabled by default. Set GRPC_SERVER_ENABLE=true in the environment to enable it.
```
corps, line 19: Goal would be to have separated pods running this so that scaling and debugging them is more trivial.  Also not a problem with our networking -- we can ensure that the xds host is unique for json vs grpc, making a whole class of mistakes go away.

src/seer/grpc.py
```diff
@@ -0,0 +1,171 @@
...
+    server.add_secure_port(f"0.0.0.0:{config.GRPC_SERVICE_PORT}", server_creds)
+
+    maintenance_server = ReflectableServer(
+        grpc.server(futures.ThreadPoolExecutor(config.GRPC_THREAD_POOL_SIZE))
+    )
+    logger.info(f"Adding listener for maintenance port {config.GRPC_MAINTENANCE_PORT}")
+    maintenance_server.add_insecure_port(f"0.0.0.0:{config.GRPC_MAINTENANCE_PORT}")
+
+    # Add new services right here.
+    summarize_pb2_grpc.add_IssueSummaryServiceServicer_to_server(issue_summary, server)
```
corps, line 140: Each new service should just be this sort of line based on the codegen file.  In a near future may be worth extracting all the surrounding junk into common libraries.

src/seer/grpc.py
```diff
@@ -0,0 +1,171 @@
...
+    server_fallback_creds = grpc.insecure_server_credentials()
+    server_creds = grpc.xds_server_credentials(server_fallback_creds)
+    logger.info(f"Adding listener for service port {config.GRPC_SERVICE_PORT}")
+    server.add_secure_port(f"0.0.0.0:{config.GRPC_SERVICE_PORT}", server_creds)
+
+    maintenance_server = ReflectableServer(
+        grpc.server(futures.ThreadPoolExecutor(config.GRPC_THREAD_POOL_SIZE))
+    )
+    logger.info(f"Adding listener for maintenance port {config.GRPC_MAINTENANCE_PORT}")
+    maintenance_server.add_insecure_port(f"0.0.0.0:{config.GRPC_MAINTENANCE_PORT}")
```
markstory, line 137: Is the intent here to expose a secondary port that `grpcurl` and related tools can use?

src/seer/grpc.py
```diff
@@ -0,0 +1,171 @@
...
+    config: AppConfig = injected,
+    issue_summary: summarize_pb2_grpc.IssueSummaryServiceServicer = injected,
+    health_servicer: HealthServicer = injected,
+) -> list[grpc.Server]:
+    server = ReflectableServer(
+        grpc.server(futures.ThreadPoolExecutor(config.GRPC_THREAD_POOL_SIZE))
+    )
+
+    server_fallback_creds = grpc.insecure_server_credentials()
+    server_creds = grpc.xds_server_credentials(server_fallback_creds)
```
markstory, line 129: I didn't find a ton of information on how grpc + xds works. I did find [this](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md) though, is that what I should read to learn more about xds + grpc?

src/seer/grpc.py
markstory: One line per service seems fair to me.

src/seer/grpc.py
dmajere: @markstory https://cloud.google.com/service-mesh/docs/service-routing/proxyless-overview
we're trying to test this.

markstory (approved):
Looks good to me.

PR #1198, opened 2024-09-26
feat(autofix+copilot): Add support for copilot to trigger autofix

roaga:
1. If a PR url is given in the start request, Autofix will comment on that PR with the analysis result.
2. If the coding step is called on that same run, Autofix will automatically create a PR when done.

Since this work has a tight deadline, and is highly experimental, no tests were added for this. But all existing functionality still works as expected.


src/seer/automation/autofix/autofix_context.py
```diff
@@ -298,6 +298,51 @@ def commit_changes(self, repo_external_id: str | None = None, repo_id: int | Non
...
                             session.commit()
 
+    def comment_root_cause_on_pr(
+        self, pr_url: str, repo_definition: RepoDefinition, root_cause: str
+    ):
+        # make root cause into markdown string
+        state = self.state.get()
+        markdown_comment = textwrap.dedent(
+            """\
+            👋 Hi there! Here is a root cause analysis of {issue} automatically generated by Autofix 🤖
```
trillville, line 308: we'll prob want to tweak this copy for the demo but this is fine for now

trillville (approved):

PR #1199, opened 2024-09-26
fix(autofix): PlanCode snippet and new file quick fixes

jennmueng:
Two quick but impactful fixes for plan+code
1. Some root causes don't have a code snippet now, that was causing a lot of the errors
2. It often likes to create files that already exist, this is a preliminary prompting fix, will have more robust follow up fix

Reduces error rate in plan+code, but not all just yet

Before:
CleanShot 2024-09-26 at 11 31 50
After:
CleanShot 2024-09-26 at 11 29 48


roaga (approved):
Let's hope that 1 point score drop is just flakiness

PR #1201, opened 2024-09-26
feat(autofix): Add Q&A interaction

roaga:
Gives Autofix the option to ask the user a question, and continue analysis once they respond.

Also adds a way to disable interactive features when requesting Autofix to start (good for evals, copilot)

### DIFF
--- a/src/seer/automation/autofix/components/coding/component.py
+++ b/src/seer/automation/autofix/components/coding/component.py
@@ -62,7 +62,7 @@ class CodingComponent(BaseComponent[CodingRequest, CodingOutput]):
                     instruction=request.instruction,
                 )
                 if not request.initial_memory
-                else ""
+                else None
             ),
             context=self.context,
         )
--- a/src/seer/automation/autofix/components/coding/component.py
+++ b/src/seer/automation/autofix/components/coding/component.py
@@ -41,6 +41,7 @@ class CodingComponent(BaseComponent[CodingRequest, CodingOutput]):
         agent = ClaudeAgent(
             tools=tools.get_tools(),
             config=AgentConfig(system_prompt=CodingPrompts.format_system_msg(), interactive=True),
+            memory=request.initial_memory,
         )
 
         task_str = (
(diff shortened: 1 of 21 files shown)


src/seer/automation/autofix/models.py
```diff
@@ -42,6 +42,7 @@ class AutofixStatus(enum.Enum):
     PROCESSING = "PROCESSING"
     NEED_MORE_INFORMATION = "NEED_MORE_INFORMATION"
     CANCELLED = "CANCELLED"
+    WAITING_FOR_USER_RESPONSE = "WAITING_FOR_USER_RESPONSE"
```
jennmueng, line 45: Do we need both `NEED_MORE_INFORMATION` and `WAITING_FOR_USER_RESPONSE`?

src/seer/automation/autofix/components/coding/component.py
```diff
@@ -52,12 +53,16 @@ def invoke(self, request: CodingRequest) -> CodingOutput | None:
...
+            (
+                CodingPrompts.format_fix_discovery_msg(
+                    event=request.event_details.format_event(),
+                    task_str=task_str,
+                    summary=request.summary,
+                    repo_names=[repo.full_name for repo in state.request.repos],
+                    instruction=request.instruction,
+                )
+                if not request.initial_memory
+                else ""
```
jennmueng, line 65: Shouldn't be passing an empty string here but simply just pass `None`, this way we wouldn't need the `if content` in `add_user_message` and instead just skip adding a user message when the run argument is none

jennmueng (commented):
quick first pass, will take a look again in a sec after standup

src/seer/automation/autofix/models.py
roaga: `NEED_MORE_INFORMATION` is for root cause selection, so I think they should be separate as the behavior once you get a response is very different

src/seer/automation/autofix/components/root_cause/component.py
```diff
@@ -31,31 +31,35 @@ def invoke(
...
-                    repo_names=[repo.full_name for repo in state.request.repos],
+                (
+                    RootCauseAnalysisPrompts.format_default_msg(
+                        event=request.event_details.format_event(),
+                        summary=request.summary,
+                        instruction=request.instruction,
+                        repo_names=[repo.full_name for repo in state.request.repos],
+                    )
+                    if not request.initial_memory
+                    else ""
```
jennmueng, line 56: same as last comment above about the run call

jennmueng (approved):
only that one point otherwise lgtm

src/seer/automation/autofix/components/root_cause/component.py
roaga: Done

src/seer/automation/autofix/components/coding/component.py
roaga: Done

PR #1204, opened 2024-09-27
fix(similarity): Add try catch around insert in grouping records

jangjodi:
Add try catch and logging for integrity error in insert grouping records

### DIFF
--- a/src/seer/grouping/grouping.py
+++ b/src/seer/grouping/grouping.py
@@ -489,7 +489,7 @@ class GroupingLookup:
                     )
                 )
                 session.commit()
-            except IntegrityError as e:
+            except IntegrityError:
                 logger.info(
                     "group_already_exists_in_seer_db",
                     extra={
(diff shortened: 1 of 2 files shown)


src/seer/grouping/grouping.py
```diff
@@ -481,12 +482,22 @@ def insert_new_grouping_record(self, issue: GroupingRequest, embedding: np.ndarr
...
-                    index_elements=(DbGroupingRecord.project_id, DbGroupingRecord.hash)
+            try:
+                session.execute(
+                    insert_stmt.on_conflict_do_nothing(
+                        index_elements=(DbGroupingRecord.project_id, DbGroupingRecord.hash)
+                    )
+                )
+                session.commit()
+            except IntegrityError:
+                logger.info(
```
JoshFerge, line 493: wanna logger.exception so it stays in sentry?

PR #1206, opened 2024-09-27
feat(autofix): Plan+Code file handling and diff fixer

jennmueng:
Makes plan+code agent much more reliable with
1. Handle incorrect file operations (edit file that doesn't exist, create file that already exists, etc)
2. Fix incorrect diffs

We now have a 0% plan/code error rate and much lower missing change rate

Before:
CleanShot 2024-09-27 at 10 53 22@2x
After:
CleanShot 2024-09-27 at 10 50 51@2x

### DIFF
--- a/src/seer/automation/autofix/components/root_cause/component.py
+++ b/src/seer/automation/autofix/components/root_cause/component.py
@@ -36,7 +36,7 @@ class RootCauseAnalysisComponent(BaseComponent[RootCauseAnalysisRequest, RootCau
             config=AgentConfig(
                 system_prompt=RootCauseAnalysisPrompts.format_system_msg(),
                 max_iterations=24,
-                interactive=True,
+                interactive=False,
             ),
             memory=request.initial_memory,
         )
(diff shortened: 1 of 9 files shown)


src/seer/automation/autofix/components/root_cause/component.py
```diff
@@ -36,7 +36,7 @@ def invoke(
             config=AgentConfig(
                 system_prompt=RootCauseAnalysisPrompts.format_system_msg(),
                 max_iterations=24,
-                interactive=True,
+                interactive=False,
             ),
             memory=request.initial_memory,
```
roaga, line 41: Remnant of local testing or...?

src/seer/automation/autofix/utils.py
```diff
@@ -1,11 +1,15 @@
...
 import re
+from functools import lru_cache
+
+from langfuse.decorators import observe
 
 VALID_BRANCH_NAME_CHARS = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ-"
 
 
-def compute_similarity(text1: str, text2: str, ignore_whitespace=True) -> float:
+@lru_cache(maxsize=1024)
```
roaga, line 11: fancy

roaga (approved):
One callout before you merge, otherwise looks good.

Though I don't fully understand the switch to a cached similarity if you could explain that

src/seer/automation/autofix/components/root_cause/component.py
jennmueng: oh yeah

src/seer/automation/autofix/utils.py
jennmueng: we'll be using the similarity multiple times across the retries so this is to optimize that

PR #1207, opened 2024-09-29
(autofix): Generate insight cards on separate thread async

roaga:
Before, insight cards would be generated on every iteration synchronously, slowing down the whole experience. Now, they're generated in a separate thread, so it's a lot faster.

Also, a smattering of prompt engineering is included in this PR.

### DIFF
--- a/src/seer/automation/autofix/components/insight_sharing/component.py
+++ b/src/seer/automation/autofix/components/insight_sharing/component.py
@@ -14,7 +14,6 @@ from seer.automation.autofix.components.insight_sharing.models import (
 )
 from seer.automation.component import BaseComponent
 from seer.automation.utils import extract_parsed_model
-from seer.dependency_injection import inject, injected
 
 
 class InsightSharingPrompts:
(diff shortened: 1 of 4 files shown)


src/seer/automation/autofix/components/insight_sharing/component.py
```diff
@@ -63,10 +62,8 @@ class InsightSharingComponent(BaseComponent[InsightSharingRequest, InsightSharin
 
     @observe(name="Sharing Insights")
     @ai_track(description="Sharing Insights")
-    @inject
-    def invoke(
-        self, request: InsightSharingRequest, gpt_client: GptClient = injected
-    ) -> InsightSharingOutput | None:
+    def invoke(self, request: InsightSharingRequest) -> InsightSharingOutput | None:
```
roaga, line 65: I had to remove dependency injection here because Zach's module was breaking under threading for some reason

jennmueng: Ah, so you'll need to do
```
from seer.bootup import module
from seer.configuration import configuration_module

module.enable()
configuration_module.enable()
```
inside the thread, so basically at the top of`share_insights`

src/seer/automation/autofix/components/insight_sharing/component.py
```diff
@@ -98,7 +95,7 @@ def invoke(
                 latest_thought=request.latest_thought,
             )
             memory = []
-            for i, message in enumerate(
+            for _, message in enumerate(
```
jennmueng, line 98: do we still need the enumerate here?

jennmueng (approved):
one comment, and replied to the other comment with how to fix dep injection, here's an approve so you're unblocked

src/seer/automation/autofix/components/insight_sharing/component.py
roaga: Removed in next commit

PR #1211, opened 2024-10-01
feat(autofix): Add unit test to root cause output

jennmueng:
Adds a simple unit test that reproduces (intentionally failing) to the root cause output. The goal for this is to easily be copied.

CleanShot 2024-10-01 at 14 59 26@2x

### DIFF
--- a/src/seer/automation/autofix/components/root_cause/models.py
+++ b/src/seer/automation/autofix/components/root_cause/models.py
@@ -1,9 +1,10 @@
 from typing import Annotated, Optional
 
-from pydantic import BaseModel, StringConstraints
+from pydantic import BaseModel, StringConstraints, field_validator
 from pydantic_xml import attr
 
 from seer.automation.agent.models import Message
+from seer.automation.autofix.utils import remove_code_backticks
 from seer.automation.component import BaseComponentOutput, BaseComponentRequest
 from seer.automation.models import EventDetails, PromptXmlModel
 from seer.automation.summarize.issue import IssueSummary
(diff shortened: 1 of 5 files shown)


src/seer/automation/autofix/components/root_cause/models.py
```diff
@@ -36,30 +37,58 @@ class RootCauseRelevantContext(BaseModel):
...
 
-class RootCauseAnalysisRelevantContext(BaseModel):
-    snippets: list[RootCauseRelevantContext]
-
-
 class RootCauseAnalysisItemPrompt(BaseModel):
     title: str
     description: str
-    reproduction: str
+    reproduction_instructions: str | None = None
```
roaga, line 73: Just making sure, will this break the frontend? i.e. does the frontend require something named `reproduction`?

src/seer/automation/autofix/components/root_cause/prompts.py
```diff
@@ -69,8 +69,10 @@ def root_cause_formatter_msg():
     def reproduction_prompt_msg():
         return textwrap.dedent(
             """\
-            Given all the above potential root causes you just gave, please provide a 1-2 sentence concise instruction on how to reproduce the issue for each root cause.
-            - Assume the user is an experienced developer well-versed in the codebase, simply give the reproduction steps.
+            Given all the above potential root causes you just gave, please provide 1-2 sentence instructions on how to reproduce the issue, then turn it into a concise unit test that reproduces the issue for each root cause. Make sure you follow any existing testing framework and patterns.
+            - This test should intentionally fail if the issue is not fixed. It will pass if the issue is fixed.
```
roaga, line 73: I think the wording "that reproduces the issue" could be changed. That's probably why the generated test expects the issue to occur.

roaga (approved):

src/seer/automation/autofix/components/root_cause/models.py
jennmueng: this is the only for prompting, this isn't the model used to store/send to frontend

PR #1212, opened 2024-10-01
[fix] Increase Anthropic max output size and fix test generation output

rohitvinnakota-codecov:
The response from Claude was being truncated due to the smaller token size. 

This PR also fixes an issue where the incorrect type was being passed into `CodeUnitTestOutput`. It should only take in a list of `FileChange` types.


jennmueng (approved):
If this works without throwing errors, then LGTM. I thought Vertex Anthropic only supported 4096 tokens

PR #1216, opened 2024-10-02
chore(anomaly-detection): support direction flag

ram-senth:
This PR takes the configured directionality in consideration during anomaly detection.


src/seer/anomaly_detection/detectors/mp_scorers.py
```diff
@@ -268,37 +272,6 @@ def _to_flag(self, mp_dist: np.float64, threshold: float):
             return "none"
         return "anomaly_higher_confidence"
 
-    def _adjust_flag_for_vicinity(
```
ram-senth, line 271: Removed this for now. Will add prophet based logic in next PR.

src/seer/automation/codegen/unit_test_github_pr_creator.py
```diff
@@ -1,10 +1,10 @@
 import logging
 import time
-from typing import List
+
+from github.PullRequest import PullRequest
```
ram-senth, line 4: This is unrelated to anomaly detection but `pre-commit` failed when I rebased on main.

src/seer/automation/codegen/unit_test_github_pr_creator.py
```diff
@@ -1,10 +1,10 @@
...
 
 from seer.automation.codebase.repo_client import RepoClient
 from seer.automation.models import FileChange
-from github.PullRequest import PullRequest
```
ram-senth, line 7: This is unrelated to anomaly detection but `pre-commit` failed when I rebased on main.

src/seer/automation/codegen/unittest_step.py
```diff
@@ -1,7 +1,6 @@
 from typing import Any
 
 from langfuse.decorators import observe
-from seer.automation.codegen.unit_test_github_pr_creator import GeneratedTestsPullRequestCreator
```
ram-senth, line 4: This is unrelated to anomaly detection but `pre-commit` failed when I rebased on main.

tests/automation/codegen/test_unit_test_step.py
```diff
@@ -1,17 +1,12 @@
 import unittest
-from unittest.mock import patch, MagicMock
```
ram-senth, line 2: This is unrelated to anomaly detection but `pre-commit` failed when I rebased on main.

tests/seer/anomaly_detection/test_utils.py
```diff
@@ -59,24 +60,34 @@ def convert_synthetic_ts(directory: str, as_ts_datatype: bool, include_anomaly_r
...
                 mp_dists.append(mp_dist)
                 window_sizes.append(window_size)
                 anomaly_starts.append(start)
                 anomaly_ends.append(end)
                 expected_types.append(expected_type)
 
     if include_anomaly_range:
-        return expected_types, timeseries, mp_dists, window_sizes, anomaly_starts, anomaly_ends
-    return timeseries, mp_dists, window_sizes
+        return (
```
corps, line 84: Hmmm...  Does it make sense to return a structured type and not a tuple?

If we used a dataclasses, we could give all these fields a name, preventing large amounts of confusion.  The overloaded return type could be simpler, too, as the anomaly details can just be nullable fields.

corps (approved):
No blocking concerns, but the type of return in the test utility is overloaded.

tests/seer/anomaly_detection/test_utils.py
aayush-se: I agree with this. I initially wrote this when there were fewer return values but as that has increased a class is a good starting point for this refactor.

ram-senth: Will include this in the next PR.

PR #1219, opened 2024-10-02
feat(autofix): Keyword search only downloads repo once

jennmueng:
The keyword search tool will only download the repo into a temp directory once. You will now need to use basetools with the context manager:
```
with BaseTools(self.context) as tools:
```
the tmp dir and downloaded repo will be deleted at the closing of the context manager.

### DIFF
--- a/src/seer/automation/autofix/components/coding/component.py
+++ b/src/seer/automation/autofix/components/coding/component.py
@@ -89,84 +89,87 @@ class CodingComponent(BaseComponent[CodingRequest, CodingOutput]):
     @observe(name="Plan+Code")
     @ai_track(description="Plan+Code")
     def invoke(self, request: CodingRequest) -> CodingOutput | None:
-        tools = BaseTools(self.context)
-
-        agent = ClaudeAgent(
-            tools=tools.get_tools(),
-            config=AgentConfig(system_prompt=CodingPrompts.format_system_msg(), interactive=True),
-            memory=request.initial_memory,
-        )
-
-        task_str = (
-            RootCausePlanTaskPromptXml.from_root_cause(request.root_cause_and_fix).to_prompt_str()
-            if isinstance(request.root_cause_and_fix, RootCauseAnalysisItem)
-            else request.root_cause_and_fix
-        )
-
-        state = self.context.state.get()
-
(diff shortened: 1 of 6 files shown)


src/seer/automation/autofix/components/coding/component.py
```diff
@@ -89,83 +89,86 @@ def _handle_missing_file_changes(
...
-                else None
-            ),
-            context=self.context,
-        )
+        with BaseTools(self.context) as tools:
+            agent = ClaudeAgent(
+                tools=tools.get_tools(),
+                config=AgentConfig(
+                    system_prompt=CodingPrompts.format_system_msg(), interactive=True
+                ),
```
roaga, line 97: This whole file is a diff, but is it just this `as tools` line at the top that's new?

roaga: Is there really no other way to implement this besides the `with` syntax? It's kind of annoying.

Could we, for example, store the tools in memory in the context? (I don't have strong ideas here)

jennmueng: the context manager is the cleanest and surest way the files get cleaned up at the end. Any other method will require you to call tools.cleanup() which is arguably more annoying right

jennmueng: (and also lead to potential filesystem leaks where we don't delete the repo and it becomes a problem)

roaga: If it's in memory, wouldn't it just disappear when the context stops being used?

jennmueng: How can you store files in memory

jennmueng: The keyword search tool download the repo to perform a search right

roaga: Oh I was misunderstanding how the tool works

roaga (approved):

src/seer/automation/autofix/components/coding/component.py
roaga: Codecov tho

PR #1230, opened 2024-10-04
chore(anomaly-detection): Some more refactoring

ram-senth:


aayush-se (approved):
Looks great thanks for this refactor!

aayush-se:
the `mp_scorers` file went down in test coverage in codecov so may need to add a few simple tests to address those cases (believe the the ClientErrors are not fully tested based on codecov report).

ram-senth:
> the `mp_scorers` file went down in test coverage in codecov so may need to add a few simple tests to address those cases (believe the the ClientErrors are not fully tested based on codecov report).

Updated to 100% coverage of the diff.

PR #1231, opened 2024-10-04
ref(autofix): New llm client and agent interface

jennmueng:
We were calling LLMs in multiple different ways, and the llm clients and the agents were getting somewhat unwieldy, so this refactoring work aims to simplify, unify, and optimize how we interact with LLMs.

# The LLM Clients:
The LLM clients now follow a similar interface to the [vercel ai sdk](https://sdk.vercel.ai/docs/ai-sdk-core/overview), the two main methods you will interact with is `generate_text` and `generate_structured`, where the former is for normal completion and the latter is for structured output completion. The structured output completion currently only works with the OpenAI provider for now.

### LLM Providers
The LLM providers, `OpenAiProvider` and `AnthropicProvider` contains all the specific logic for each provider, you select the model via the provider and pass it to the generate calls, for example `model=OpenAiProvider.model('gpt-4o-mini')`. This allows us to bake in defaults such as temperature into the model selection.

### The calling methods
You can call the LLM clients via the two generate calls:
```py
# generate_text
response = llm_client.generate_text(
  model=AnthropicProvider.model("claude-3-5-sonnet@20240620"),
  prompt=CodingPrompts.format_incorrect_diff_fixer(
    file_path,
    file_missing_obj.diff_chunks,
    file_missing_obj.file_content,
  ),
  temperature=0.0,
  run_name="Incorrect Diff Fixer",
)
```
```py
# generate_structured
response = llm_client.generate_structured(
  model=OpenAiProvider.model("gpt-4o-mini"),
  system_prompt="...",
  messages=[...]
)
```
You can pass messages, or a `prompt` which will be appended as the user message.

# The LLM Agents
You initialize an llm agent as such:
```py
agent = AutofixAgent(
  tools=tools.get_tools(),
  config=AgentConfig(interactive=True),
  memory=request.initial_memory,
  context=self.context,
  name="Plan+Code",
)
```
and each call has its own `RunConfig`, you can run it as many times as you want to share the tools and memory across multiple agent runs.
```py
response = agent.run(
  run_config=RunConfig(
    prompt=(
      CodingPrompts.format_fix_discovery_msg(
        event=request.event_details.format_event(),
        task_str=task_str,
        summary=request.summary,
        repo_names=[repo.full_name for repo in state.request.repos],
        instruction=request.instruction,
      )
      if not request.initial_memory
      else None
    ),
    system_prompt=CodingPrompts.format_system_msg(),
    model=AnthropicProvider.model("claude-3-5-sonnet@20240620"),
    memory_storage_key="plan_and_code",
    run_name="Plan",
  ),
)
```

### Naming
As you can see with the above, you can just pass run names to the above calls and they will be grouped + named in langfuse neatly:
CleanShot 2024-10-08 at 10 52 29

### DIFF
--- a/src/seer/automation/agent/client.py
+++ b/src/seer/automation/agent/client.py
@@ -2,7 +2,7 @@ import json
 import logging
 import re
 from dataclasses import dataclass
-from typing import Generic, Iterable, Optional, Type, TypeVar, Union, cast
+from typing import ClassVar, Generic, Iterable, Optional, Type, TypeVar, Union, cast
 
 import anthropic
 from anthropic import NOT_GIVEN
@@ -40,28 +40,38 @@ class LlmGenerateStructuredResponse(BaseModel, Generic[StructuredOutputType]):
     metadata: LlmResponseMetadata
 
 
+class LlmProviderDefaults(BaseModel):
+    temperature: float | None = None
+
+
 class LlmProviderDefinition(BaseModel):
     model_name: str
     provider_name: LlmProviderType
+    defaults: LlmProviderDefaults | None = None
+
+
+class LlmModelDefaultConfig(BaseModel):
+    match: str
+    defaults: LlmProviderDefaults
 
 
 @dataclass
 class OpenAiProvider:
     provider: LlmProviderDefinition
 
-    default_configs = [
-        {
-            "match": r"^o1-mini.*",
-            "temperature": 1.0,
-        },
-        {
-            "match": r"^o1-preview.*",
-            "temperature": 1.0,
-        },
-        {
-            "match": r".*",
-            "temperature": 0.0,
-        },
+    default_configs: ClassVar[list[LlmModelDefaultConfig]] = [
+        LlmModelDefaultConfig(
+            match=r"^o1-mini.*",
+            defaults=LlmProviderDefaults(temperature=1.0),
+        ),
+        LlmModelDefaultConfig(
+            match=r"^o1-preview.*",
+            defaults=LlmProviderDefaults(temperature=1.0),
+        ),
+        LlmModelDefaultConfig(
+            match=r".*",
+            defaults=LlmProviderDefaults(temperature=0.0),
+        ),
     ]
 
     @staticmethod
@@ -70,38 +80,50 @@ class OpenAiProvider:
 
     @classmethod
     def model(cls, model_name: str) -> "OpenAiProvider":
+        model_config = cls._get_config(model_name)
         return cls(
             provider=LlmProviderDefinition(
-                model_name=model_name, provider_name=LlmProviderType.OPENAI
+                model_name=model_name,
+                provider_name=LlmProviderType.OPENAI,
+                defaults=model_config.defaults if model_config else None,
             )
         )
 
-    def _get_config(self, model_name: str):
-        for config in self.default_configs:
-            if re.match(config["match"], model_name):  # type: ignore
+    @classmethod
+    def _get_config(cls, model_name: str):
+        for config in cls.default_configs:
+            if re.match(config.match, model_name):
                 return config
         return None
 
     def generate_text(
         self,
         *,
-        message_dicts: Iterable[ChatCompletionMessageParam],
-        tool_dicts: Iterable[ChatCompletionToolParam] | None = None,
+        prompt: str | None = None,
+        messages: list[Message] | None = None,
+        system_prompt: Optional[str] = None,
+        tools: Optional[list[FunctionTool]] = [],
         temperature: float | None = None,
         max_tokens: int | None = None,
     ):
-        openai_client = self.get_client()
+        message_dicts, tool_dicts = self._prep_message_and_tools(
+            messages=messages,
+            prompt=prompt,
+            system_prompt=system_prompt,
+            tools=tools,
+        )
 
-        config = self._get_config(self.provider.model_name)
-        if config:
-            if temperature is None:
-                temperature = config["temperature"]
+        openai_client = self.get_client()
 
         completion = openai_client.chat.completions.create(
             model=self.provider.model_name,
-            messages=message_dicts,
+            messages=cast(Iterable[ChatCompletionMessageParam], message_dicts),
             temperature=temperature,
-            tools=(tool_dicts if tool_dicts else openai.NotGiven()),
+            tools=(
+                cast(Iterable[ChatCompletionToolParam], tool_dicts)
+                if tool_dicts
+                else openai.NotGiven()
+            ),
             max_tokens=max_tokens or openai.NotGiven(),
         )
 
@@ -140,7 +162,7 @@ class OpenAiProvider:
     def generate_structured(
         self,
         *,
-        prompt: str,
+        prompt: str | None = None,
         messages: list[Message] | None = None,
         system_prompt: Optional[str] = None,
         tools: Optional[list[FunctionTool]] = [],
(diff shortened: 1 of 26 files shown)


src/seer/automation/agent/client.py
```diff
@@ -1,203 +1,292 @@
...
+class OpenAiProvider:
+    provider: LlmProviderDefinition
+
+    default_configs: ClassVar[list[LlmModelDefaultConfig]] = [
+        LlmModelDefaultConfig(
+            match=r"^o1-mini.*",
+            defaults=LlmProviderDefaults(temperature=1.0),
+        ),
+        LlmModelDefaultConfig(
+            match=r"^o1-preview.*",
```
roaga, line 46: Why default to o1?

src/seer/automation/agent/client.py
```diff
@@ -1,203 +1,292 @@
...
+            defaults=LlmProviderDefaults(temperature=0.0),
+        ),
+    ]
+
+    @staticmethod
+    def get_client() -> openai.Client:
+        return openai.Client()
+
+    @classmethod
+    def model(cls, model_name: str) -> "OpenAiProvider":
```
roaga, line 60: the lowercase i in OpenAi pains me

src/seer/automation/agent/client.py
roaga: And what's with all the regex?

src/seer/automation/agent/client.py
```diff
@@ -1,203 +1,292 @@
...
+        messages: list[Message] | None = None,
         system_prompt: Optional[str] = None,
         tools: Optional[list[FunctionTool]] = [],
-        response_format: Optional[dict] = None,
-        temperature: float = 0.0,
-    ) -> tuple[Message, Usage]:
-        raise NotImplementedError
+        temperature: float | None = None,
+        max_tokens: int | None = None,
+    ):
```
roaga, line 86: Should they default to None or an actual value, like 0 for temp and 2048 for tokens?

src/seer/automation/agent/client.py
```diff
@@ -1,203 +1,292 @@
...
 from seer.dependency_injection import inject, injected
 
 logger = logging.getLogger(__name__)
 
-T = TypeVar("T")
 
+@dataclass
+class OpenAiProvider:
+    provider: LlmProviderDefinition
+
```
roaga, line 39: Does it make sense to use inheritance from an abstract `AIProvider`?

src/seer/automation/agent/client.py
```diff
@@ -289,60 +326,135 @@ def _format_claude_response_to_message(self, completion: anthropic.types.Message
...
+            model = cast(AnthropicProvider, model)
+
+            return model.generate_text(
+                max_tokens=max_tokens,
+                messages=messages,
+                prompt=prompt,
+                system_prompt=system_prompt,
+                temperature=temperature or default_temperature,
+                tools=tools,
             )
```
roaga, line 402: You probably don't have to duplicate that generate_text call

src/seer/automation/agent/tools.py
```diff
@@ -61,17 +64,17 @@ def to_dict(
...
-        if model == "gpt":
-            return GPTFunctionSchema(
+        if provider == LlmProviderType.OPENAI:
+            return ChatCompletionToolParam(
                 type="function",
                 function={
                     "name": self.name,
                     "description": self.description,
-                    "parameters": base_schema,
+                    "parameters": base_schema,  # type: ignore
```
roaga, line 73: why ignore the type? Mypy issue?

roaga (approved):
This is really clean 🥔 Few minor questions below

But note that I can't check if this breaks anything just by reading the code, and since it's such a big refactor that touches multiple endpoints, pls make sure everything is stable end-to-end in all of them (including interactive flow on the frontend) 🙏

src/seer/automation/agent/client.py
jennmueng: Oh this is to set default configuration values for when you select a model. So we're matching the model name, if you pass in `o1-mini...` then it'll use the default temperature `1.0` instead of `0.0`, and these are ordered, so every other model will match the general regex case below.

src/seer/automation/agent/client.py
jennmueng: We're defaulting to none here but using the default config to set them, hmm I think we shouldn't defualt a max tokens because it's highly dependent on model.

src/seer/automation/agent/client.py
jennmueng: It's intentional as the inputs can be different, and we make that distinction in the LLMClient and nowhere else

src/seer/automation/agent/client.py
jennmueng: fair enough, but I left it this way as the inputs can change (we have a lot more input fields that aren't in here yet)

src/seer/automation/agent/tools.py
jennmueng: mypy issue yes

src/seer/automation/agent/client.py
roaga: Idk that feels kind of fragile to me. Why not just have an enum of supported models (and corresponding default configs) for each provider?

jennmueng: Ah, this is a pattern i took from other AI SDKs including vercel and langfuse, this is because new models come out so fast and we don't want to have to explicitly define it, for example if a new o1-mini-2024-10-05 comes out or something the regex will match it without us needing to make a code change

src/seer/automation/agent/models.py
```diff
@@ -41,20 +48,98 @@ class Message(BaseModel):
...
-                new_item["type"] = "function"
-                parsed_tool_calls.append(new_item)
-            message["tool_calls"] = parsed_tool_calls
-
-        if self.tool_call_id:
-            message["tool_call_id"] = self.tool_call_id
-
-        return message
+    def to_message(self, provider_name: LlmProviderType) -> dict[str, Any]:
+        if provider_name == LlmProviderType.OPENAI:
```
corps, line 52: Kind of a minor nit, but this sort of overloading a method on an enum like this obfuscates a few things

1.  It makes it harder to understand how any specific provider works, since you have to hunt down its individual conditionals in multiple places
2.  Makes it unclear what the interface is for implementing. new LlmProviderType as you effectively are hoping on hitting exceptions that tell you which methods you need to implement a new conditional for.

Does it make sense to invert this an implement providers as their own classes implementing an abstract interface, that receive the Message and convert that to the dictionary, rather than heavily overloading this and other methods to behave so widely for all providers?

Another risk is because you're relying on the conditional structure to separate implementations, there is a large risk of an accidental bug in which one provider's implementation bleeds into the others.

src/seer/automation/agent/tools.py
```diff
@@ -61,17 +64,17 @@ def to_dict(
             "required": self.required,
         }
 
-        if model == "gpt":
-            return GPTFunctionSchema(
+        if provider == LlmProviderType.OPENAI:
```
corps, line 67: Similar kind of issue where I wonder if having the provider be a class implementing an interface that receives the tool and returning the schema makes more sense.  From the perspective of the interface, the return type can be `dict` but each subclass can specialize that return type thanks to covariance.

src/seer/automation/agent/models.py
```diff
@@ -41,20 +48,98 @@ class Message(BaseModel):
...
+class LlmGenerateStructuredResponse(BaseModel, Generic[StructuredOutputType]):
+    parsed: StructuredOutputType
+    metadata: LlmResponseMetadata
+
+
+class LlmProviderDefaults(BaseModel):
+    temperature: float | None = None
+
+
+class LlmProviderDefinition(BaseModel):
```
corps, line 131: Like this could definitely be, an abstract base class that presents an interface.

src/seer/automation/agent/models.py
jennmueng: Let's move all this logic into the providers 👍

src/seer/automation/agent/models.py
jennmueng: moved all of this logic into the providers itself

PR #1232, opened 2024-10-04
change: Add grpc bootstrap

dmajere:
This is a follow up to https://github.com/getsentry/seer/pull/1195
Add traffic director bootstrap script to docker image and its execution.


corps:
This works for now, we'll need to revisit this though when we need to change seer into a proper grpc consumer of sentry.

corps (approved):

PR #1234, opened 2024-10-07
feature(anomaly-detection): direction detection using prophet model

ram-senth:
This PR uses Prophet model to predict expected value of time series and uses the prediction to check for direction. 

Anomaly detected with direction set to UP:
image

Anomaly not detected with direction set to DOWN:
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ gocd: ## Build GoCD pipelines
 	cd ./gocd/generated-pipelines && find . -type f \( -name '*.yaml' \) -print0 | xargs -n 1 -0 yq -p json -o yaml -i
 .PHONY: gocd
 
-upgrade-package-versions:
+upgrade-package-versions: ## Upgrade package versions in requirements.txt from constraints in requirements-constraints.txt
 	pip-compile --upgrade --strip-extras --quiet requirements-constraints.txt -o requirements.txt
 
 HEAD_SHA:=$(shell git rev-parse --short HEAD)
--- a/src/seer/anomaly_detection/detectors/location_detectors.py
+++ b/src/seer/anomaly_detection/detectors/location_detectors.py
@@ -97,8 +97,8 @@ class ProphetLocationDetector(LocationDetector):
 
     """
 
-    uncertainty_samples: bool = Field(
-        default=True, description="Whether to use uncertainty samples for Prophet"
+    uncertainty_samples: int = Field(
+        default=25, description="Whether to use uncertainty samples for Prophet"
     )
 
     @sentry_sdk.trace
@@ -135,7 +135,7 @@ class ProphetLocationDetector(LocationDetector):
 
         # Predict and compare with streamed value
         forecast = model.predict(future)
-        if self.uncertainty_samples:
+        if self.uncertainty_samples > 0:
             prophet_trend_upper = forecast.loc[len(forecast) - 1]["trend_upper"]
             prophet_trend_lower = forecast.loc[len(forecast) - 1]["trend_lower"]
             if streamed_value > prophet_trend_upper:
--- a/src/seer/anomaly_detection/detectors/location_detectors.py
+++ b/src/seer/anomaly_detection/detectors/location_detectors.py
@@ -119,9 +119,9 @@ class ProphetLocationDetector(LocationDetector):
             history_timestamps (npt.NDArray[np.float64]): Historical time series timestamps
 
         Returns:
-            PointLocation: The detected relative location of the streamed value as compared to the trend or expected value (UP, DOWN, or NONE).
-            UP: The streamed value is above the trend or expected value.
-            DOWN: The streamed value is below the trend or expected value.
+            PointLocation: The detected relative location of the streamed value as compared to the predicted value (UP, DOWN, or NONE).
+            UP: The streamed value is above the predicted value.
+            DOWN: The streamed value is below the predicted value.
             NONE: The streamed value is within the expected range of recent data points.
         """
         # Create Prophet model and fit on historical data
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -90,7 +90,7 @@ gocd: ## Build GoCD pipelines
 	cd ./gocd/generated-pipelines && find . -type f \( -name '*.yaml' \) -print0 | xargs -n 1 -0 yq -p json -o yaml -i
 .PHONY: gocd
 
-upgrade-package-versions:
+upgrade-package-versions: ## Upgrade package versions in requirements.txt from constraints in requirements-constraints.txt
```
ram-senth, line 93: Unrelated change but ensures that `make help` gives help for `upgrade-package-versions` as well.

src/seer/anomaly_detection/detectors/location_detectors.py
```diff
@@ -0,0 +1,149 @@
...
+    ) -> PointLocation:
+        """
+        Detect trend using linear regression on recent data points.
+
+        Args:
+            cur_val (np.float64): The current value
+            ts (npt.NDArray[np.float64]): Historical time series data
+
+        Returns:
+            Trend: The detected trend (INCREASING, DECREASING, or STABLE)
```
aayush-se, line 64: Does `PointLocation.NONE` represent `STABLE`? It seems like in detect that `PointLocation.NONE` can be returned for not enough data or stable (within threshold). Just making sure that this is the correct understanding of this

aayush-se: Also nit but this comment should say we are returning a PointLocation

src/seer/anomaly_detection/detectors/mp_scorers.py
```diff
@@ -261,6 +291,45 @@ def _to_flag(self, mp_dist: np.float64, threshold: float):
...
+        context: npt.NDArray[np.float64]
+            Recent time series values for context
+
+        Returns:
+        AnomalyFlags
+            The adjusted anomaly flag
+        """
+        if flag == "none" or direction == "both":
+            return flag
+        # trend_detector = LinearRegressionLocationDetector()
```
aayush-se, line 322: nit: can remove this comment

aayush-se (approved):
Looks good just a few minor nits

src/seer/anomaly_detection/detectors/location_detectors.py
ram-senth: NONE says that the observed value is not too far away from the expected value (i.e neither above or below). Will fix the comments.

src/seer/anomaly_detection/detectors/location_detectors.py
```diff
@@ -0,0 +1,154 @@
...
+    Detects relative location of the streamed value using Facebook's Prophet forecasting model.
+
+    This detector uses Prophet to fit a model on historical data and make a prediction
+    for the current timestamp. It then compares the streamed value against the predicted
+    value (or prediction interval if uncertainty samples are used) to determine if the
+    point is above, below, or within the expected range.
+
+    """
+
+    uncertainty_samples: bool = Field(
```
trillville, line 100: unless their documentation is out of date and something has changed since i used this library, this should be an integer https://facebook.github.io/prophet/docs/diagnostics.html and it basically reflects the number of monte carlo simulations to run (this wsa the very slow step in my experience)

src/seer/anomaly_detection/detectors/location_detectors.py
```diff
@@ -0,0 +1,154 @@
...
+        model.fit(ts)
+
+        # Create future dataframe with the current timestamp for prediction
+        future = pd.DataFrame({"ds": np.append(history_timestamps, [streamed_timestamp])})
+        future["ds"] = future["ds"].astype(float)
+
+        # Predict and compare with streamed value
+        forecast = model.predict(future)
+        if self.uncertainty_samples:
+            prophet_trend_upper = forecast.loc[len(forecast) - 1]["trend_upper"]
```
trillville, line 139: why trend_upper and lower vs yhat_upper and lower?

src/seer/anomaly_detection/detectors/location_detectors.py
trillville: i wonder if we are effectively passing 1 uncertainty sample?

src/seer/anomaly_detection/detectors/location_detectors.py
trillville: also, have we done sensitivity analysis WRT how this logic should be framed? e.g. what happens when MP says its an anomaly, but its not _that_ high - does this tend to agree?

src/seer/anomaly_detection/detectors/mp_scorers.py
```diff
@@ -195,6 +216,15 @@ def batch_score(
             scores.append(0.0 if np.isnan(val) or np.isinf(val) else val - threshold)
             flag = self._to_flag(val, threshold)
 
+            flag = self._adjust_flag_for_direction(
```
trillville, line 219: can we make it so this only happens when it predicts an anomaly? we can skip this step entirely when there is no anomaly right?

src/seer/anomaly_detection/detectors/location_detectors.py
ram-senth: You are right, `uncertainty_samples` is an integer. I made a mistake. Will review the results with a slightly higher number.

src/seer/anomaly_detection/detectors/location_detectors.py
ram-senth: So prophet has two knobs to control uncertainty - `uncertainty_samples` allows us to get a threshold around trend while `mcmc_samples` adds threshold around cyclicality. I played around with mcmc_samples a bit (it also uses monte carlo simulation). It slowed it down significantly but did not add too much value especially since we are always predicting just one time step. Hence I limited the experiments to trend based prediction.

src/seer/anomaly_detection/detectors/mp_scorers.py
ram-senth: Yes that check is happening inside the method `_adjust_flag_for_direction`.

src/seer/anomaly_detection/detectors/location_detectors.py
ram-senth: https://facebook.github.io/prophet/docs/uncertainty_intervals.html has more details about those two knobs.

src/seer/anomaly_detection/detectors/mp_scorers.py
ram-senth: Done

src/seer/anomaly_detection/detectors/location_detectors.py
ram-senth: Updated comments

src/seer/anomaly_detection/detectors/location_detectors.py
ram-senth: I changed this field to int and defaulted it to 1. Bumping up the number of samples doesn't help much, most likely because we are predicting only one time step. I have some plots showing the results with uncertainty_samples set as 10 and 25. I think having it as 1 does a pretty decent job for our needs.

src/seer/anomaly_detection/detectors/mp_scorers.py
ram-senth: Here is a plot that includes the prediction time for each time step (last plot):
image

PR #1246, opened 2024-10-08
feat(autofix) Add fix instruction to root cause selection

roaga:
The user can optionally provide an instruction with the root cause selection, which is used in the prompt for the plan+code step.

### DIFF
--- a/src/seer/automation/autofix/models.py
+++ b/src/seer/automation/autofix/models.py
@@ -83,6 +83,7 @@ class CustomRootCauseSelection(BaseModel):
 
 class CodeContextRootCauseSelection(BaseModel):
     cause_id: int
+    instruction: str | None = None
 
 
 RootCauseSelection = Union[CustomRootCauseSelection, CodeContextRootCauseSelection]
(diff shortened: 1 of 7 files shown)


src/seer/automation/autofix/models.py
```diff
@@ -416,10 +420,10 @@ def get_selected_root_cause_and_fix(self) -> RootCauseAnalysisItem | str | None:
                         for cause in root_cause_step.causes
                         if cause.id == root_cause_step.selection.cause_id
                     )
-                    return cause
+                    return cause, root_cause_step.selection.instruction
```
jennmueng, line 423: Would it be cleaner if we added the instruction as a field to the root cause model?

jennmueng (approved):
suggested improvement at your discretion. lgtm otherwise.

src/seer/automation/autofix/models.py
roaga: I don't think it makes sense for the fix instruction to be added to the cause model, both semantically and because the cause is generated in the previous step, when we don't have the instruction yet.

PR #1264, opened 2024-10-10
feat(copilot): Add new comments for copilot, add extra info

roaga:
1. Adds a comment to the suspect PR with a link to the new fix PR when generated.
2. Adds a link to the suspect PR in the fix PR description if one is provided.

### DIFF
--- a/src/seer/automation/codebase/repo_client.py
+++ b/src/seer/automation/codebase/repo_client.py
@@ -452,6 +452,17 @@ class RepoClient:
         data.raise_for_status()  # Raise an exception for HTTP errors
         return data.text
 
+    def _get_auth_headers(self):
+        requester = self.repo._requester
+        if requester.auth is None:
+            raise Exception("No auth token found for GitHub API")
+        headers = {
+            "Accept": "application/vnd.github+json",
+            "Authorization": f"Bearer {requester.auth.token}",
+            "X-GitHub-Api-Version": "2022-11-28",
+        }
+        return headers
+
     def comment_root_cause_on_pr_for_copilot(
         self, pr_url: str, run_id: int, issue_id: int, comment: str
     ):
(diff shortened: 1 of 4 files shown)


src/seer/automation/codebase/repo_client.py
```diff
@@ -479,6 +479,32 @@ def comment_root_cause_on_pr_for_copilot(
...
+
+        requester = self.repo._requester
+        if requester.auth is None:
+            raise Exception("No auth token found for GitHub API")
+
+        headers = {
+            "Accept": "application/vnd.github+json",
+            "Authorization": f"Bearer {requester.auth.token}",
+            "X-GitHub-Api-Version": "2022-11-28",
+        }
```
jennmueng, line 503: We should abstract this away at this point we do this a couple times in the repoclient, not blocker, just note

src/seer/automation/codebase/repo_client.py
```diff
@@ -479,6 +479,32 @@ def comment_root_cause_on_pr_for_copilot(
         response = requests.post(url, headers=headers, json=params)
         response.raise_for_status()
 
+    def comment_pr_generated_for_copilot(
```
jennmueng, line 482: nit: should make this a more generic name not just for copilot

jennmueng (approved):
lgtm, one note and 1 lil nit

src/seer/automation/codebase/repo_client.py
roaga: Done

src/seer/automation/codebase/repo_client.py
roaga: want to keep it explicit since it's throwaway work. This way future readers will know why this weird logic exists and when it's triggered

PR #1266, opened 2024-10-10
feat(unit-test): Use codecov unit test gh app and fallback to autofix app if not present

jennmueng:
For unit test gen, use the codecov github app for repo client if available, if not fallback to autofix app

Tested with both unit test gen and autofix to make sure this doesn't break anything.

### DIFF
--- a/src/seer/automation/codebase/repo_client.py
+++ b/src/seer/automation/codebase/repo_client.py
@@ -3,7 +3,7 @@ import os
 import shutil
 import tarfile
 import tempfile
-from typing import Literal
+from enum import Enum
 
 import requests
 import sentry_sdk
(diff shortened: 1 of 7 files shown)


src/seer/automation/codebase/repo_client.py
```diff
@@ -77,6 +78,25 @@ def get_read_app_credentials(config: AppConfig = injected) -> tuple[int | str |
...
+        return get_write_app_credentials()
+
+    return app_id, private_key
+
+
+class RepoClientType(str, Enum):
+    READ = "read"
+    WRITE = "write"
+    CODECOV_UNIT_TEST = "codecov_unit_test"
+
```
roaga, line 98: kinda ugly to have a special codecov type next to read and write for a util shared in `automation`

roaga (approved):
It's ugly but it's urgent and seems to not break anything so I'll approve

src/seer/automation/codebase/repo_client.py
jennmueng: fair enough, but i think the other solutions are although less ugly they are less clear in their uses

PR #1276, opened 2024-10-12
feature(anomaly detection): Flag Smoothing

aayush-se:
- Implement flag based anomaly smoothing to stream and batch detection
- dynamic smoothing window size based on config


src/seer/anomaly_detection/detectors/smoothers.py
```diff
@@ -0,0 +1,120 @@
...
+from typing import List, Tuple
+
+import numpy as np
+import numpy.typing as npt
+from pydantic import BaseModel, Field
+
+from seer.anomaly_detection.models.external import AnomalyDetectionConfig
+
+
+class Smoother(BaseModel, abc.ABC):
```
ram-senth, line 11: I would rename the base class as `FlagSmoother` that is the generic name and the other class as `MajorityVoteFlagSmoother`, which is more specific to the actual algorithm being used.

src/seer/anomaly_detection/detectors/anomaly_detectors.py
```diff
@@ -177,6 +189,16 @@ def detect(
                 )
                 if flags_and_scores is None:
                     raise ServerError("Failed to score the matrix profile distance")
+
+                # Apply smoothing to the flags
+                flag_smoother = FlagSmoother()
```
ram-senth, line 194: Can we have this dependency injected (just like the scorer, window size selector etc)?

src/seer/anomaly_detection/detectors/smoothers.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    Abstract base class for smoothing data (specifically flags)
+    """
+
+    @abc.abstractmethod
+    def smooth(
+        self,
+        orig_ts: npt.NDArray,
```
ram-senth, line 19: Why do we need `orig_ts` parameter? I do not see it being used anywhere (unless I missed something)?

ram-senth (approved):

PR #1308, opened 2024-10-17
feat(autofix): Support user editing code changes

roaga:
Allows the user to edit or reject hunks of the code changes. We update the changes stored in the step.

This also required changing the PR creation logic to support creating a PR from these updated diffs (`FilePatch`) as well as the old method (`FileChange`).

One limitation is that there is no unique commit message for each file patch.


src/seer/automation/codebase/repo_client.py
```diff
@@ -322,60 +322,91 @@ def _create_branch(self, branch_name):
...
         return ref
 
-    def _commit_file_change(self, change: FileChange, branch_ref: str):
-        contents = (
-            self.repo.get_contents(change.path, ref=branch_ref)
-            if change.change_type != "create"
-            else None
-        )
+    def _commit_file_change(
+        self, branch_ref: str, patch: FilePatch | None = None, change: FileChange | None = None
```
jennmueng, line 326:
```suggestion
        self, *, branch_ref: str, patch: FilePatch | None = None, change: FileChange | None = None
```
Given that we use this with kwargs only anyways and either field is required

src/seer/automation/codebase/repo_client.py
```diff
@@ -322,60 +322,91 @@ def _create_branch(self, branch_name):
...
+                path,
+                commit_message or "File change",
                 new_contents or "",
                 contents.sha,  # FYI: It wants the sha of the content blob here, not a commit sha.
                 branch=branch_ref,
             )
 
     def create_branch_from_changes(
-        self, pr_title: str, file_changes: list[FileChange], branch_name: str | None = None
+        self,
```
jennmueng, line 384:
```suggestion
        self,
        *
```
same here

jennmueng (approved):

PR #1309, opened 2024-10-18
Better staging build support, bump versions

corps:
Includes settings and ability to submit staging builds in cloud build format, rather than uploading entire image of local build.
Also bumps some library versions.


jennmueng (approved):
🫶

PR #1338, opened 2024-10-24
Adding dependency injection docs

corps:


DI.md
```diff
@@ -0,0 +1,378 @@
...
+
+Nice!  This approach is actually great.  Explicit parameters are obvious, easy to use, and naturally deal with both
+the Robot Leg problem and the Lifecycle problem directly.  **If you can get away with solving any problem with a simple
+function, do it**.
+
+However, there are some caveats here.  What if `my_method` isn't always called directly by you or your test?  For
+instance, suppose that `my_method` is celery task that is invoked by your celery runner.  Suppose `my_method` is a
+callback that doesn't provide any easy way to control its parameterization?   Or suppose that the mutated
+state isn't explicit, but merely implicit due to some other process side effect?  You may have trouble finding a way
+to thread your parameter in all the right places to maintain internal consistency of the instance:
```
roaga, line 67: extra newline?

DI.md
```diff
@@ -0,0 +1,378 @@
...
+still bind a value to the invocation that uses the default.
+
+```python
+def unit_test_my_method():
+    # You can provide a settings value directly for a simple unit test, to validate a special behavior
+    assert not my_method(10, settings=Settings(DISABLE=True))
+    # Or you can simply allow the inject decorator to provide the value directly.
+    with module.constant(Settings, Settings(DISABLE=True)):
+        assert my_method(10)
+        with celery_runner():
```
roaga, line 108: is it the `with` keyword that allows the DI to work here?

DI.md
```diff
@@ -0,0 +1,378 @@
...
+
+
+### Dependency Injection as the Alternative
+
+With dependency injection, you build upon the option of passing in parameters directly with the concept of "injection",
+the idea that a context can provide a default for a given value type.
+
+```python
+@inject
+def my_method(a: int, settings: Settings = injected):
```
roaga, line 87: Questions that may pop up:

How do I inject multiple different parameters? Do I need different keywords? If we use the same keyword, does the library just infer based on type? What if they are two different parameters of the same type?

DI.md
```diff
@@ -0,0 +1,378 @@
...
+    assert not my_method(10, settings=Settings(DISABLE=True))
+    # Or you can simply allow the inject decorator to provide the value directly.
+    with module.constant(Settings, Settings(DISABLE=True)):
+        assert my_method(10)
+        with celery_runner():
+            # Some implicitly run celery job will still receive the same configured Settings object above
+            assert resolve(Settings).CACHE[key] == value
+```
+
+There is some additional complexity created here, but there is some gained flexibility in the ability to control
```
roaga, line 113: Additional complexity? Could you touch more on any drawbacks of using DI?

DI.md
```diff
@@ -0,0 +1,378 @@
...
+
+Also note that `module.enable()` **only applies to the current thread!**.  This means that if you run a multi-threaded
+application, you may need to "enable" modules again when you enter new threads.  This is meant to ensure no unintended
+cross thread state sharing.  If you must share state between threads, consider using Queues or other concurrency
+primitives.
+
+## Differentiating values on common types
+
+In the easy case, dependency injection provides **one value** for any given type annotation, usually because that type
+is essentially a Singleton.
```
roaga, line 183: unneeded newline?

DI.md
roaga: I think you answer these well below, but it might be nice to put a "Quick Start" with a few examples of how to set up and use DI at the top before diving in-deep

DI.md
```diff
@@ -0,0 +1,378 @@
...
+2. Not "enabled" that module in your current context.
+
+What does "enabling" a module mean?  Well, if you check seer source code files, you'll often see a line that looks like
+`module.enable()`.  This essentially places all of that modules providers into scope for `injected` values of `inject`
+methods.
+
+```python
+module.enable()
+my_method() # now @inject will find `Settings` provider from above.
+```
```
roaga, line 163: does the `with` keyword and `.enable()` do the same thing in terms of setting a "current context"?

roaga (commented):
Overall, really helpful docs! Two comments:
1. It would help to have quick and simple usage guide at the top so that diving into the "why" and "how" is optional
2. Some sections could use a _little_ extra explanation and simpler language. Left comments below

Thanks for putting this together 🥔

DI.md
aayush-se: ^ Strongly agree with a quick start for those who are familiar with the concepts

DI.md
```diff
@@ -0,0 +1,378 @@
...
+
+1.  All previously instantiated values from providers will be ignored, future injection will create new instantiations
+2.  All `provider` definitions are available to bind to future types.  Any conflicts between definitions for the same
+type will be resolved with the "last" enabled module's providers.
+
+Note that it is perfectly ok to enable a module before definitions are added, say at the top of a file,
+but that enable and the provider definitions must occur **before invoking a method that depends upon them**.  Otherwise,
+the ordering is loose and lazy.
+
+Also note that `module.enable()` **only applies to the current thread!**.  This means that if you run a multi-threaded
```
aayush-se, line 175: Extra period after the exclamation

DI.md
```diff
@@ -0,0 +1,378 @@
+# Dependency Injection in Seer
+
+[Dependency Injection](https://en.wikipedia.org/wiki/Dependency_injection) is a technique in which shared objects
+and configuration used in a program are "injected" into functions on a need to know basis, as opposed to, for instance,
```
ram-senth, line 4: How about adding the high level goal of DI, something like `Main goal of DI is to separate the concern of constructing objects and managing dependencies from using those objects. This, along with interface based programming, helps with building loosely coupled programs that are easier to modify and maintain."

jennmueng:
Splitting up and merging separately

PR #1349, opened 2024-10-28
feat(autofix): Add simple issue short circut, remove repro steps

roaga:
This PR:
1. Adds a validator to the Issue Summary model so that eval datasets with old summary models can still be run.
2. Comments out the reproduction/unit test step in the root cause step.
3. Adds a check at the beginning of the root cause step that asks an LLM if tools are needed to identify the root cause (this call takes under 1 second). If it says no, the tools and certain parts of the prompt are removed.

These changes:
- reduce total run cost from ~60 cents to ~30 cents.
- improve eval scores by 3-5 points.
- on simple issues, get to the root cause much faster.

### DIFF
--- a/src/seer/automation/autofix/components/root_cause/prompts.py
+++ b/src/seer/automation/autofix/components/root_cause/prompts.py
@@ -7,19 +7,32 @@ from seer.automation.summarize.issue import IssueSummary
 
 class RootCauseAnalysisPrompts:
     @staticmethod
-    def format_system_msg():
+    def format_system_msg(has_tools: bool = True):
         return textwrap.dedent(
-            """\
+            f"""\
             You are an exceptional principal engineer that is amazing at finding the root cause of any issue.
 
-            You have tools to search a codebase to find the root cause of an issue. Please use the tools as many times as you want to find the root cause of the issue.
+            {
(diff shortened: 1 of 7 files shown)


src/seer/automation/autofix/components/root_cause/prompts.py
```diff
@@ -7,19 +7,32 @@
...
         return textwrap.dedent(
-            """\
+            f"""\
             You are an exceptional principal engineer that is amazing at finding the root cause of any issue.
 
-            You have tools to search a codebase to find the root cause of an issue. Please use the tools as many times as you want to find the root cause of the issue.
+            {
+                "You have tools to search a codebase to find the root cause of an issue. Please use the tools as many times as you want to find the root cause of the issue."
+                if has_tools
+                else "You do not have tools to search a codebase to find the root cause of an issue."
```
jennmueng, line 18: Is this needed when there are no tools, any reason to provide this rather than just leaving it blank in that case?

jennmueng (approved):
This is great, especially with the eval improvement, we can definitely short circuit plan+code on simple code changes too

src/seer/automation/autofix/components/root_cause/prompts.py
roaga: good catch, that's not needed

PR #1350, opened 2024-10-28
ref(similarity): Change negative distance to 0

jangjodi:
Change negative stacktrace differences to 0
Attempting to rule out if seer is returning negative distances


JoshFerge (commented):
why would distance ever be negative? any other context on this change?

PR #1351, opened 2024-10-28
fix: Add message column

jangjodi:


armenzg (approved):
🎉 Please add the reason why we need it in the description of the pull request.

PR #1354, opened 2024-10-29
chore(anomaly detection): fix flaky data cleanup test

aayush-se:
- Fix flaky test by sorting the `old_timeseries_points` and `new_timeseries_points` to ensure consistency of order during comparison


aayush-se:
Reason that setting the `order by` on the child relationship may not work is when data is changed (added, removed, etc) from the DB, the new/changed data does not automatically sort itself. One fix is to use an [ordering list](https://docs.sqlalchemy.org/en/20/orm/extensions/orderinglist.html), but we may not need that extra sorting overhead on our db when it is only required for a test like this.

PR #1366, opened 2024-11-01
chore(anomaly_detection): apply direction logic for alerts with both as direction

ram-senth:
This change ensures that the direction logic is applied for anomalous points even when the alert is configured for both directions. This change helps reduce noise from anomalies that do not breach any threshold.

Note: Apart the core logic of enabling direction logic, this PR also includes multiple fixes for the time stamps not being handled correctly that surfaced due to this change.

### DIFF
--- a/src/seer/anomaly_detection/detectors/anomaly_detectors.py
+++ b/src/seer/anomaly_detection/detectors/anomaly_detectors.py
@@ -84,15 +84,12 @@ class MPBatchAnomalyDetector(AnomalyDetector):
             A tuple with the matrix profile, matrix profile distances, anomaly scores, anomaly flags and the window size used
 
         """
+        if len(timeseries.values) == 0:
+            raise ServerError("No values to detect anomalies for")
+        if len(timeseries.timestamps) != len(timeseries.values):
+            raise ServerError("Timestamps and values are not of the same length")
+
         ts_values = timeseries.values
-        if len(ts_values) == 0:
-            return MPTimeSeriesAnomalies(
-                flags=np.array([]),
-                scores=np.array([]),
-                matrix_profile=np.array([]),
-                window_size=0,
-                thresholds=np.array([]),
-            )
         window_size = ws_selector.optimal_window_size(ts_values)
         if window_size <= 0:
             # TODO: Add sentry logging of this error
@@ -174,6 +171,13 @@ class MPStreamAnomalyDetector(AnomalyDetector):
         Returns:
             A MPTimeSeriesAnomalies object with the matrix profile, matrix profile distances, anomaly scores, anomaly flags and the window size used
         """
+        if len(timeseries.values) == 0:
+            raise ServerError("No values to detect anomalies for")
+        if len(self.history_values) != len(self.history_timestamps):
+            raise ServerError("History values and timestamps are not of the same length")
+
+        if len(self.history_values) - self.window_size + 1 != len(self.history_mp):
+            raise ServerError("Matrix profile is not of the right length.")
         stream = None
         with sentry_sdk.start_span(description="Initializing MP stream"):
             # Initialize stumpi
@@ -223,6 +227,7 @@ class MPStreamAnomalyDetector(AnomalyDetector):
                 thresholds.extend(flags_and_scores.thresholds)
 
                 # Add new data point as well as its matrix profile to baseline
+                self.history_timestamps = np.append(self.history_timestamps, cur_timestamp)
                 self.history_values = stream.T_
                 self.history_mp = np.vstack([self.history_mp, cur_mp])
 
--- a/src/seer/anomaly_detection/detectors/location_detectors.py
+++ b/src/seer/anomaly_detection/detectors/location_detectors.py
@@ -1,6 +1,7 @@
 import abc
 import logging
 from enum import Enum
+from typing import Optional
 
 import numpy as np
 import numpy.typing as npt
@@ -30,7 +31,7 @@ class LocationDetector(BaseModel, abc.ABC):
         streamed_timestamp: np.float64,
         history_values: npt.NDArray[np.float64],
         history_timestamps: npt.NDArray[np.float64],
-    ) -> PointLocation:
+    ) -> Optional[PointLocation]:
         return NotImplemented
 
 
@@ -53,7 +54,7 @@ class LinearRegressionLocationDetector(LocationDetector):
         streamed_timestamp: np.float64,
         history_values: npt.NDArray[np.float64],
         history_timestamps: npt.NDArray[np.float64],
-    ) -> PointLocation:
+    ) -> Optional[PointLocation]:
         """
         Detect relative location of the streamed value in the context of recent data points using linear regression.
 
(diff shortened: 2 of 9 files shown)


src/seer/anomaly_detection/detectors/anomaly_detectors.py
```diff
@@ -223,6 +223,7 @@ def detect(
                 thresholds.extend(flags_and_scores.thresholds)
 
                 # Add new data point as well as its matrix profile to baseline
+                self.history_timestamps = np.append(self.history_timestamps, cur_timestamp)
```
ram-senth, line 226: Bug fix: The timestamps were not appended before, resulting in length mismatch between time stamps and the values.

src/seer/anomaly_detection/tasks.py
```diff
@@ -67,7 +67,7 @@ def delete_old_timeseries_points(alert: DbDynamicAlert, date_threshold: float):
 def update_matrix_profiles(alert: DbDynamicAlert, anomaly_detection_config: AnomalyDetectionConfig):
 
     timeseries = TimeSeries(
-        timestamps=np.array([timestep.timestamp for timestep in alert.timeseries]),
+        timestamps=np.array([timestep.timestamp.timestamp() for timestep in alert.timeseries]),
```
ram-senth, line 70: Bug fix: Using datetime object instead of float as timestamp

src/seer/anomaly_detection/tasks.py
```diff
@@ -79,7 +79,7 @@ def update_matrix_profiles(alert: DbDynamicAlert, anomaly_detection_config: Anom
     )
     updateed_timeseries_points = 0
     for timestep in alert.timeseries:
-        timestep.anomaly_algo_data = algo_data_map[timestep.timestamp]
+        timestep.anomaly_algo_data = algo_data_map[timestep.timestamp.timestamp()]
```
ram-senth, line 82: Bug fix: Using datetime object instead of float as timestamp

tests/seer/anomaly_detection/test_anomaly_detection.py
```diff
@@ -95,7 +95,7 @@ def test_detect_anomalies_online(self, mock_reset_cleanup, mock_save_timepoint,
             "tests/seer/anomaly_detection/test_data/synthetic_series", as_ts_datatype=False
         )
         ts_values = loaded_synthetic_data.timeseries[0]
-        ts_timestamps = np.arange(1, len(ts_values), 1) + datetime.now().timestamp()
+        ts_timestamps = np.arange(1, len(ts_values) + 1) + datetime.now().timestamp()
```
ram-senth, line 98: Bug fix: Off by one error. ts_timestamps had one value less than what the ts_values had.

tests/seer/anomaly_detection/detectors/test_anomaly_detectors.py
```diff
@@ -116,15 +116,29 @@ def test_invalid_window_size(self):
...
     def setUp(self):
         self.detector = MPStreamAnomalyDetector(
-            history_timestamps=np.array([1, 2, 3]),
-            history_values=np.array([1.0, 2.0, 3.0]),
-            history_mp=np.array([0.1, 0.2, 0.3, 0.4]),
-            window_size=2,
+            history_timestamps=np.array([1, 2, 3, 4]),
+            history_values=np.array([1.0, 2.0, 3.0, 4.0]),
+            history_mp=np.array([[0.3, 0.3, 0.3, 0.3], [0.4, 0.5, 0.6, 0.7]]),
+            window_size=3,
```
ram-senth, line 141: Now that we have validations, the matrix profile dimensions should match what is expected for the given timeseries.

src/seer/anomaly_detection/tasks.py
```diff
@@ -42,7 +42,13 @@ def cleanup_timeseries(alert_id: int, date_threshold: float):
                 expected_seasonality=alert.config["expected_seasonality"],
             )
             deleted_timeseries_points = delete_old_timeseries_points(alert, date_threshold)
-            updated_timeseries_points = update_matrix_profiles(alert, config)
+            if len(alert.timeseries) > 0:
```
ram-senth, line 45: Bug fix: Avoid calling anomaly detection if the time series is empty

src/seer/anomaly_detection/detectors/mp_scorers.py
```diff
@@ -332,12 +333,16 @@ def _adjust_flag_for_direction(
         AnomalyFlags
             The adjusted anomaly flag
         """
-        if flag == "none" or direction == "both":
+        if flag == "none":
```
ram-senth, line 336: This is the core of enabling direction detection for all detected anomalies.

src/seer/anomaly_detection/detectors/location_detectors.py
```diff
@@ -70,7 +71,7 @@ def detect(
         recent_data = np.concatenate([history_values[-self.window_size :], [streamed_value]])
 
         if len(recent_data) < self.window_size + 1:
-            return PointLocation.NONE  # Not enough data to determine trend
+            return None  # Not enough data to determine trend
```
aayush-se, line 74: Could this be another enum value which represents the "not enough data" option rather than a `none`?

src/seer/anomaly_detection/tasks.py
```diff
@@ -42,7 +42,13 @@ def cleanup_timeseries(alert_id: int, date_threshold: float):
...
+                updated_timeseries_points = update_matrix_profiles(alert, config)
+            else:
+                # Reset the window size to 0 if there are no timeseries points left
+                alert.anomaly_algo_data = {"window_size": 0}
+                logger.warn(f"Alert with id {alert_id} has enpty timeseries data after pruning")
```
aayush-se, line 50: nit: typo "empty"

tests/seer/anomaly_detection/detectors/test_anomaly_detectors.py
```diff
@@ -199,47 +224,41 @@ def test_stream_detect_spiked_history_spiked_stream_long_ts(self):
...
+        assert stream_anomalies.flags[2] == "anomaly_higher_confidence"
+        # Fourth and fifth may or may not be flagged depending on prophet's prediction which has some randomness
+        assert stream_anomalies.flags[5] == "none"
+        assert stream_anomalies.flags[6] == "none"
+        assert stream_anomalies.flags[7] == "none"
+        assert stream_anomalies.flags[8] == "none"
+        assert stream_anomalies.flags[9] == "none"
+        assert stream_anomalies.flags[10] == "none"
+        assert stream_anomalies.flags[11] == "none"
+        assert stream_anomalies.flags[12] == "none"
```
aayush-se, line 241: I think this should be a loop with a filter for the fourth and fifth points to ensure consistency rather than individual asserts for each point.

tests/seer/anomaly_detection/detectors/test_anomaly_detectors.py
```diff
@@ -199,47 +224,41 @@ def test_stream_detect_spiked_history_spiked_stream_long_ts(self):
...
```
aayush-se, line 261: Same as previous comment

aayush-se (approved):
Looks good just a few suggestions on the PointLocation Enum and the tests.

src/seer/anomaly_detection/detectors/location_detectors.py
ram-senth: No, typically you do not want to use enums to represent an invalid value, unless of course, it is an enum of error codes.

PR #1370, opened 2024-11-02
feat(autofix): Prompt agent to ask questions if it's taking a while

roaga:
If the Autofix agent is about to hit the max iterations, and every 6 iterations, we prompt it saying that it's taking a while and should ask the user for help. #1369

### DIFF
--- a/src/seer/automation/autofix/autofix_agent.py
+++ b/src/seer/automation/autofix/autofix_agent.py
@@ -45,9 +45,9 @@ class AutofixAgent(LlmAgent):
         if self.context.state.get().steps[-1].status == AutofixStatus.WAITING_FOR_USER_RESPONSE:
             return False
 
-        if not super().should_continue(run_config):
-            return False
+        return super().should_continue(run_config)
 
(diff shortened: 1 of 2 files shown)


src/seer/automation/autofix/autofix_agent.py
```diff
@@ -45,7 +45,22 @@ def should_continue(self, run_config: RunConfig) -> bool:
...
+            self.config.interactive
+            and self.iterations > 0
+            and (
+                self.iterations == run_config.max_iterations - 3
+                or (self.iterations % 6 == 0 and self.iterations < run_config.max_iterations - 3)
+            )
+        ):
+            self.add_user_message(
+                "You're taking a while. If you need help, ask me a concrete question using the tool provided."
+            )
```
jennmueng, line 61: I think this should be extracted out of `should_continue` and be its own method that gets called. Should keep `should_continue` to no side effects for cleanliness. Maybe at the start of the agent iteration loop.

jennmueng (approved):

PR #1374, opened 2024-11-04
fix(anomaly detection): Fix Stream Smoothing

aayush-se:
- Fix stream smoothing by keeping track of `original_flag` in the `anomaly_algo_data` for each flag
- Use a voting threshold smoothing for the last set of points and their original flags


aayush-se:
Opening PR to check coverage. Some more minor changes are still required and tests will be updated.

ram-senth (approved):

PR #1376, opened 2024-11-04
build: Bump sentry-sdk to 2.18.x

jennmueng:


roaga (approved):
does this enable continuous profiling automatically?

jennmueng:
@roaga good callout, need to enable it

PR #1381, opened 2024-11-05
feat(evals): New eval with root cause scoring in it

jennmueng:
During a full eval, we can now score the root cause along with it to get more signal

### DIFF
--- a/src/seer/automation/autofix/tasks.py
+++ b/src/seer/automation/autofix/tasks.py
@@ -699,8 +699,6 @@ def run_autofix_evaluation_on_item(
 
                 # Will output 4 scores for a run item:
                 # - `"highest_score"`: The score for the highest scored cause out of all the returned root causes.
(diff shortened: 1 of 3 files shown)


src/seer/automation/autofix/tasks.py
```diff
@@ -686,3 +687,47 @@ def run_autofix_evaluation_on_item(
...
+                    causes,
+                    n_panel=scoring_n_panel,
+                    model=scoring_model,
+                    langfuse_session_id=trace_id,
+                )
+
+                # Will output 4 scores for a run item:
+                # - `"highest_score"`: The score for the highest scored cause out of all the returned root causes.
+                # - `"positioning_score"`: Positioning of the highest scored cause, if the highest scored cause is first, this score is `1.0`. If it is last, it will be `0.0`. The score is calculated proportional to the number of causes provided.
+                # - `"mean_score"`: The mean score of all the root causes.
```
roaga, line 703: is positioning_score  and mean_score outdated now?

roaga (approved):

src/seer/automation/autofix/tasks.py
jennmueng: yes

PR #1384, opened 2024-11-05
chore(autofix): Switch to claude for root cause step

roaga:
Model switch from GPT 4o to Claude 3.5 Haiku results in better latency, better coding, and better root cause results.

(see row 2 below):

Screenshot 2024-11-05 at 6 46 30 PM

### DIFF
--- a/src/seer/automation/autofix/components/root_cause/component.py
+++ b/src/seer/automation/autofix/components/root_cause/component.py
@@ -61,7 +61,7 @@ class RootCauseAnalysisComponent(BaseComponent[RootCauseAnalysisRequest, RootCau
             try:
                 response = agent.run(
                     run_config=RunConfig(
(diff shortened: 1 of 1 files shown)


trillville:
how much higher is the error rate? hard to tell from these metrics

roaga:
> how much higher is the error rate? hard to tell from these metrics

Good runs: 98
Errored runs: 5
Error rate: 0.04
Errored in root cause: 0
Errored in plan: 3
Error rate in root cause: 0.00
Error rate in plan: 0.02
Error rate in something after plan: 0.02
Runs with unapplied changes: 24
Missing change rate: 0.19

jennmueng (approved):
Bug in our eval, but the notebook error rate is correct

roaga:
actually going for haiku now instead of sonnet, evals are better

PR #1396, opened 2024-11-06
chore(anomaly-detection): disable direction detection for points that are not close to thresholds

ram-senth:
Disabling this as a stopgap approach. We need to dig into the interplay between the flip-flop logic and this before restoring.


aayush-se (approved):
Looks good we can probably make a ticket for the future fixes that we would need to put in too

PR #1398, opened 2024-11-07
feat(autofix): Short circuit plan+code when fix and feedback are simple

jennmueng:
We short circuit to a zero-shot claude 3.5 sonnet v2 call when the fix is obvious or the feedback is obvious. Increases our overall eval score by 0.01 points


src/seer/automation/agent/client.py
```diff
@@ -503,7 +503,7 @@ def generate_text(
     def generate_structured(
         self,
         *,
-        prompt: str,
+        prompt: str | None = None,
```
jennmueng, line 506: hotfix

src/seer/automation/autofix/components/coding/component.py
```diff
@@ -90,18 +88,149 @@ def _handle_missing_file_changes(
...
+
+            with self.context.state.update() as cur:
+                cur.steps[-1].initial_memory_length = len(expanded_files_messages) + 1
+
+        return memory
+
+    @observe(name="Is Obvious")
+    @ai_track(description="Is Obvious")
+    @inject
+    def _is_obvious(
```
jennmueng, line 178: We throw all obvious checks as inline helper methods in the coding component as they're pretty simple anyways

roaga (approved):
LGTM, but should we consider the easy/medium/hard scores in addition to the overall average?

PR #1413, opened 2024-11-08
fix(autofix): Fix plan+code overpicking the simple path

jennmueng:
It now asks whether the current context contains enough files for a fix, and decides on that.

### DIFF
--- a/src/seer/automation/autofix/components/coding/prompts.py
+++ b/src/seer/automation/autofix/components/coding/prompts.py
@@ -239,7 +239,7 @@ class CodingPrompts:
 
                 {fix_instruction}
 
-                Is the code change simple and exists in only a single file?"""
+                Is the code change simple and does it exist ONLY in files you can already see in your context here or do you need to look at other files?"""
             )
             .format(
                 event_details=event_details.format_event(),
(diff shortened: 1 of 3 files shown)


src/seer/automation/autofix/components/coding/prompts.py
```diff
@@ -239,7 +239,7 @@ def format_is_obvious_msg(
 
                 {fix_instruction}
 
-                Is the code change simple and exists in only a single file?"""
+                Is the code change simple and does it exist ONLY in files you can already see in your context here or do you need to look at other files?"""
```
roaga, line 242: Why ask for two fields? Why not just ask the 2nd question, i.e. do you need to search for other files?

If a fix is entirely contained in the files the agent already has, we're fine to short circuit, even if it's a big change, no?

I imagine asking just this question will allow us to short circuit more often

roaga (approved):
LGTM code-wise, question for your consideration though below

src/seer/automation/autofix/components/coding/prompts.py
jennmueng: fair

PR #1416, opened 2024-11-11
feat(autofix): Eval scores as a % fixed

jennmueng:


src/seer/automation/autofix/evaluations.py
```diff
@@ -366,53 +352,41 @@ def score_root_cause_single_it(
...
-        / n_panel,
-        2,
-    )
+) -> tuple[float, bool]:
+    results = [score_fix_single_it(dataset_item, predicted_diff_str, model) for _ in range(n_panel)]
+
+    mean_score = round(sum([result[0] for result in results]) / n_panel, 2)
+
+    # If at least half of the panel says the fix is correct, then the fix is correct.
+    verdict = sum(1 for result in results if result[1]) >= len(results) / 2
```
roaga, line 373: hmmmm....

Would it be better to average the 1s and 0s of the panel?

src/seer/automation/autofix/evaluations.py
```diff
@@ -366,53 +352,41 @@ def score_root_cause_single_it(
...
-        "highest_score": highest_score,
-        "position_score": position_score,
-        "mean_score": mean_score,
-    }
+) -> tuple[float, bool]:
+    results = [score_root_cause_single_it(dataset_item, causes, model) for _ in range(n_panel)]
+
+    mean_score = round(sum([result[0] for result in results]) / len(results), 2)
+
+    # If at least half of the panel says the fix is correct, then the fix is correct.
```
roaga, line 386: same question here

src/seer/automation/autofix/evaluations.py
```diff
@@ -282,7 +279,8 @@ def score_fix_single_it(dataset_item: DatasetItemClient, predicted_diff: str, mo
             - Consider that there are multiple ways to fix an issue
 
             Think step-by-step inside a <thoughts> tag before giving a score.
-            Return the score inside a <score> tag."""
+            Return the score inside a <score> tag.
+            Then, return your verdict of whether the predicted solution fixes the issue with a boolean value inside a <verdict> tag, such as <verdict>True</verdict> or <verdict>False</verdict>."""
```
roaga, line 283: Is this strong enough language for what we're trying to measure?

roaga (approved):
LGTM, but open questions

src/seer/automation/autofix/evaluations.py
jennmueng: Hmm I think no, as we are looking for a complete boolean of whether this solution fixes the issue or not, this way the average across multiple run items that we get on the item dashboard actually is a "% fixed"

src/seer/automation/autofix/evaluations.py
jennmueng: hmm, does it need to be stronger given the above prompting already?

PR #1419, opened 2024-11-11
fix(autofix): preserve trailing newlines in files when applying patches

jennmueng:
Autofix was removing trailing newlines when applying files for pr creation.

The added unit test fails without this change, passes with the change.

### DIFF
--- a/src/seer/automation/models.py
+++ b/src/seer/automation/models.py
@@ -555,7 +555,14 @@ class FilePatch(BaseModel):
             return None
 
         # For M type
-        return self._apply_hunks(file_contents.splitlines(keepends=True))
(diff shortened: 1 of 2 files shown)


roaga (commented):
This solves the case where we accidentally remove newlines at the end, but does it account for the case where we actually have a diff change that is supposed to remove newlines at the end?

jennmueng:
@roaga It doesn't, and is a limitation, but I don't see that being a case that autofix would do or be used for.

Because it's expected practice to keep trailing newlines at the end of code files, it should be maintaining that for customer code (when it already exists)

roaga (approved):

PR #1424, opened 2024-11-12
fix(autofix): Use source_start instead of target_start in hunk applies

jennmueng:
We should be using `source_start` when accessing a line index in the original file.
 
This hunk apply function was fine for simple changes, but when the change gets complex with multiple hunks, previous hunks will cause the resulting ("`target`") output to have more lines than the original ("`source`") file, causing it to try to access a line that was out of bounds.

These new test cases catch that case, and fail before this change is made.


roaga (approved):
Thanks for catching this!

PR #1439, opened 2024-11-15
fix(autofix): Store the correct agent memory in simple fixer

jennmueng:
We were storing the wrong agent's memory.


roaga (approved):
I don't think I understand how this caused the bug yet but this change makes sense

PR #1443, opened 2024-11-15
chore(dynamic alert thresholds): fix timeout of detection calls with …

ram-senth:
…more than 7 days of current window


src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -227,28 +227,47 @@ def _combo_detect(
             )
             raise ClientError("Insufficient history data")
 
+        orig_curr_len = len(ts_with_history.current)
         logger.info(
-            f"Detecting anomalies for time series with {len(ts_with_history.current)} datapoints and history of {len(ts_with_history.history)} datapoints"
+            f"Detecting anomalies for time series with {len(ts_with_history.history)} datapoints and history of {orig_curr_len} datapoints"
         )
+        trim_current_by = 0
+        MAX_STREAM_DETECTION_LEN = 400
```
aayush-se, line 235: Is there a reason this value is fixed to 400 and not based off of the ad_config?

aayush-se (approved):
Looks good

src/seer/anomaly_detection/anomaly_detection.py
ram-senth: This is more a performance tweak to ensure the preview plot is rendered. So kept it hardcoded at 400.

PR #1445, opened 2024-11-18
fix(autofix): Fix duplicate step creation on interactive plan+code

roaga:
When rethinking or giving feedback, plan+code would often create unnecessary extra DEFAULT (insight) steps. This caused lots of visual bugs.  #1444 

This PR adds a check to stop duplication, the same way the root cause step already does this.

### DIFF
--- a/src/seer/automation/autofix/event_manager.py
+++ b/src/seer/automation/autofix/event_manager.py
@@ -130,7 +130,10 @@ class AutofixEventManager:
 
     def send_coding_start(self):
         with self.state.update() as cur:
-            plan_step = cur.add_step(self.plan_step)
+            plan_step = cur.find_or_add(self.plan_step)
+            if plan_step.status != AutofixStatus.PROCESSING:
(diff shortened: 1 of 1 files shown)


src/seer/automation/autofix/event_manager.py
```diff
@@ -130,7 +130,10 @@ def set_selected_root_cause(self, payload: AutofixRootCauseUpdatePayload):
 
     def send_coding_start(self):
         with self.state.update() as cur:
-            plan_step = cur.add_step(self.plan_step)
+            plan_step = cur.find_or_add(self.plan_step)
```
jennmueng, line 133:
```suggestion
            plan_step = cur.find_step(key=self.plan_step.key)
```
Would it be cleaner to do explicit finding then adding only if it doesn't exist here?

roaga: I did that to use the same exact logic as the root cause step, but I've updated both now because you're right that is cleaner.

jennmueng (approved):

PR #1455, opened 2024-11-19
feat(anomaly detection): Cron to cleanup disabled alerts

aayush-se:
- Currently, cleanup only happens upon detection call leaving stale timestamps for disabled alerts to accumulate
- Created cron which runs weekly to cleanup (delete) disabled alerts based on when they were last queued for detection (28 days) or if last_queued is null then deletes old alerts
- Cleaning alerts as a byproduct cleans the timeseries that are associated with them because of the parent-child relationship

### DIFF
--- a/src/seer/anomaly_detection/tasks.py
+++ b/src/seer/anomaly_detection/tasks.py
@@ -1,5 +1,6 @@
 import logging
 from datetime import datetime, timedelta
+from operator import and_, or_
 
 import numpy as np
 import sentry_sdk
@@ -116,20 +117,32 @@ def toggle_data_purge_flag(alert_id: int):
 @sentry_sdk.trace
 def cleanup_disabled_alerts():
 
-    logger.info("Cleaning up timeseries data for alerts that have been inactive for over 28 days")
+    logger.info(
+        "Cleaning up timeseries data for alerts that have been inactive (detection has not been run) for over 28 days"
+    )
 
     date_threshold = datetime.now() - timedelta(days=28)
 
     with Session() as session:
-        # Get all alerts that haven't been queued in the last 28 days indicating that they are disabled and are safe to cleanup
+        # Get and delete alerts that haven't been queued for detection in the last 28 days indicating that they are disabled and are safe to cleanup
         alerts = (
             session.query(DbDynamicAlert)
-            .filter(DbDynamicAlert.last_queued_at < date_threshold)
+            .filter(
+                or_(
+                    DbDynamicAlert.last_queued_at < date_threshold,
+                    and_(
+                        DbDynamicAlert.last_queued_at.is_(None),
+                        DbDynamicAlert.created_at < date_threshold,
+                    ),
+                )
+            )
             .all()
         )
 
-        deleted_count = 0
+        deleted_count = len(alerts)
+
         for alert in alerts:
-            deleted_count += delete_old_timeseries_points(alert, date_threshold.timestamp())
+            session.delete(alert)
+
         session.commit()
-        logger.info(f"Deleted {deleted_count} timeseries points")
+        logger.info(f"Deleted {deleted_count} alerts")
--- a/src/seer/anomaly_detection/tasks.py
+++ b/src/seer/anomaly_detection/tasks.py
@@ -117,11 +117,11 @@ def toggle_data_purge_flag(alert_id: int):
 @sentry_sdk.trace
 def cleanup_disabled_alerts():
 
-    logger.info(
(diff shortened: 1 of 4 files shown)


src/seer/anomaly_detection/tasks.py
```diff
@@ -109,3 +110,26 @@ def toggle_data_purge_flag(alert_id: int):
         )
         alert.data_purge_flag = new_flag
         session.commit()
+
+
+@celery_app.task
+@sentry_sdk.trace
+def cleanup_disabled_alerts():
+
+    logger.info("Cleaning up timeseries data for alerts that have been inactive for over 28 days")
```
ram-senth, line 119: By inactive we mean that anomaly detection was not called for the alert, right? Do we want to make that explicit here?

src/seer/anomaly_detection/tasks.py
```diff
@@ -109,3 +110,26 @@ def toggle_data_purge_flag(alert_id: int):
...
+
+    date_threshold = datetime.now() - timedelta(days=28)
+
+    with Session() as session:
+        # Get all alerts that haven't been queued in the last 28 days indicating that they are disabled and are safe to cleanup
+        alerts = (
+            session.query(DbDynamicAlert)
+            .filter(DbDynamicAlert.last_queued_at < date_threshold)
```
ram-senth, line 127: Do we need to handle the edge case where the last_queued_at is null, as will be the case until the first detection happens?

src/seer/anomaly_detection/tasks.py
```diff
@@ -109,3 +110,26 @@ def toggle_data_purge_flag(alert_id: int):
...
+            .all()
+        )
+
+        deleted_count = 0
+        for alert in alerts:
+            deleted_count += delete_old_timeseries_points(alert, date_threshold.timestamp())
```
ram-senth, line 133: So we just delete the time series points but the actual entry in the alert table will still remain? Should we delete that as well?

tests/seer/anomaly_detection/test_cleanup_tasks.py
```diff
@@ -154,3 +154,43 @@ def test_cleanup_timeseries(self):
...
+        # Set last_queued_at to be over 28 days ago for alerts 1 and 2
+        with Session() as session:
+            for alert_id in [external_alert_id1, external_alert_id2]:
+                alert = (
+                    session.query(DbDynamicAlert)
+                    .filter(DbDynamicAlert.external_alert_id == alert_id)
+                    .one_or_none()
+                )
+                assert alert is not None
+                alert.last_queued_at = datetime.now() - timedelta(days=29)
```
ram-senth, line 173: We should add a unit that includes one alert with last_queued_at as null

src/seer/anomaly_detection/tasks.py
aayush-se: The default value for `last_queued_at` is the current time so that value should be filled in upon creation of the alert regardless of detection

src/seer/anomaly_detection/tasks.py
aayush-se: I left it in case they re-enable the alert. For example, if they go from dynamic --> static --> dynamic, then the alert entry should still be in the table right? But if deleting the alert then re-creating it is fine, then I can also delete the alert itself within this cron

ram-senth: Yes it will recreate the alert if revived. So we should delete it as well.

src/seer/anomaly_detection/tasks.py
aayush-se: Actually, I will include a null check in here because of the old alerts that were created before the data pruning logic was implemented likely resulting in some null values

src/seer/anomaly_detection/tasks.py
aayush-se: Directly deleted the alerts which in turn also deletes the associated timeseries data from `DbDynamicAlertTimeSeries`

src/seer/anomaly_detection/tasks.py
```diff
@@ -109,3 +111,38 @@ def toggle_data_purge_flag(alert_id: int):
...
         alert.data_purge_flag = new_flag
         session.commit()
+
+
+@celery_app.task
+@sentry_sdk.trace
+def cleanup_disabled_alerts():
+
+    logger.info(
+        "Cleaning up timeseries data for alerts that have been inactive (detection has not been run) for over 28 days"
```
ram-senth, line 121: Minor nit - add  `date_threshold` in this info message

ram-senth (approved):

PR #1484, opened 2024-11-21
fix(autofix): Handle empty messages and move cleaning calls inside LLMClient

### LINKED ISSUE #1483: (autofix) Fix Anthropic message errors
https://sentry.sentry.io/issues/6066997870/

roaga:
1. Adds a cleaning util for LLM memory that fixes empty content messages
2. Moves the `clean_tool_call_assistant_messages` inside `generate_structured` and `generate_text` so that we don't have to think about cleaning memory when using the LLMClient

Fixes #1483

### DIFF
--- a/src/seer/automation/agent/client.py
+++ b/src/seer/automation/agent/client.py
@@ -474,9 +474,9 @@ class LlmClient:
         default_temperature = defaults.temperature if defaults else None
         # More defaults to come
 
-        messages = self.clean_message_content(messages if messages else [])
+        messages = LlmClient.clean_message_content(messages if messages else [])
         if not tools:
-            messages = self.clean_tool_call_assistant_messages(messages)
+            messages = LlmClient.clean_tool_call_assistant_messages(messages)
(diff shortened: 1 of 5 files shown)


src/seer/automation/agent/client.py
```diff
@@ -532,8 +539,7 @@ def generate_structured(
         else:
             raise ValueError(f"Invalid provider: {model.provider_name}")
 
-    @staticmethod
-    def clean_tool_call_assistant_messages(messages: list[Message]) -> list[Message]:
+    def clean_tool_call_assistant_messages(self, messages: list[Message]) -> list[Message]:
```
jennmueng, line 542: If this is not using `self` and is still used outside the class, would it be not cleaner to just keep it static?

roaga: Made them static now because they don't leverage self, but currently they are not used outside of this class.

jennmueng (approved):

PR #1493, opened 2024-11-22
fix(autofix): Add fallback to non-draft pr

### LINKED ISSUE #1492: (autofix) Handle repos that don't support draft PRs
https://sentry.sentry.io/issues/6082349622/

Should just fall back to a normal PR

roaga:
If our attempt to create a draft PR is rejected by the API, because the user's repo doesn't support drafts, we try again with a regular PR.
Fixes #1492


jennmueng (approved):
Ideally there would be a way for us to just check if draft prs are supported for a repo or not, but on a quick search looks like there isn't one

PR #1512, opened 2024-11-28
feat(autofix) create a single commit for the suggested fix

Mitan:
This PR changes the autofix logic to create a single commit for the suggested fix.

- All the files changes are grouped together using the Github data API.
- The original commit messages might not be meaningful after user-suggested patching, so they are not used

### DIFF
--- a/src/seer/automation/codebase/repo_client.py
+++ b/src/seer/automation/codebase/repo_client.py
@@ -363,7 +363,6 @@ class RepoClient:
             patch_type = "M"
         
         contents = self.repo.get_contents(path, ref=branch_ref) if patch_type != "A" else None
-        # logger.info(f"Type of contents is {type(contents)} {contents.encoding} {contents.decoded_content}, {contents.content}")
         detected_encoding = detect_encoding(contents.decoded_content) if contents else None
 
         if isinstance(contents, list):
--- a/src/seer/automation/codebase/repo_client.py
+++ b/src/seer/automation/codebase/repo_client.py
@@ -16,6 +16,7 @@ from unidiff import PatchSet
 from seer.automation.autofix.utils import generate_random_string, sanitize_branch_name
 from seer.automation.codebase.utils import get_language_from_path
 from seer.automation.models import FileChange, FilePatch, InitializationError, RepoDefinition
+from seer.automation.utils import detect_encoding
 from seer.configuration import AppConfig
 from seer.dependency_injection import inject, injected
 from seer.utils import class_method_lru_cache
@@ -281,6 +282,7 @@ class RepoClient:
         data = requests.get(comparison.diff_url, headers=self._get_auth_headers(accept_type="diff"))
         data.raise_for_status()  # Raise an exception for HTTP errors
 
+        # todo note utf-8 hardcoded here
         patch_set = PatchSet(data.content.decode("utf-8"))
 
         added_files = [patch.path for patch in patch_set.added_files]
@@ -306,6 +308,7 @@ class RepoClient:
             if isinstance(contents, list):
                 raise Exception(f"Expected a single ContentFile but got a list for path {path}")
 
+            # todo note this assumes utf-8 encoding
             return contents.decoded_content.decode()
         except Exception as e:
             logger.exception(f"Error getting file contents: {e}")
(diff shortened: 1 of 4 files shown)


Mitan:
Manually tested by patching several files for add, delete and modify
Screenshot 2024-12-02 at 1 52 50 PM

src/seer/automation/codebase/repo_client.py
```diff
@@ -337,27 +337,27 @@ def _create_branch(self, branch_name):
...
-        if not patch and not change:
-            raise ValueError("Either patch or change must be provided")
-
+    def get_one_file_autofix_change( self, *, branch_ref: str,
+                               patch: FilePatch | None = None,
+                               change: FileChange | None = None) -> InputGitTreeElement | None:
+        """
+        This method is used to get a single change to be committed by Autofix. 
+        It supports both FilePatch and FileChange objects.
+        """
```
roaga, line 346: Is this really only used by Autofix? The repo_client utils are used by Autofix and the Codecov unit test gen as well.

src/seer/automation/codebase/repo_client.py
```diff
@@ -372,28 +372,15 @@ def _commit_file_change(
...
-                new_contents or "",
-                contents.sha,  # FYI: It wants the sha of the content blob here, not a commit sha.
-                branch=branch_ref,
-            )
+        # don't create a blob if the file is being deleted
+        blob = self.repo.create_git_blob(new_contents, "utf-8") if new_contents else None
+        
+        return InputGitTreeElement(
+            path=path,
+            mode='100644',
```
roaga, line 380: What's this mode number?

src/seer/automation/codebase/repo_client.py
```diff
@@ -372,28 +372,15 @@ def _commit_file_change(
...
-                path,
-                commit_message or "File change",
-                new_contents or "",
-                contents.sha,  # FYI: It wants the sha of the content blob here, not a commit sha.
-                branch=branch_ref,
-            )
+        # don't create a blob if the file is being deleted
+        blob = self.repo.create_git_blob(new_contents, "utf-8") if new_contents else None
+        
+        return InputGitTreeElement(
```
roaga, line 378: I'm concerned about this encoding. We had an issue previously where we couldn't read files in certain encodings, and we fixed it by detecting the encoding in advance (see https://github.com/getsentry/seer/pull/1495). So I would think that when writing code files, we'd want to maintain the original encoding, not change it to utf-8.

src/seer/automation/codebase/repo_client.py
```diff
@@ -372,28 +372,15 @@ def _commit_file_change(
...
+            path=path,
+            mode='100644',
+            type='blob',
+            sha=blob.sha if blob else None)
+    
 
     def create_branch_from_changes(
```
roaga, line 385: this doesn't seem to maintain support for the Codecov/unit test gen use case -- am I mistaken?

src/seer/automation/codebase/repo_client.py
Mitan: it's the git magic number for a regular non-exectuable file,  i will add a comment for it
https://stackoverflow.com/questions/737673/how-to-read-the-mode-field-of-git-ls-trees-output

src/seer/automation/codebase/repo_client.py
Mitan: Turns out it's not implemented for autofix, I will create an issue for it and fix in this PR

src/seer/automation/codebase/repo_client.py
Mitan: Renamed the method

src/seer/automation/codebase/repo_client.py
```diff
@@ -337,33 +310,37 @@ def _create_branch(self, branch_name):
...
         if patch_type == "create":
             patch_type = "A"
         elif patch_type == "delete":
             patch_type = "D"
         elif patch_type == "edit":
             patch_type = "M"
-        commit_message = change.commit_message if change else None
 
+        # todo there is some code duplication with get_file_content method
+        #  the difference is that detected_encoding is also used later in blog creation
```
roaga, line 336: Is this TODO still valid? Can we either address it or remove it?

src/seer/automation/codebase/repo_client.py
```diff
@@ -410,22 +376,37 @@ def create_branch_from_changes(
...
-
-        branch_ref.update()
+                    logger.exception(f"Error processing file change: {e}")
+        # latest commit is the head of new branch
+        latest_commit = self.repo.get_git_commit(self.get_branch_head_sha(new_branch_name))
+        base_tree = latest_commit.tree
+        new_tree = self.repo.create_git_tree(tree_elements, base_tree)
+        
+        new_commit = self.repo.create_git_commit(
+            message="Changes suggested by Autofix",
```
roaga, line 404: Are we still asking the Autofix agent to generate commit messages elsewhere? If so, maybe we could remove that since they seem unused now.

We could also use the title of the PR as the commit message, no?

roaga (approved):
Approving to unblock, but left some cleanup comments that I think should be addressed before merging

src/seer/automation/codebase/repo_client.py
Mitan: Thanks, the duplicated code is just a couple of statements and there is a different logic with passing the detected encoding further, so I will remove the todo

src/seer/automation/codebase/repo_client.py
roaga: What I meant was that in the old approach, where we had multiple commits, we asked an LLM to describe each commit. Is that LLM generation still happening somewhere?

PR #1528, opened 2024-12-03
chore(anomaly_detection): config refactor

ram-senth:


aayush-se (approved):
Nice changes will set us up for visualizing the prophet thresholds easily

PR #1575, opened 2024-12-09
fix(autofix): Bump updated_at BEFORE throwing the state in the dbmodel

jennmueng:
Runs were timing out because there actually are two `updated_at` fields, the main one being used in the class instances are the ones inside the `value` json. We're supposed to be syncing the field on the db model with the one from the json for the sake of db querying, but there was a bug in the big PR that updates this field in the value AFTER the value json was set.

So the updated_at inside the json was the original one, never bumped.

### DIFF
--- a/src/seer/automation/state.py
+++ b/src/seer/automation/state.py
@@ -104,6 +104,12 @@ class DbState(State[_State]):
         """
         pass
 
+    def before_update(self, value: _State):
+        """
+        Can be used to run some logic before the update is applied to the db
+        """
+        pass
+
     @contextlib.contextmanager
     def update(self):
         """
@@ -120,6 +126,7 @@ class DbState(State[_State]):
             assert r
             value = self.model.model_validate(r.value)
             yield value
+            self.before_update(value)
             db_state = DbRunState(id=self.id, value=value.model_dump(mode="json"))
             self.apply_to_run_state(value, db_state)
             session.merge(db_state)
--- a/tests/automation/autofix/test_autofix_state.py
+++ b/tests/automation/autofix/test_autofix_state.py
@@ -7,24 +7,6 @@ from seer.automation.autofix.state import ContinuationState
 from seer.automation.state import DbStateRunTypes
 
 
-def test_update_preserves_last_triggered():
-    """Test that update preserves last_triggered_at while updated_at is set to current time"""
-    continuation = next(generate(AutofixContinuation))
-    trigger_time = datetime.datetime(2023, 1, 2)
-    continuation.last_triggered_at = trigger_time
-    state = ContinuationState.new(group_id=1, value=continuation, t=DbStateRunTypes.AUTOFIX)
-
-    with state.update():
-        # Don't make any changes
-        pass
-
-    result = state.get()
-    assert result.last_triggered_at == trigger_time
-    # updated_at should be set to current time
-    assert isinstance(result.updated_at, datetime.datetime)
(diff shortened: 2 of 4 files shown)


tests/automation/autofix/test_autofix_state.py
```diff
@@ -7,16 +7,53 @@
...
+def test_update_preserves_last_triggered():
+    """Test that update preserves last_triggered_at while updated_at is set to current time"""
+    continuation = next(generate(AutofixContinuation))
+    trigger_time = datetime.datetime(2023, 1, 2)
+    continuation.last_triggered_at = trigger_time
+    state = ContinuationState.new(group_id=1, value=continuation, t=DbStateRunTypes.AUTOFIX)
+
+    with state.update() as cur:
+        # Even if we try to set updated_at, it will be overridden by before_update
+        cur.last_triggered_at = trigger_time
```
roaga, line 19: I'm confused by this test. Didn't you already set `last_triggered_at = trigger_time` above? So what is this testing?

src/seer/automation/state.py
```diff
@@ -104,6 +104,12 @@ def apply_to_run_state(self, value: _State, run_state: DbRunState):
         """
         pass
 
+    def before_update(self, value: _State):
```
roaga, line 107: why is the param named `value` here but `state` in the implementations?

Also can you not just call `state.mark_updated()` in here instead of duplicating in the sub classes?

roaga (commented):
Looks good, but some nits. And more generally it feels weird that we were/are maintaining two separate updated_at fields, but I guess it's fine to keep things the way they are

src/seer/automation/state.py
jennmueng: - It's `value` here to match with what the field is on the db model
- `_State` doesn't have `mark_updated`. It also makes more sense to do it this way to support more functionality specific to each state type

tests/automation/autofix/test_autofix_state.py
```diff
@@ -7,16 +7,53 @@
...
+    state = ContinuationState.new(group_id=1, value=continuation, t=DbStateRunTypes.AUTOFIX)
+
+    with state.update():
+        # Don't make any changes
+        pass
+
+    # Even without changes, updated_at should still be updated
+    result = state.get()
+    assert result.updated_at > initial_time
+    assert isinstance(result.updated_at, datetime.datetime)
```
roaga, line 59: Is this test not a duplicate now?

tests/automation/autofix/test_autofix_state.py
```diff
@@ -7,16 +7,53 @@
...
+
+    with state.update():
+        # Don't make any changes
+        pass
+
+    result = state.get()
+    assert result.last_triggered_at == trigger_time
+    # updated_at should be set to current time
+    assert isinstance(result.updated_at, datetime.datetime)
+    assert result.updated_at > datetime.datetime(2023, 1, 1)
```
roaga, line 25: where did you pull this date from for this assertion? You defined `trigger_time` as Jan 2nd, not Jan 1st, above. So couldn't `updated_at` pass this even without calling `state.update()`?

tests/automation/autofix/test_autofix_state.py
jennmueng: you're right, this test was one of the ai generated ones lol, let's just dump it

roaga (approved):

PR #1588, opened 2024-12-09
fix(anomaly detection): Use thresholds for dynamic window

aayush-se:
- `thresholds` and `original_flags` for the respective detector (SuSS vs. fixed) were not being properly used and therefore defaulting to just SuSS
- Ensure that use_suss flag is being properly used (entire history was being sent not the required portion)
- Clean up `combine_anomalies` for readability and performance (if no fixed window is being used, should not perform list iteration)

### DIFF
--- a/src/seer/anomaly_detection/accessors.py
+++ b/src/seer/anomaly_detection/accessors.py
@@ -387,29 +387,17 @@ class DbAlertDataAccessor(AlertDataAccessor):
         combined_thresholds = anomalies_suss.thresholds
         combined_original_flags = anomalies_suss.original_flags
         if anomalies_fixed is not None:
-            combined_flags = [
-                (anomalies_suss.flags[i] if use_suss[i] else anomalies_fixed.flags[i])
-                for i in range(len(anomalies_suss.flags))
-            ]
-
-            combined_scores = [
-                (anomalies_suss.scores[i] if use_suss[i] else anomalies_fixed.scores[i])
-                for i in range(len(anomalies_suss.scores))
-            ]
-
-            combined_thresholds = [
-                (anomalies_suss.thresholds[i] if use_suss[i] else anomalies_fixed.thresholds[i])
-                for i in range(len(anomalies_suss.thresholds))
-            ]
-
-            combined_original_flags = [
-                (
-                    anomalies_suss.original_flags[i]
-                    if use_suss[i]
-                    else anomalies_fixed.original_flags[i]
-                )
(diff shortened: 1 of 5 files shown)


src/seer/anomaly_detection/accessors.py
```diff
@@ -382,29 +382,34 @@ def combine_anomalies(
...
+        combined_thresholds = anomalies_suss.thresholds
+        combined_original_flags = anomalies_suss.original_flags
+        if anomalies_fixed is not None:
+            for i in range(len(anomalies_suss.flags)):
+                if use_suss[i]:
+                    combined_flags[i] = anomalies_suss.flags[i]
+                    combined_scores[i] = anomalies_suss.scores[i]
+                    combined_original_flags[i] = anomalies_suss.original_flags[i]
+                    if i < len(anomalies_suss.thresholds):
+                        combined_thresholds[i] = anomalies_suss.thresholds[i]
```
aayush-se, line 396: Thresholds can be less than the flags because of the case in batch detection, we have 1 threshold for all of the points so this check prevents index errors

tests/seer/anomaly_detection/test_cleanup_tasks.py
```diff
@@ -167,7 +167,6 @@ def test_cleanup_timeseries(self):
                         and new.anomaly_algo_data["mp_suss"]["l_idx"] == algo_data["l_idx"]
                         and new.anomaly_algo_data["mp_suss"]["r_idx"] == algo_data["r_idx"]
                     )
-                    assert new.anomaly_algo_data["original_flag"] == algo_data["original_flag"]
```
aayush-se, line 170: Removed this line because during the recomputation of the matrix profiles after pruning, the flags indeed can change resulting in new "original_flags"

ram-senth (approved):

src/seer/anomaly_detection/accessors.py
```diff
@@ -382,29 +382,44 @@ def combine_anomalies(
...
+        combined_original_flags = anomalies_suss.original_flags
+        if anomalies_fixed is not None:
+            combined_flags = [
+                (anomalies_suss.flags[i] if use_suss[i] else anomalies_fixed.flags[i])
                 for i in range(len(anomalies_suss.flags))
-            ],
-            scores=[
+            ]
+
+            combined_scores = [
```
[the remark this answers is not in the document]
aayush-se: I believe that is part of the outdated diff. They are all now combined into one loop

Issue #1549, opened 2024-12-05
(autofix) Detect endcoding correctly in autofix


Mitan:
I'm concerned about this encoding. We had an issue previously where we couldn't read files in certain encodings, and we fixed it by detecting the encoding in advance (see https://github.com/getsentry/seer/pull/1495). So I would think that when writing code files, we'd want to maintain the original encoding, not change it to utf-8.

_Originally posted by @roaga in https://github.com/getsentry/seer/pull/1512#discussion_r1871746126_
            

Currently autofix assumes utf-8 encoding for all file content. This logic needs to be aligned with the one used for Codecov from https://github.com/getsentry/seer/pull/1495

PR #1598, opened 2024-12-10
Refact get_file_content method

Mitan:
There is some code duplication for get file content method in repo_client file. Since encoding detection was introduced in https://github.com/getsentry/seer/issues/1549, some refact is needed for get_file_content method to pass the encoding param through.


src/seer/automation/codebase/code_search.py
```diff
@@ -60,7 +60,11 @@ def _read_file_with_encoding(self, file_path: str) -> Optional[List[str]]:
         try:
             # Read only first 32KB to detect encoding for large files
             with open(file_path, "rb") as f:
-                raw_data = f.read(32768)
+                # note the number of bytes to read is increased from 32KB to 256KB
+                # theoretically, this could cause an incorrectly detected encoding if there are some special characters
+                # in the file after the first 256KB of data
+                # Hopefully it is a corner case, especially since files larger than 1MB are ignored from search
+                raw_data = f.read(262144)
```
roaga, line 67: why the increase?

tests/automation/autofix/components/coding/test_coding_component.py
```diff
@@ -254,12 +254,13 @@ def test_invoke_with_missing_and_existing_files(
         component._is_obvious = MagicMock(return_value=False)
 
         # Simulate file content for different scenarios
+        # updated the return values to include encoding
```
roaga, line 257: unnecessary comment

roaga (approved):

src/seer/automation/codebase/code_search.py
Mitan: It was suggested by Jenn, as a way to minimize the chances  to detect encodding incorrectly because of hitting some weird characters in the end of the file (we discussed during the standup).

tests/automation/autofix/components/coding/test_coding_component.py
Mitan: thanks, will remove

PR #1616, opened 2024-12-12
Updated README for onboarding

Mitan:


README.md
```diff
@@ -30,9 +33,9 @@ These instructions require access to internal Sentry resources and are intended
 Download model artifacts:
 
 ```bash
-gsutil cp -r gs://sentry-ml/seer/models ./models
+gsutil cp -r gs://sentry-ml/seer/models .
 ```
-
+If you see a prompt "Reauthentication required. Please insert your security key and press enter...", re-authenticate using the command `gcloud auth login` and set the project id to the one for Seer.
```
roaga, line 38: You could also try `gcloud auth application-default login` to avoid setting the project id in some cases

roaga (approved):
Thanks for the updates!

README.md
Mitan: _application-default_ didn't work for me there and was causing that "Reauthentication required. Please insert your security key and press enter..." prompt. But if you remove _application-default_, the problem solved

PR #1617, opened 2024-12-12
feat(autofix): Stream output in Autofix agent

roaga:
Replaces Autofix agent LLM calls with streaming. Saves the output as it comes in the step object in the run state. All other functionality is identical.

I added a function for streaming with OpenAI as well, and tested that it works, but we don't use it at the moment. Just future proofing.

### DIFF
--- a/src/seer/automation/autofix/autofix_agent.py
+++ b/src/seer/automation/autofix/autofix_agent.py
@@ -70,9 +70,9 @@ class AutofixAgent(LlmAgent):
             )
 
     def get_completion(self, run_config: RunConfig):
-    """
-    Streams the preliminary output to the current step and only returns when output is complete
-    """
+        """
+        Streams the preliminary output to the current step and only returns when output is complete
+        """
         content_chunks = []
         tool_calls = []
         usage = Usage()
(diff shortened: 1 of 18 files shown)


src/seer/automation/autofix/components/root_cause/component.py
```diff
@@ -128,7 +128,7 @@ def invoke(
                     for i, item in enumerate(cause_model.code_context):
                         item.id = i
                         # Find line range for the snippet
-                        if item.snippet.file_path and item.snippet.snippet:
```
roaga, line 131: tacking on a fix for an unrelated bug

src/seer/automation/autofix/components/root_cause/component.py
```diff
@@ -128,7 +128,7 @@ def invoke(
...
+                        if item.snippet and item.snippet.file_path and item.snippet.snippet:
```
roaga, line 131: fix for unrelated bug

trillville (approved):

src/seer/automation/autofix/autofix_agent.py
```diff
@@ -63,6 +69,49 @@ def _check_prompt_for_help(self, run_config: RunConfig):
                 "You're taking a while. If you need help, ask me a concrete question using the tool provided."
             )
 
+    def get_completion(self, run_config: RunConfig):
```
jennmueng, line 72: Little comment for cleanliness so folks grok right away why this method is special for the autofix agent

```suggestion
    def get_completion(self, run_config: RunConfig):
    """
    Streams the preliminary output to the current step and only returns when output is complete
    """
```

jennmueng (approved):
LGTM, well tested! One little suggestion

PR #1626, opened 2024-12-13
fix(autofix): Cache repo client objects to avoid re-initialization

roaga:
See [this profile](https://sentry.sentry.io/profiling/profile/seer/ba56c3e354674bf5bc18c9ca696400e9/flamegraph/?colorCoding=by%20system%20vs%20application%20frame&eventId=acec842a014c44818c8bc5a7263abf65&fov=2984499968,38,7125656064,17&query=&sorting=call%20order&tid=138847897326720&view=top%20down).

We were creating new repo clients over and over again, and each time, it added 1-1.5s.
- before invoking a step, ~1.5s redundant
- pre-filling coding step memory: ~5s (more for more files)
- every expand_document tool call: ~1s

This PR stores created repo clients in memory in the context so that throughout a single step, we only need to initialize one repo client for each repo.

### DIFF
--- a/src/seer/automation/autofix/autofix_context.py
+++ b/src/seer/automation/autofix/autofix_context.py
@@ -69,6 +69,7 @@ class AutofixContext(PipelineContext):
         self.organization_id = request.organization_id
         self.project_id = request.project_id
         self.repos = request.repos
+        self._repo_clients: dict[tuple[str, RepoClientType], RepoClient] = {}
 
         self.sentry_client = sentry_client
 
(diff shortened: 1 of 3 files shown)


trillville (approved):
🚀

src/seer/automation/autofix/autofix_context.py
```diff
@@ -134,33 +135,33 @@ def get_repo_client(
...
         repo_external_id: str | None = None,
         type: RepoClientType = RepoClientType.READ,
-    ):
+    ) -> RepoClient:
         """
         Gets a repo client for the current single repo or for a given repo name.
         If there are more than 1 repos, a repo name must be provided.
         """
-        repo_client: RepoClient | None = None
+        # Find the repo definition
```
roaga, line 143:
```suggestion
```

roaga:
Callout from Ram: is the RepoClient holding open any connections? If so, that would be a problem.

However, I think it's not doing that. It's just wrapping HTTP requests.

But @jennmueng can you confirm?

jennmueng (commented):
Nice find! `functools.cache` might be a simpler solution. Also, yeah it should just be http requests from the github client!

src/seer/automation/codebase/repo_client.py
```diff
@@ -118,6 +119,7 @@ class RepoClient:
     def __init__(
         self, app_id: int | str | None, private_key: str | None, repo_definition: RepoDefinition
     ):
+        print("NEW REPO CLIENT")  # TODO
```
jennmueng, line 122: print?

jennmueng (approved):

Issue #1630, opened 2024-12-14
(autofix) Investigate how to remove slow down from waiting for threadpool


roaga:
https://sentry.sentry.io/profiling/profile/seer/74268052129c4acc90f8a7801b96830c/flamegraph/?colorCoding=by%20system%20vs%20application%20frame&fov=17989857280,39,5821208576,17&query=manage_run&sorting=call%20order&tid=133443867681920&view=top%20down

Seeing the above profile, the new `manage_run` causes a 6 second delay waiting for an insight card to finish. 

Can we increase the workers to 2 so insights show up faster?
Can we not start a new insight sharing thread when we know we're about to wrap up?

PR #1631, opened 2024-12-14
fix(autofix): Prevent long wait times for insights threadpool at end of run

roaga:
Issue #1630 

I discovered in profiling that when the root cause step completed, there was sometimes a huge wait (the case I saw was 6 seconds) for the thread pool to finish it's last insight card generation, before moving onto the root cause formatter. (similar for the coding step)

Why? Because we started a new insight sharing thread after every LLM call, including the final call of the agent run. Due to Zach's changes that make us wait for threads to complete before ending the run, this caused big delays.

Solutions in this PR:
1. Don't generate an insight card on the final output of the run. That was somewhat redundant anyway if we're about to show the final output card, plus this avoids starting a new thread from scratch that blocks everything else.
2. Adds a second thread worker. That way while running in parallel with the main agent, one insight doesn't have to wait for the previous to complete, and it's unlikely that insights will "fall behind" the main agent.

We could completely eliminate this delay if we didn't wait for the threads to be completed, and I'm not sure that's a bad idea, but I'm still assuming we don't want to sacrifice that safety. This PR mitigates most of the issue while maintaining thread safety.

### DIFF
--- a/src/seer/automation/autofix/autofix_agent.py
+++ b/src/seer/automation/autofix/autofix_agent.py
@@ -135,14 +135,11 @@ class AutofixAgent(LlmAgent):
 
         # log thoughts to the user
         cur = self.context.state.get()
-        initial_memory_length = cur.steps[-1].initial_memory_length
         if (
             completion.message.content  # only if we have something to summarize
             and self.config.interactive  # only in interactive mode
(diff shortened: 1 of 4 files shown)


jennmueng (approved):
awesome!!

PR #1633, opened 2024-12-16
feat(celery): Set celery worker queue via CELERY_WORKER_QUEUE

jennmueng:
Setting `CELERY_WORKER_QUEUE` env var will make that instance of seer both publish tasks to and the celery worker consume from the queue with that name.

Will follow up w/ SRE to set this value.

### DIFF
--- a/src/celery_app/config.py
+++ b/src/celery_app/config.py
@@ -1,4 +1,3 @@
-from enum import StrEnum
 from typing import Any
 
 from seer.bootup import module
@@ -10,11 +9,6 @@ class CeleryConfig(dict[str, Any]):
     pass
 
 
-class CeleryQueues(StrEnum):
-    DEFAULT = "seer"
-    CUDA = "seer-cuda"
-
-
 @module.provider
 def celery_config(app_config: AppConfig = injected) -> CeleryConfig:
     return CeleryConfig(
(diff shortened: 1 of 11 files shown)


src/celery_app/config.py
```diff
@@ -23,11 +17,11 @@ def celery_config(app_config: AppConfig = injected) -> CeleryConfig:
...
         enable_utc=True,
-        task_default_queue=CeleryQueues.DEFAULT,
+        task_default_queue=app_config.CELERY_WORKER_QUEUE,
         task_queues={
-            CeleryQueues.DEFAULT: {
-                "exchange": CeleryQueues.DEFAULT,
-                "routing_key": CeleryQueues.DEFAULT,
+            app_config.CELERY_WORKER_QUEUE: {
+                "exchange": app_config.CELERY_WORKER_QUEUE,
+                "routing_key": app_config.CELERY_WORKER_QUEUE,
```
aayush-se, line 24: 2 points:
- shouldn't the exchange and routing_key be separated/distinct values to define specific task routes and to ensure future flexibility
- I think it would be a good idea to keep this default queue as a fallback

jennmueng: On point #1, looks like for our use case, for our needs I don't forsee us needing specific exchange/routing key names, and our use case of just needing separate queues for different seer deployments should be satisfied with a single key being used for all these values. I think adding more environment variables might add more complexity to our setup when it's not needed. If you think it's worth the tradeoff for future flexibility then happy to implement, no strong opinions here.

#2, we tried using the default queue before and we ran into issues, don't fully recall what they were but it might have something to do with us using a shared rabbitmq cluster and main sentry also has celery. But this was why we set a default of `seer`

aayush-se: No strong opinions here either. Lets just keep it as is then and should be an easy update as needed in the future.

aayush-se (approved):

PR #1636, opened 2024-12-16
feat(autofix): Add google search tool

roaga:
Adds a tool to the Autofix agent that allows it to ask a question and receive an answer based on Google Search results, using Gemini 2.0 Flash.

This results in:
- a 19 second drop in total latency compared to the last eval ran on main (idk how reliable this is tho)
- a 1.5 point increase in "is root cause correct", now at 95%
- a 2 point increase in "is fixed:, now at 58%

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ If you see a prompt "Reauthentication required. Please insert your security key 
 * Run type checker: `make mypy`
 * Run tests: `make test`
 * Open a shell: `make shell`
+* Update `requirements.txt` based on `requirements-constraints.txt`: `make upgrade-package-versions`
 
 ### Reset Development Environment
 
--- a/requirements-constraints.txt
+++ b/requirements-constraints.txt
@@ -116,3 +116,5 @@ rapidfuzz==3.10.*
 pytest-vcr==1.*
 vcrpy==6.*
 chardet==5.2.*
+google-genai==0.*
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -80,6 +80,7 @@ If you see a prompt "Reauthentication required. Please insert your security key
 * Run type checker: `make mypy`
 * Run tests: `make test`
 * Open a shell: `make shell`
+* Update `requirements.txt` based on `requirements-constraints.txt`: `make upgrade-package-versions`
```
roaga, line 83: tacking on readme update for posterity

requirements-constraints.txt
```diff
@@ -116,3 +116,5 @@ rapidfuzz==3.10.*
 pytest-vcr==1.*
 vcrpy==6.*
 chardet==5.2.*
+google-genai==0.*
+httpx==0.27.2 # v0.28 is breaking OpenAI and Anthropic as of Dec 16 2024
```
roaga, line 120: this constraint on httpx could be lifted in the future, but not yet

roaga:
Q: is this going to work automatically in prod or do we need SRE to enable anything?

jennmueng (approved):
🚀 LGTM. Yes I think we'll need SRE to enable the Gemini Model Card? Also, this is pending security review so I'll approve but let's leave it draft to avoid merging until the above 2 checks are done

trillville (approved):

PR #1670, opened 2024-12-21
feat(anomaly detection): Maintain historical timeseries

aayush-se:
- Maintains historical data in a new table `DbDynamicAlertTimeSeriesHistory` which stores the required information for us to recreate and test any issues
    - `alert_id` : int
    - `timestamp` : int
    - `value` : float
    - `anomaly_type` : str
    - `saved_at` : datetime
- Cron runs once a week to clean up data in this table over 90 days old


src/migrations/versions/5a7ff418f726_migration.py
```diff
@@ -0,0 +1,37 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "dynamic_alert_time_series_history",
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column("alert_id", sa.BigInteger(), nullable=False),
+        sa.Column("timestamp", sa.DateTime(), nullable=False),
```
ram-senth, line 25: Since we are cleaning up history based on timestamp, do we need an index on this field?

src/migrations/versions/5a7ff418f726_migration.py
```diff
@@ -0,0 +1,37 @@
...
+down_revision = "3914e6cdb818"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "dynamic_alert_time_series_history",
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
```
ram-senth, line 23: Curious how come alert_id is biginteger but this id is integer? We will be having more rows in this table than the alert table, right?

src/seer/anomaly_detection/tasks.py
```diff
@@ -163,3 +188,17 @@ def cleanup_disabled_alerts():
...
+
+
+@celery_app.task
+@sentry_sdk.trace
+def cleanup_old_timeseries_history():
+    date_threshold = datetime.now() - timedelta(days=90)
+    with Session() as session:
+        deleted_count = (
+            session.query(DbDynamicAlertTimeSeriesHistory)
+            .filter(DbDynamicAlertTimeSeriesHistory.saved_at < date_threshold)
```
ram-senth, line 200: This condition is applied across all alerts, so can be a costly operation. Also, should we be filtering on `timestamp` instead of `save_at`?

src/migrations/versions/5a7ff418f726_migration.py
aayush-se: The id column is to avoid duplicate values in the rows since the other columns can be the same including the timestamped ones if they are added at the same time.

src/migrations/versions/5a7ff418f726_migration.py
aayush-se: Good catch -- fixed

src/seer/anomaly_detection/tasks.py
aayush-se: The reason I am filtering on `saved_at` instead of `timestamp` is because I thought that we want the timeseries points to be available 90 days after they are deleted, not created. But if that is not the case, I can change it to `timestamp` instead

src/migrations/versions/5a7ff418f726_migration.py
ram-senth: I was asking about the timestamp column. Since the where clause of the SQL will have timestamp, we need an index (not primary index as id will still be the primary index).

src/seer/anomaly_detection/tasks.py
aayush-se: Regarding the costly operation, is there another way to do this since this table doesn't have any specific relationships to other tables aside from it being a storage table?

ram-senth: I was thinking of having 90 days of time series data regardless of when it was saved to seer. As to the deletion operation, can you check if sqlalchemy has any batch deletion feature that we can leverage?

aayush-se: Filtering by `timeseries` and using the Delete object to remove values more efficiently:

- https://docs.sqlalchemy.org/en/14/core/dml.html#sqlalchemy.sql.expression.Delete.where
- https://stackoverflow.com/a/71038403

ram-senth (approved):

PR #1680, opened 2025-01-03
chore(anomaly detection): Convert exception to warning

aayush-se:
- Converts the exception into a warning which is logged
- This exception is thrown when the cleanup times out due to `compute_matrix_profile` taking a long time due to a cold start but the task is requeued and will succeed in the subsequent attempt if this occurs.


aayush-se:
- Using the `pool_pre_ping` strategy as implemented in #1699

PR #1681, opened 2025-01-06
test(autofix): Add more vcr-based tests for autofix scenarios

jennmueng:
The test cases for autofix tasks now are thorough and run the complete workflow.

Reimplements test cases for code/diff changes and user messages/interjections with no mocking.


src/seer/automation/codebase/repo_client.py
```diff
@@ -182,7 +182,7 @@ def check_repo_read_access(repo: RepoDefinition):
         return False
 
     @classmethod
-    @functools.cache
+    @functools.lru_cache(maxsize=8)
```
jennmueng, line 185: `@functools.cache` was breaking test cases as it was maintaining the cache across multiple test cases rather than cleaning it
> @functools.cache maintains its cache for the entire lifetime of the Python process
@functools.lru_cache creates a new cache instance each time the decorated function is redefined, which typically happens for each test module
This means that lru_cache naturally provides better test isolation out of the box, even without explicit cache clearing. This is one of the less-documented but important differences between the two decorators.

roaga: Is cache clearing not feasible instead? This sounds like it would cause more repo_client initialization calls and add lag back in some cases

tests/conftest.py
```diff
@@ -24,7 +24,8 @@
 @pytest.fixture(autouse=True, scope="session")
 def configure_environment():
     os.environ["LANGFUSE_HOST"] = ""  # disable Langfuse logging for tests
-    os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = "/app/tests/test_gcloud_credentials.json"
+    if os.environ.get("CI"):
+        os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = "/app/tests/test_gcloud_credentials.json"
```
roaga, line 28: oh? does this fix the need to comment this line out locally?

src/seer/automation/codebase/repo_client.py
jennmueng: lru cache in this use case should work exactly the same as the existing cache, the limit of 8 was set instead of an infinite size as I don't foresee there being more than 8 repos in an autofix run

tests/conftest.py
jennmueng: yeah this is so we are actually able to call vertex locally when making the vcr cassettes

src/seer/automation/codebase/repo_client.py
roaga: I'm thinking about the entire Python process vs. decorated function is redefined. In Autofix's context, when do new processes start? And when is the function redefined? Are they the same or different?

jennmueng: in a prod environment it should be equivalent as the function/class is defined when the process starts, the only difference is in testing environments

jennmueng: so in autofix prod, each celery process has its own cache for either the lru cache or cach method and they fundamentally would work the same

roaga: ok i trust u

roaga (dismissed):
Thank u sir 🚀

jennmueng:
Confirmed that the lru-cache only initializes the repo client once per celery worker process handling an autofix run (same as prior cache)

roaga (approved):
Thank you for double checking

Though RIP on the cassette merge conflicts

PR #1694, opened 2025-01-07
fix(autofix): Remove langfuse flushes to reduce cold starts

roaga:
Langfuse automatically flushes in the background. Correct me if I'm wrong, but I don't think we ever force terminate the program and would lose any data by relying on the automatic flushes.

These manual flushes block the main program and add ~750ms-1s of cold start on every LLM call, every keyword search, and every missing file fixer. (some profiles showed it taking up to 10s sometimes) Langfuse officially says to only use this when necessary as it is known to hurt performance.

So this would be a nice improvement to snappiness if we can get away with it.

### DIFF
--- a/src/seer/langfuse.py
+++ b/src/seer/langfuse.py
@@ -45,14 +45,9 @@ def append_langfuse_observation_metadata(new_metadata: dict, langfuse: Langfuse 
     MUST BE RUN WITHIN A LANGFUSE OBSERVATION!
     """
     try:
-        observation_id = langfuse_context.get_current_observation_id()
-        langfuse_context.flush()
-        if observation_id:
-            observation = langfuse.get_observation(observation_id)
-
-            langfuse_context.update_current_observation(
-                metadata=(observation.metadata or {}) | new_metadata,
-            )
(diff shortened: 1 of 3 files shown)


src/seer/langfuse.py
```diff
@@ -45,14 +45,9 @@ def append_langfuse_observation_metadata(new_metadata: dict, langfuse: Langfuse
...
-        observation_id = langfuse_context.get_current_observation_id()
-        langfuse_context.flush()
-        if observation_id:
-            observation = langfuse.get_observation(observation_id)
-
-            langfuse_context.update_current_observation(
-                metadata=(observation.metadata or {}) | new_metadata,
-            )
+        langfuse_context.update_current_observation(
+            metadata=new_metadata,
```
roaga, line 49: the source code seems to say that this metadata will be merged with existing metadata (not that we have to worry about it, we just track one boolean field)

jennmueng (approved):
lgtm and this is great as a performance improvement. originally the manual flushing was so the langfuse traces would keep getting updated in realtime as you look at running evals etc

PR #1698, opened 2025-01-07
fix(anomaly detection): Fix Integrity Error

aayush-se:
- Use `session.flush()` after the executed statement to separate delete and insert queries

Sources:
- https://m1lt0n.github.io/sqlalchemy/orm/python/delete-before-insert-sqlalchemy/
- https://github.com/sqlalchemy/sqlalchemy/issues/2501
- https://stackoverflow.com/questions/4201455/sqlalchemy-whats-the-difference-between-flush-and-commit
- https://docs.sqlalchemy.org/en/20/orm/session_basics.html#flushing

### DIFF
--- a/src/seer/anomaly_detection/accessors.py
+++ b/src/seer/anomaly_detection/accessors.py
@@ -8,6 +8,7 @@ import numpy as np
 import sentry_sdk
 import stumpy  # type: ignore # mypy throws "missing library stubs"
 from pydantic import BaseModel
+from sqlalchemy import delete
 
 from seer.anomaly_detection.models import (
     AlgoConfig,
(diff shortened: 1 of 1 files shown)


src/seer/anomaly_detection/accessors.py
```diff
@@ -239,10 +238,13 @@ def save_alert(
                         "external_alert_id": external_alert_id,
                     },
                 )
-                delete_q = delete(DbDynamicAlert).where(
-                    DbDynamicAlert.external_alert_id == external_alert_id
+                alert = (
+                    session.query(DbDynamicAlert)
+                    .filter_by(external_alert_id=external_alert_id)
+                    .one()
```
ram-senth, line 244: I'm still wondering if we need this query. Basically we are loading the entire time series into memory just to delete it right?

aayush-se: reverted back to execute and included the flush

ram-senth (approved):

PR #1721, opened 2025-01-09
chore(anomaly_detection): optimization of hydration method

ram-senth:
Removing multiple iterations on the time series and also pre-allocating lists. These changes result in about 15% reduction in time taken by the hydration method on local.

### DIFF
--- a/src/seer/anomaly_detection/accessors.py
+++ b/src/seer/anomaly_detection/accessors.py
@@ -93,67 +93,64 @@ class DbAlertDataAccessor(AlertDataAccessor):
     ) -> DynamicAlert:
         rand_offset = random.randint(0, 24)
         timestamp_threshold = (datetime.now() - timedelta(days=28, hours=rand_offset)).timestamp()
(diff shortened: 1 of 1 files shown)


src/seer/anomaly_detection/accessors.py
```diff
@@ -93,67 +93,64 @@ def _hydrate_alert(
...
-        mp_fixed = []
-        ts = []
-        values = []
-        original_flags = []
-        use_suss = []
 
         timeseries = db_alert.timeseries
+        n_points = len(timeseries)
+        ts = [0.0] * n_points
+        values = [0.0] * n_points
```
aayush-se, line 100: Could these two be defined as a numpy array directly? There is a conversion from Python array --> np.array on lines 182-183 which could result in a complete optimization for this

aayush-se (approved):
This looks great! Predefining the array size and assigning as we iterate rather than appending to the array will definitely increase speed.

PR #1738, opened 2025-01-11
fix(seer): Fix multiple bootup problem

jennmueng:
Celery's `on_configure` is called to configure the "celery app" instance, this includes when you need to intialize it to send a task. This means that when we were submitting a task, such as through `apply_async` **we were actually calling `bootup` again.**

This is fixed by moving the bootup calls for both the celery worker and celerybeat to its own init signal handlers

Fixes [SEER-9P](https://sentry.sentry.io/issues/5731122470/)

----
This issue was able to be reproduced locally by simulating _multiple instances of seer_ locally, then in the logs it was clear that when we were scheduling the task, the _bootup_ call was being called from wherever the task was being scheduled from, **not** from the forked worker process like we thought. Although I didn't get the SQLAlchemy error like in prod, it was clearly calling bootup multiple times when bootup has already been called in that instance, I'm not sure why I'm not getting that error locally.

### DIFF
--- a/src/celery_app/app.py
+++ b/src/celery_app/app.py
@@ -16,17 +16,25 @@ celery_app = Celery("seer")
 
 # This abstract helps tests that want to validate the entry point process.
 def setup_celery_entrypoint(app: Celery):
-    app.on_configure.connect(init_celery_app)
+    app.on_configure.connect(on_configure)
+    app.on_after_finalize.connect(on_after_finalize)
 
 
+# on_configure signal sent when celery app is being configured. This is *also* called when the celery app is imported and a task is scheduled from outside celery as well.
 @inject
-def init_celery_app(*args: Any, sender: Celery, config: CeleryConfig = injected, **kwargs: Any):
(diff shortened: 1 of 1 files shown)


src/celery_app/app.py
```diff
@@ -37,6 +42,24 @@ def capture_worker_name(sender, instance, **kwargs):
...
+
+
+# celerybeat_init signal sent after celerybeat app has been initialized.
+@signals.beat_init.connect
+def bootup_celery_beat(sender, **kwargs):
+    bootup(
+        start_model_loading=False,
+        integrations=[CeleryIntegration(propagate_traces=True)],
+    )
+
```
roaga, line 61: why are these signals defined with a decorator vs within `setup_celery_entrypoint`?

roaga (approved):
thank you for your sacrifice 🚀

ram-senth (approved):

src/celery_app/app.py
jennmueng: in `setup_celery_entrypoint` those are signals on the celery app instance, while these are signals only to be run on the worker and celerybeat

PR #1748, opened 2025-01-14
fix(autofix) Retry stream from scratch

### LINKED ISSUE #1671: (autofix): Investigate/implement exponential backoff for streaming vertex ai calls
Looks like we're getting overloaded errors from anthropic on vertex, but we should be using a built-in exponential backoff retry strategy from the SDK. Investigate whether this is actually being used on the streaming calls and identify whether we need to implement one ourselves for the streaming calls

https://sentry.sentry.io/issues/6152252202/?notification_uuid=17a60c1a-a237-4557-8ea6-076b1e32ecd5&project=6178942

We need to watch out for error events and retry:
https://docs.anthropic.com/en/api/messages-streaming#error-events

kddubey:
[Investigation](https://github.com/getsentry/ml-models/blob/main/autofix/retry_stream/investigate_errors.ipynb) found that backoff would address almost all of the overloaded errors.

Fix #1671 by retrying the completion from scratch, unlike #1675.

## How has this been tested?

New unit tests and local runs described below

<details>
<summary>Local autofix runs</summary>

Repro issue locally:

1. `git checkout main`
2. Replace all `AnthropicProvider`s with `AnthropicProviderFlaky`s in `coding/components.py`.
3. Run the Autofix coding step on a local issue, confirm you see the overloaded error and the run fails.
4. `git stash` the stuff in step 2 b/c we'll use it again when testing the fix.

Test fix:

1. `git checkout kddubey/autofix/retry-stream-from-scratch`
2. `git stash pop` to replace all `AnthropicProvider`s with `AnthropicProviderFlaky`s in `coding/components.py`.
3. Run Autofix on a local issue, confirm you see the overloaded error in your logs, but it quickly tries again and finishes the run.
4. `git stash` to undo step (2) to test an unflaky API
5. Run Autofix on the same issue, confirm the run finishes fine.

</summary>


src/seer/automation/autofix/autofix_agent.py
```diff
@@ -3,6 +3,8 @@
 from concurrent.futures import Executor, Future, ThreadPoolExecutor
 from typing import Optional
 
+import backoff
```
kddubey, line 6: this is already in our dep tree. fyi [last commit](https://github.com/litl/backoff) was 2 years ago

src/seer/utils.py
```diff
@@ -68,3 +73,44 @@ def closing_queue(*queues: Queue):
                 queue.get_nowait()
             except Empty:
                 pass
+
+
+def backoff_on_exception(
```
kddubey, line 78: we have the package [`backoff`](https://github.com/litl/backoff) in our [dep tree](https://github.com/getsentry/seer/blob/main/requirements.txt#L41-L44) but the author seems to have ghosted the project

jennmueng (approved):

roaga:
Would this strategy apply to OpenAI Timeout errors?

kddubey:
> Would this strategy apply to OpenAI Timeout errors?

[This error](https://sentry.sentry.io/issues/6111669349/?project=6178942&query=issue.priority%3A%5Bhigh%2C%20medium%5D%20APITimeoutError&referrer=issue-stream&sort=date&statsPeriod=10w&stream_index=0&utc=true) isn't raised during streaming/iteration so it should be retried already (twice by default). I haven't yet seen OpenAI raising errors during streaming. But when we do, we'll fill in [`OpenAIProvider.is_completion_exception_retryable`](https://github.com/getsentry/seer/pull/1748/files#diff-203419f91fa86e4fe638980e6024ac956ab5040abf9e0bbb0d6b4a22ddc7c7b2) and those requests will get retried.

PR #1749, opened 2025-01-14
fix(autofix): Truncate long keyword search context

roaga:
Sometimes the code context returned by keyword search would be really long (usually when there was a ton of code on one line, like some js core bundles), causing the LLM to error out. This PR limits each piece of context to 2000 characters.

NOTE: waiting on security issue to be resolved before I push working cassettes

### DIFF
--- a/src/seer/automation/autofix/tools.py
+++ b/src/seer/automation/autofix/tools.py
@@ -207,6 +207,8 @@ class BaseTools:
         file_names = []
         for result in results:
             for match in result.matches:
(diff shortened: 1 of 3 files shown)


jennmueng (commented):
should we implement the max context size inside the `CodeSearcher` instead? Say a `max_context_size` param to the constructor similar to `max_file_size_bytes`

jennmueng (approved):
lgtm, a test would be nice

PR #1762, opened 2025-01-16
fix(api): Reintroduce request signing

jennmueng:
_Previously request signing was disabled due to a quirk in our networking setup that made the request url passed to the client signer not the same as the request seen on the server side._

This reimplements the request signing with an optional nonce but without the salted url verification. This is a stop-gap measure to reimplement signature signing for now until we have bandwidth to invest in a more complete solution.

### DIFF
--- a/src/seer/json_api.py
+++ b/src/seer/json_api.py
@@ -91,9 +91,7 @@ def json_api(blueprint: Blueprint, url_rule: str) -> Callable[[_F], _F]:
                     body=raw_data,
                     signature=parts[1],
                 ):
-                    raise Unauthorized(
-                        f"Rpcsignature did not match for given url {request.url} and data"
-                    )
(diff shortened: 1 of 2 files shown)


src/seer/json_api.py
```diff
@@ -82,20 +82,18 @@ def decorator(
...
-                #             f"Rpcsignature did not match for given url {request.url} and data"
-                #         )
-                pass
+                parts = auth_header.split()
+                if len(parts) != 2 or not compare_signature(
+                    body=raw_data,
+                    signature=parts[1],
+                ):
+                    raise Unauthorized(
+                        f"Rpcsignature did not match for given url {request.url} and data"
```
mdtro, line 95: Since the `request.url` is no longer a part of the signature calculation, we should adjust the error message here.

mdtro (approved):
Looks good outside of my suggestion above on the error message details. 👍

PR #1770, opened 2025-01-21
chore(anomaly-detection): add timeout for calls from UI that include both batch and stream computation

ram-senth:
Some of the detection calls from the UI are taking too long. This PR adds timeout for stream detection of multiple timesteps to address.

### DIFF
--- a/src/seer/anomaly_detection/anomaly_detection.py
+++ b/src/seer/anomaly_detection/anomaly_detection.py
@@ -314,7 +314,7 @@ class AnomalyDetection(BaseModel):
         ts_external: List[TimeSeriesPoint] = ts_with_history.current
 
         historic = convert_external_ts_to_internal(ts_with_history.history)
-        # We are doing 4 detection operations, allocating 1/4th of the time budget for
+        # We are doing 4 detection operations, so allocating 1/4th of the time budget for each one.
         time_budget_ms = time_budget_ms // 4 if time_budget_ms else None
 
         # Run batch detect on history data
@@ -389,7 +389,9 @@ class AnomalyDetection(BaseModel):
             )
 
     @sentry_sdk.trace
-    def detect_anomalies(self, request: DetectAnomaliesRequest) -> DetectAnomaliesResponse:
+    def detect_anomalies(
+        self, request: DetectAnomaliesRequest, time_budget_ms: int = 4500
+    ) -> DetectAnomaliesResponse:
         """
         Main entry point for anomaly detection.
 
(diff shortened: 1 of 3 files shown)


tests/seer/anomaly_detection/test_anomaly_detection.py
```diff
@@ -457,13 +457,44 @@ def test_detect_anomalies_combo_large_current(self):
...
+
+        config = AnomalyDetectionConfig(
+            time_period=60, sensitivity="low", direction="both", expected_seasonality="auto"
+        )
+
+        loaded_synthetic_data = convert_synthetic_ts(
+            "tests/seer/anomaly_detection/test_data/synthetic_series", as_ts_datatype=True
+        )
+        ts_history = loaded_synthetic_data.timeseries[0][:180]
+        print(len(ts_history))
```
aayush-se, line 477: nit: leftover print

src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -314,17 +314,16 @@ def _combo_detect(
         ts_external: List[TimeSeriesPoint] = ts_with_history.current
 
         historic = convert_external_ts_to_internal(ts_with_history.history)
+        # We are doing 4 detection operations, so allocating 1/4th of the time budget for each one.
+        time_budget_ms = time_budget_ms // 4 if time_budget_ms else None
```
aayush-se, line 318: Batch detection over the history should take longer than stream detection on a single point. Shouldn't this be reflected in the `time_budget_ms`? ie: Batch calls can each take 1/3 the time and the stream calls 1/6th (2 * 1/3 + 2 * 1/6 in total)?

src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -390,7 +389,9 @@ def _update_anomalies(
             )
 
     @sentry_sdk.trace
-    def detect_anomalies(self, request: DetectAnomaliesRequest) -> DetectAnomaliesResponse:
+    def detect_anomalies(
+        self, request: DetectAnomaliesRequest, time_budget_ms: int = 4500
```
aayush-se, line 393: qq: the default is set to 4500ms since the upper limit is 5 seconds and we want to allow for ~500ms buffer?

aayush-se (approved):
Just had 1 question on the allocation of time for batch/stream in combo and 1 clarifying question. Aside from that, the core logic looks good!

src/seer/anomaly_detection/anomaly_detection.py
ram-senth: it actually depends on a few factors - how much history data vs current data the timeseries has, how many anomalies are found in each section etc (prophet calls are the costliest and happens only when anomaly is detected). Hence I kept it simple.

src/seer/anomaly_detection/anomaly_detection.py
ram-senth: Yes that is correct, alerting system times out after 5 mins.

aayush-se: Wait 5 minutes or seconds?

ram-senth: 5 mins.

PR #1771, opened 2025-01-22
fix(autofix): Fallback to 4o-mini on root cause output token limit

jennmueng:
Fallback to gpt-4o-mini when we hit length limits on the root cause structured output call, as it has a 16k output window vs 4k for gpt-4o

Slightly lower quality output with 4o mini is preferrable to the erroring that the user might see.

Fixes [SEER-G4](https://sentry.sentry.io/issues/6137031127/)

### DIFF
--- a/src/seer/automation/autofix/components/root_cause/component.py
+++ b/src/seer/automation/autofix/components/root_cause/component.py
@@ -1,5 +1,7 @@
 import logging
 
+import openai
+import sentry_sdk
 from langfuse.decorators import observe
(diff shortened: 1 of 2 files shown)


roaga (commented):
The Sentry issue seems to occur only during the insight card generation call, not the root cause formatter, no? Unless I'm missing something, this change won't stop that error. Also, the insight card output will change and likely resolve this issue on its own with my upcoming (blocked) PR.

jennmueng:
oh yeah you're right, i misread the stacktrace, closing.

PR #1783, opened 2025-01-22
feat(autofix): Better insight card quality and working comment thread

roaga:
Changes the prompting for the insight card output.
Adds support for a comment thread within a step in Autofix.


src/seer/automation/autofix/components/comment_thread.py
```diff
@@ -0,0 +1,72 @@
...
+        self, request: CommentThreadRequest, llm_client: LlmClient = injected
+    ) -> CommentThreadOutput | None:
+        output = llm_client.generate_structured(
+            prompt=CommentThreadPrompts.format_default_msg(
+                selected_text=request.selected_text,
+                thread_memory=request.thread_memory,
+            ),
+            messages=request.run_memory,
+            system_prompt=CommentThreadPrompts.format_system_msg(),
+            model=GeminiProvider.model("gemini-1.5-flash"),
```
jennmueng, line 62: is gemini flash enabled in `internal-sentry`?

src/seer/automation/autofix/components/insight_sharing/models.py
```diff
@@ -1,53 +1,7 @@
...
-    step_type: str | None = None
+from seer.automation.component import BaseComponentOutput
 
 
 class InsightSharingOutput(BaseComponentOutput):
     insight: str
-    codebase_context: list[CodeSnippetContext] = Field(default_factory=list)
-    stacktrace_context: list[StacktraceContext] = Field(default_factory=list)
-    breadcrumb_context: list[BreadcrumbContext] = Field(default_factory=list)
     justification: str = ""
```
jennmueng, line 6: will the frontend error if these lists don't exist? you can deploy the new frontend first but as a note because sentry is an SPA users that don't do a full refresh might still have the old UI

jennmueng (approved):

src/seer/automation/autofix/components/insight_sharing/models.py
roaga: the frontend will not error, I've tried it. It just shows the text alone in the insight card without the other data.

src/seer/automation/autofix/components/comment_thread.py
roaga: If I'm not mistaken, I don't think we have to enable gemini models in either GCP project? We're already using 2.0 exp in prod, right?

PR #1795, opened 2025-01-24
fix(autofix) Fix Anthropic overloaded error detection

kddubey:
Looks like my guess for the shape of the overloaded error in #1748 was wrong. [This error's stacktrace](https://sentry.sentry.io/issues/6147058185) shows that `is_exception_retryable(exception)` returns false, and the request doesn't get retried. Changing it to this more liberal check that should still be safe.


roaga (approved):
I think this is fine, but have you checked if these error behaviors have changed at all in newer versions of the Anthropic Python SDK? Probably not, but would be great if they made our lives easier somehow

kddubey:
Stacktrace shows that the error behavior is the same as before

PR #1804, opened 2025-01-27
ref(autofix): remove repo_id and namespace_id

jennmueng:
Remove unused `repo_id` and `namespace_id` from the codebase indexing days, we now use `repo_external_id`

### DIFF
--- a/src/seer/automation/autofix/autofix_context.py
+++ b/src/seer/automation/autofix/autofix_context.py
@@ -212,15 +212,12 @@ class AutofixContext(PipelineContext):
     def commit_changes(
         self,
         repo_external_id: str | None = None,
-        repo_id: int | None = None,
         make_pr: bool = False,
         pr_to_comment_on_url: str | None = None,
     ):
         state = self.state.get()
         for codebase_state in state.codebases.values():
-            if (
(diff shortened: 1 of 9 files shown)


src/seer/automation/autofix/autofix_context.py
```diff
@@ -212,15 +212,12 @@ def process_event_paths(self, event: EventDetails):
...
-        repo_id: int | None = None,
         make_pr: bool = False,
         pr_to_comment_on_url: str | None = None,
     ):
         state = self.state.get()
         for codebase_state in state.codebases.values():
-            if (
-                repo_external_id is None and repo_id is None
-            ) or codebase_state.repo_external_id == repo_external_id:
+            if (repo_external_id is None) or codebase_state.repo_external_id == repo_external_id:
```
roaga, line 220:
```suggestion
            if repo_external_id is None or codebase_state.repo_external_id == repo_external_id:
```

roaga (approved):
does the frontend/sentry backend need any cleanup too?

jennmueng:
> does the frontend/sentry backend need any cleanup too?

yes but this should allow compatibility with the existing update payload

PR #1810, opened 2025-01-28
fix(autofix) Retry more errors during streaming

kddubey:
Some leftover work from #1748 

1. Add Anthropic and [OpenAI internal server errors](https://platform.openai.com/docs/guides/error-codes)
2. Add retrying for non-streamed and streamed Gemini completions, which they [explicitly say to do](https://cloud.google.com/vertex-ai/generative-ai/docs/error-code-429). I found this error in [this cassette](https://github.com/getsentry/seer/blob/3a391fdeb7f3614eec0ba477898a516a43de5e5e/tests/automation/autofix/_encrypted_cassettes/test_autofix_run_full.yaml.encrypted)


src/seer/automation/agent/client.py
```diff
@@ -830,8 +841,10 @@ def generate_text_stream(
 
                 # Update token counts if available
                 if chunk.usage_metadata:
-                    total_prompt_tokens = chunk.usage_metadata.prompt_token_count
-                    total_completion_tokens = chunk.usage_metadata.candidates_token_count
+                    if chunk.usage_metadata.prompt_token_count:
```
kddubey, line 844: even if `chunk.usage_metadata is not None`, it turns out that `chunk.usage_metadata.prompt_token_count` can be `None`. this change makes sure it's always an int so that the addition in the `finally` block below works

tests/automation/autofix/test_retry_stream.py
```diff
@@ -122,18 +129,45 @@ def get_client(self):
...
+    retryable_exception=ResourceExhausted(
+        message="Resource exhausted. Please try again later. Please refer to https://cloud.google.com/vertex-ai/generative-ai/docs/error-code-429 for more details."
+    ),
+    get_obj_with_create_stream_method_from_client=lambda client: client.models,
+    create_stream_method_name="generate_content_stream",
+)
 
 
-# Give this function a name instead of making it a lambda so that the cassette is named
+# Give these functions names instead of making them lambdas so that the cassettes are named
```
roaga, line 152:
```suggestion
```

roaga (approved):

PR #1849, opened 2025-02-02
feat(autofix): New root cause model

roaga:
Changes the root cause to be formatted as a reproduction timeline instead of a title/description/relevant code. Evals remained stable.

For backwards compatibility, I left the description as an optional field so we can read old runs from the db (yes, this drops the title and code snippets, but I think this is fine for temporary backwards compatibility and keeping the rest of the codebase clean).

Requires the frontend PR to go in first.


src/seer/automation/autofix/components/root_cause/models.py
```diff
@@ -1,149 +1,76 @@
...
+                    title=event.title,
+                    code_snippet_and_analysis=event.code_snippet_and_analysis,
+                    timeline_item_type=event.timeline_item_type,
+                    relevant_code_file=event.relevant_code_file,
+                    is_most_important_event=event.is_most_important_event,
                 )
-                if model.code_context
-                else None
-            ),
+                for event in (model.root_cause_reproduction or [])
```
jennmueng, line 56: Do we need to be mapping the list of timeline events back and forth between the models? Aren't they both named `root_cause_reproduction` in both cases and share the same `list[TimelineEvent] | None` type?

roaga: ah good point lol

PR #1870, opened 2025-02-04
feat(autofix): Refactor plan+code into a Solution step and a Code Changes step

roaga:
Adds a Solution step that auto-runs after the Root Cause step and outputs a similar timeline. This is internally the same as the Plan part of the old Plan+Code. This also makes the Plan+Code step just Code.

The frontend PR should go in first.

### DIFF
--- a/src/seer/automation/autofix/components/coding/prompts.py
+++ b/src/seer/automation/autofix/components/coding/prompts.py
@@ -53,6 +53,10 @@ class CodingPrompts:
             return f"The user has provided the following instruction for the fix: {root_cause}"
 
     @staticmethod
+    def format_custom_solution(custom_solution: str):
+        return f"The user has provided the following solution idea: {custom_solution}"
+
+    @staticmethod
     def format_fix_msg(has_tools: bool = True, custom_solution: str | None = None):
         return textwrap.dedent(
             """\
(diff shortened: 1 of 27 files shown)


src/seer/automation/autofix/components/coding/prompts.py
```diff
@@ -355,6 +161,11 @@ def format_is_obvious_msg(
                     if root_cause_extra_instruction
                     else ""
                 ),
+                custom_solution_str=(
+                    f"The user has provided the following solution idea: {custom_solution}"
+                    if custom_solution
+                    else ""
+                ),
```
jennmueng, line 168: should we pull this out into a `format_custom_solution`

src/seer/automation/autofix/tasks.py
```diff
@@ -172,14 +177,71 @@ def run_autofix_execution(
         cur.mark_triggered()
 
     event_manager = AutofixEventManager(state)
-    event_manager.send_coding_start()
+
+    # delete all steps after root cause step
+    root_cause_step_index = next(
```
jennmueng, line 182: should we rename this function to `run_autofix_solution`?

src/seer/automation/autofix/components/coding/prompts.py
roaga: done

src/seer/automation/autofix/tasks.py
roaga: done

jennmueng (approved):

PR #1876, opened 2025-02-05
chore(autofix): disable thread killing in CI 

roaga:
this should _hopefully_ fix the seg fault issue. I unskipped most of the skipped tests, except one which needs to be updated for the new flow and isn't straightforward to do so. Can't verify if this will work though 🤷


src/seer/automation/autofix/steps/steps.py
```diff
@@ -128,7 +128,8 @@ def _check_for_kill(self):
                 with self.context.state.update() as cur:
                     cur.signals.remove(kill_signal)
                 self.thread_kill = True
-                os._exit(1)  # kills the thread and the whole step
+                if not os.environ.get("CI"):  # don't kill threads in CI
+                    os._exit(1)  # kills the thread and the whole step
```
jennmueng, line 132:
```suggestion
                if not os.environ.get("CI"):  # don't kill threads in CI
                    os._exit(1)  # kills the thread and the whole step
                else:
                    return
```

PR #1880, opened 2025-02-06
feat(autofix): Add unit test options to coding step

roaga:
The coding step can be invoked with options to just write a fix, to just write a repro test, or both. It defaults to just a fix.
Evals are stable between writing a fix and writing a fix with a test.

This PR should go in before the frontend change.

### DIFF
--- a/src/seer/automation/autofix/components/coding/prompts.py
+++ b/src/seer/automation/autofix/components/coding/prompts.py
@@ -1,5 +1,5 @@
 import textwrap
-from typing import Optional
+from typing import Literal, Optional
 
 from seer.automation.autofix.components.coding.models import (
     FuzzyDiffChunk,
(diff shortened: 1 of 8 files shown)


src/seer/automation/autofix/components/coding/prompts.py
```diff
@@ -143,9 +157,18 @@ def format_is_obvious_msg(
...
+                Does the code change needed for {mode_str} exist ONLY in files you can already see in your context here or do you need to look at other files?"""
             )
             .format(
+                mode_str=(
+                    "fixing the issue"
+                    if mode == "fix"
+                    else (
+                        "writing a unit test to reproduce the issue"
+                        if mode == "test"
+                        else "writing a unit test and fixing the issue"
```
jennmueng, line 169: singular unit test or multiple?

roaga: Intended to be singular (one test for one issue), but it writes multiple typically

jennmueng (approved):

PR #1882, opened 2025-02-06
feat(relevant warnings) Predict relevant warnings

kddubey:
Needs https://github.com/getsentry/sentry/pull/85133 to go in first

Background: [tech spec](https://www.notion.so/sentry/Relevant-warnings-in-Seer-1888b10e4b5d8036884dc0327632361f?pvs=4)

Currently, the tests consist of mocks and patches. In the future, can store real data and VCR tests like autofix

[demo](https://github.com/getsentry/data-analysis/blob/main/error_prevention/demo.ipynb)


src/seer/automation/codegen/relevant_warnings_step.py
```diff
@@ -0,0 +1,162 @@
...
+            request
+        )
+
+        # 5. Save results.
+        try:
+            # self._post_results_to_overwatch(
+            #     run_id=self.context.run_id,
+            #     relevant_warnings_output=relevant_warnings_output,
+            # )
+            pass
```
kddubey, line 155: temp hack. will figure out how to not post during tests

src/seer/automation/codegen/relevant_warnings_step.py
```diff
@@ -0,0 +1,162 @@
...
+
+    @staticmethod
+    def _post_results_to_overwatch(
+        run_id: str, relevant_warnings_output: CodePredictRelevantWarningsOutput
+    ):
+        request = {
+            "run_id": run_id,
+            "results": relevant_warnings_output.model_dump()["relevant_warning_results"],
+        }
+        headers = get_codecov_auth_header(request)
```
kddubey, line 77: will update this in a sec. it's slightly wrong

src/seer/app.py
```diff
@@ -259,6 +270,39 @@ def codegen_unit_tests_state_endpoint(
...
+
+
+@json_api(blueprint, "/v1/automation/codegen/relevant-warnings/results")
+def codegen_relevant_warnings_results_endpoint(
+    data: CodegenRelevantWarningsStateRequest,
+) -> CodePredictRelevantWarningsOutput:
+    state = get_relevant_warnings_state(data)
+    return CodePredictRelevantWarningsOutput(
+        relevant_warning_results=state.relevant_warning_results
+    )
```
kddubey, line 303: no longer needed. will prolly delete

src/seer/app.py
```diff
@@ -286,6 +330,13 @@ def codecov_request_endpoint(
         return codegen_pr_review_endpoint(data.data)
     elif data.request_type == "unit-tests":
         return codegen_unit_tests_endpoint(data.data)
+    elif data.request_type == "relevant-warnings":
+        if not isinstance(data.data, CodegenRelevantWarningsRequest):
+            raise TypeError(
+                f"Invalid request data type: {type(data.data)}. "
+                "Expected CodegenRelevantWarningsRequest."
+            )
+        return codegen_relevant_warnings_endpoint(data.data)
```
kddubey, line 339: not needed

src/seer/automation/codegen/prompts.py
```diff
@@ -212,3 +214,96 @@ def format_pr_review_plan_step(diff_str: str):
...
+
+    class DoesFixingWarningFixIssue(BaseModel):
+        reasoning: str
+        relevance_probability: float
+        does_fixing_warning_fix_issue: bool
+        short_description: str | None = None
+        short_justification: str | None = None
+
+    @staticmethod
+    def format_prompt(formatted_warning: str, formatted_error: str):
```
kddubey, line 284: should add a code comment but the focus around "does fixing warning fix issue" is [taken from BPR](https://github.com/codecov/bug-prediction-research/blob/f79fc1e7c86f7523698993a92ee6557df8f9bbd1/src/scripts/ask_oracle.py#L137)

src/seer/automation/codegen/prompts.py
```diff
@@ -212,3 +214,96 @@ def format_pr_review_plan_step(diff_str: str):
...
+
+            We need to know if this warning is directly relevant to the issue.
+            By relevant, we mean that fixing the warning would prevent the issue.
+            We're not deciding whether or not the warning is alarming by itself; we want to know how relevant the warning is to the issue at hand.
+            We're not looking for warnings which vaguely relate to the issue. The relationship should be somewhat direct.
+
+            Before giving your final answer, think about the context in which the error occurs.
+            What are its possible causes? How would it be fixed? Next think about how the warning is related or unrelated to the issue.
+            Your `reasoning` should be at most 200 words.
+            Also, give a score between 0 and 1 for how likely it is that addressing this warning would prevent the issue. This score, the `relevance_probability`, can be very granular, e.g., 0.32.
```
kddubey, line 301: https://arxiv.org/abs/2305.14975

tests/automation/test_automation_utils.py
```diff
@@ -86,3 +89,101 @@ def test_remove_cdata(self):
...
+            id="small_chars_per_token_creates_more_batches",
+        ),
+        _BatchTextsByTokensTestCase(
+            texts=["Hello", "world", "test"],
+            max_tokens=3,
+            avg_num_chars_per_token=10.0,
+            batches_expected=[["Hello", "world", "test"]],
+            id="large_chars_per_token_allows_more_texts_per_batch",
+        ),
+        _BatchTextsByTokensTestCase(
```
kddubey, line 149: some of these are llm-generated lol. will double check

src/seer/automation/codegen/relevant_warnings_component.py
```diff
@@ -0,0 +1,273 @@
...
+            (issue, filename)
+            for filename, issues in request.filename_to_issues.items()
+            for issue in issues
+        ]
+        issues_formatted = [
+            self._format_issue_with_related_filename(issue, pr_filename)
+            for issue, pr_filename in issues_with_pr_filename
+        ]
+
+        model = GoogleProviderEmbeddings.model(
```
suejung-sentry, line 149: wow this is fancy and cool!

src/seer/app.py
```diff
@@ -259,6 +270,39 @@ def codegen_unit_tests_state_endpoint(
     )
 
 
+@json_api(blueprint, "/v1/automation/codegen/relevant-warnings")
+def codegen_relevant_warnings_endpoint(
+    data: CodegenRelevantWarningsRequest,
+) -> CodegenRelevantWarningsResponse:
+    return codegen_relevant_warnings(data)
```
suejung-sentry, line 277: Great, the inputs and outputs align with overwatch, lgtm!

Oh I'll update the url in overwatch from `/relevant_warnings` to `relevant-warnings` to match here

src/seer/app.py
```diff
@@ -259,6 +270,39 @@ def codegen_unit_tests_state_endpoint(
...
+
+
+@json_api(blueprint, "/v1/automation/codegen/relevant-warnings/state")
```
suejung-sentry, line 280: Overwatch is technically not using this one right now either, but since it's hooked up already can just keep it around, either way

src/seer/automation/codegen/relevant_warnings_step.py
```diff
@@ -0,0 +1,159 @@
...
+        requests.post(
+            url="https://overwatch.codecov.dev/api/ai/seer/relevant-warnings",
```
suejung-sentry, line 79: At the very end, we can point this at prod overwatch (not deployed yet), which sounds like it'll end up at `overwatch.codecov.io`

suejung-sentry (commented):
This is a work of art! Amazing stuff!

src/seer/automation/codebase/models.py
```diff
@@ -58,3 +60,89 @@ class GithubPrReviewComment(TypedDict):
...
+            base += f"~{self.end_line}"
+        return base
+
+
+# Mostly copied from https://github.com/codecov/bug-prediction-research/blob/main/src/core/database/models.py
+class StaticAnalysisRule(BaseModel):
+    id: int
+    code: str
+    tool: str
+    is_autofixable: bool | None  # refers to "Quick fix"?
```
giovanni-guidini, line 107: yes, it does

src/seer/automation/codebase/models.py
```diff
@@ -58,3 +60,89 @@ class GithubPrReviewComment(TypedDict):
     start_line: NotRequired[int]
     start_side: NotRequired[Literal["LEFT", "RIGHT"]]
     in_reply_to: NotRequired[str]
+
+
+# Copied from https://github.com/codecov/bug-prediction-research/blob/main/src/core/typings.py
+class Location(BaseModel):
```
giovanni-guidini, line 66: Depending on the tool we can have the full code region (start_line, start_column, end_line, end_column)

Would we be interested in that?

src/integrations/codecov/codecov_auth.py
```diff
@@ -10,24 +11,29 @@
 OUTGOING_REQUEST_SIGNATURE_HEADER = "HTTP-X-GEN-AI-AUTH-SIGNATURE"
 
 
+@inject
+def get_codecov_auth_header(
+    request: dict[str, Any], config: AppConfig = injected
+) -> dict[str, str]:
+    request_data = json.dumps(request).encode("utf-8")
+    key = config.CODECOV_OUTGOING_SIGNATURE_SECRET
```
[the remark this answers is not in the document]
kddubey: nice, good to know. decided to just pass in `X-GEN-AI-AUTH-SIGNATURE`

src/seer/automation/codegen/models.py
```diff
@@ -15,9 +16,21 @@ class CodegenStatus(str, Enum):
...
 
 
+class RelevantWarningResult(BaseModel):
+    warning_id: int
+    issue_group_id: int
+    does_fixing_warning_fix_issue: bool
+    relevance_probability: float
+    reasoning: str
+    short_description: str
+    short_justification: str
```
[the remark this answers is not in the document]
kddubey: `short_justification` is inspired by that first figma mockup that had a short justification as a nested bullet. it's not in the latest mockup, but decided to generate it anyway b/c it's cheap and we can see if it's useful

src/seer/app.py
kddubey: true. i'll leave it in for now. can delete later when we're certain we won't use it

src/seer/automation/codebase/models.py
kddubey: hmm good question. i don't think `start_column` or `end_column` are particularly useful rn

i realized it would prolly simplify things if the `encoded_location` string (which matches the regex in `Location`) was instead a:

```python
from pydantic import BaseModel

class Location(BaseModel):
    filename: str
    start_line: str
    end_line: str
```

so a warning in the overwatch request has `location={"filename": ..., "start_line": ..., "end_line": ...}` instead of `encoded_location`. can also include `start_column, end_column` keys since pydantic will just ignore them if they're not in the model

low priority tho, so up to you if you wanna make this change in overwatch

src/seer/automation/agent/client.py
```diff
@@ -1395,6 +1403,90 @@ def construct_message_from_stream(
             raise ValueError(f"Invalid provider: {model.provider_name}")
 
 
+@dataclass
+class GoogleProviderEmbeddings:
```
jennmueng, line 1407: nit: should this be in a separate file like `embeddings.py`?

src/seer/automation/codegen/prompts.py
```diff
@@ -212,3 +214,104 @@ def format_pr_review_plan_step(diff_str: str):
...
+        Put it right after the system message so that the system message and error are cached
+        together.
+        """
+        return textwrap.dedent(
+            f"""\
+            Here is an issue we had in our codebase:
+
+            {formatted_error}
+            """
+        )
```
jennmueng, line 248:
```suggestion
        return textwrap.dedent(
            """\
            Here is an issue we had in our codebase:

            {formatted_error}
            """
        ).format(formatted_error=formatted_error)
```
dedent won't correctly work if the formatted error has indents in it too

src/seer/automation/codegen/prompts.py
```diff
@@ -212,3 +214,104 @@ def format_pr_review_plan_step(diff_str: str):
...
+            f"""\
+            {_RelevantWarningsPromptPrefix.format_prompt_error(formatted_error)}
+
+            Carefully analyze the issue above. Focus on the error, the stacktrace.
+            Think about the context in which the error occurs. What are possible causes to it? How do you fix it?
+            Answer "does this issue originate from within the application or is it caused from an external service not behaving well?"
+            For example a TypeError or ValueError is likely to be caused by the application, while a 500 error is likely to be caused by an external service.
+            You should have a somewhat high bar for answering that the issue is unfixable.
+            """
+        )
```
jennmueng, line 271:
```suggestion
        return textwrap.dedent(
            """\
            {error_prompt}

            Carefully analyze the issue above. Focus on the error, the stacktrace.
            Think about the context in which the error occurs. What are possible causes to it? How do you fix it?
            Answer "does this issue originate from within the application or is it caused from an external service not behaving well?"
            For example a TypeError or ValueError is likely to be caused by the application, while a 500 error is likely to be caused by an external service.
            You should have a somewhat high bar for answering that the issue is unfixable.
            """
        ).format(prompt_error=_RelevantWarningsPromptPrefix.format_prompt_error(formatted_error))
```
same here

src/seer/automation/codegen/prompts.py
```diff
@@ -212,3 +214,104 @@ def format_pr_review_plan_step(diff_str: str):
...
+            Next, give a score between 0 and 1 for how likely it is that addressing this warning would prevent the issue based on your `reasoning`.
+            This score, the `relevance_probability`, can be very granular, e.g., 0.32.
+            Then give your final answer in `does_fixing_warning_fix_issue`.
+
+            Finally, if you believe the warning is relevant to the issue (`does_fixing_warning_fix_issue=true`), then fill in two more sections:
+              - `short_description`: a short and sweet description of the problem caused by not addressing the warning.
+                This description must focus on the problem and not the warning itself. It should be at most 10 words.
+              - `short_justification`: a short summary of your reasoning for why the warning is relevant to the issue. This justification should be at most 10 words.
+            """
+        )
```
jennmueng, line 317:
```suggestion
        return textwrap.dedent(
            """\
            {error_prompt}

            Here is a warning that just surfaced in our codebase:

            {formatted_warning}

            We need to know if this warning is directly relevant to the issue.
            By relevant, we mean that fixing the warning would prevent the issue.
            We're not deciding whether or not the warning is alarming by itself; we want to know how relevant the warning is to the issue at hand.
            We're not looking for warnings which vaguely or hypothetically relate to the issue. The relationship should be direct.
            The locations of the warning and the error don't have to be identical, but they should be referring to similar code/logic.

            Before giving your final answer, think out loud in a `reasoning` section about the context in which the error occurs, independent of the warning. What are the possible causes of the error?
            Then think about how the warning is related or unrelated to the issue. Is the warning referring to similar code/logic as the error's exception stacktrace?
            Your `reasoning` should be at most 400 words. Feel free to mention what you're uncertain about, and what you're more confident about.

            Next, give a score between 0 and 1 for how likely it is that addressing this warning would prevent the issue based on your `reasoning`.
            This score, the `relevance_probability`, can be very granular, e.g., 0.32.
            Then give your final answer in `does_fixing_warning_fix_issue`.

            Finally, if you believe the warning is relevant to the issue (`does_fixing_warning_fix_issue=true`), then fill in two more sections:
              - `short_description`: a short and sweet description of the problem caused by not addressing the warning.
                This description must focus on the problem and not the warning itself. It should be at most 10 words.
              - `short_justification`: a short summary of your reasoning for why the warning is relevant to the issue. This justification should be at most 10 words.
            """
        ).format(
          error_prompt=_RelevantWarningsPromptPrefix.format_prompt_error(formatted_error),
          formatted_warning=formatted_warning
        )
```
and here

src/seer/automation/codegen/relevant_warnings_component.py
```diff
@@ -0,0 +1,281 @@
...
+        ]
+        return AssociateWarningsWithIssuesOutput(candidate_associations=candidate_associations)
+
+
+def _is_issue_fixable_cache_key(llm_client: LlmClient, issue: IssueDetails) -> tuple[str]:
+    return hashkey(issue.id)
+
+
+@cached(cache=LRUCache(maxsize=2048), key=_is_issue_fixable_cache_key)
+def _is_issue_fixable(llm_client: LlmClient, issue: IssueDetails) -> bool:
```
jennmueng, line 181: nit: can just inject llm_client here and not need to pass it

jennmueng (approved):
only some nits nothing major, otherwise lgtm on a high level

PR #1883, opened 2025-02-06
chore(autofix): Replace 4o mini with Flash 2

roaga:
Replaces all uses of 4o mini in Autofix with Flash 2.0. This means short circuiters and formatters. Cut 45 seconds off run time, improved root cause eval from 80 to 84%, but dropped fix eval from 39% to 35%.

### DIFF
--- a/src/seer/automation/autofix/components/root_cause/models.py
+++ b/src/seer/automation/autofix/components/root_cause/models.py
@@ -19,7 +19,7 @@ class TimelineEvent(BaseModel):
     timeline_item_type: (
         Literal["internal_code", "external_system", "human_action"] | str
     )  # TODO put back to literal only when not breaking anything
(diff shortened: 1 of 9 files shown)


src/seer/automation/autofix/components/root_cause/models.py
```diff
@@ -19,7 +19,7 @@ class TimelineEvent(BaseModel):
     timeline_item_type: (
         Literal["internal_code", "external_system", "human_action"] | str
     )  # TODO put back to literal only when not breaking anything
-    relevant_code_file: RelevantCodeFile | None
+    relevant_code_file: RelevantCodeFile
```
jennmueng, line 22: if this breaks we revert

jennmueng (approved):

PR #1892, opened 2025-02-07
fix(ai): Fix gemini tool usage

jennmueng:
We were not handling Gemini tools correctly
1. In the `tool` type, we were passing in the tool and its arguments again not the tool responses
2. Gemini expects all tools to be sent back in a single `Content` not as multiple tool messages like we have


roaga (approved):
ty

roaga:
oh the mypy issue is back, probably can just ignore it

PR #1895, opened 2025-02-09
feat(autofix): Use an LLM to apply edits to full files in the coding step

roaga:
Inspired by Cursor's apply model:
- remove strict diff requirements from the coding prompts
- apply unstructured code suggestions to full files with an LLM (gpt-4o-mini) using speculative decoding (300-500 tokens per second from my observations)
- automatically create the diff using unifieddifflib
- process code changes in parallel

Pros:
- improved fix evals by 4 points up to 41% (after all these changes, we're up to 88% correct RC and 41% correct fix on the new evals)
- removes unreliability of incorrect diffs, missing changes, and poorly applied changes (which should solve some of the issues that have been plaguing us forever, including recent reports)
- usually insignificant difference in overall latency. It takes a few seconds to apply, but reduces tokens the main agent needs to write and removes incorrect diff fixer.

Cons:
- adds more latency if editing a very big file (in the future, we could consider using a smaller model on Fireworks with predicted output for even more speed; that's what Cursor does)


src/seer/automation/autofix/components/coding/utils.py
```diff
@@ -47,8 +47,19 @@ def extract_diff_chunks(diff_text: str) -> list[FuzzyDiffChunk]:
                 )
                 current_original = []
                 current_new = []
+                current_diff_contents = []  # reset the diff contents for the new chunk
             current_header = line
-        elif line.startswith("-"):
+            current_diff_contents.append(line)
+            continue
+
+        # Skip empty lines that are immediately before or after a header as they are not part of the chunk
```
roaga, line 55: this fixes the bug we've had for a while where it adds random empty lines in the diff

PR #1904, opened 2025-02-11
fix(autofix): Some fixes

roaga:
Not going ahead with DeepClaude, but discovered some bug in converting tool calls between OpenAI, Anthropic, and Gemini, so fixing that. Also adding support for reasoning-related parameters for OpenAI.

Also improving the get_valid_file_paths check that the AI suggested.

And fixes the exception logging in get_file_content so we don't get spammed by Sentry.

And change the index at which we insert the initial prompt in the solution/coding step to prevent some hallucinations we were seeing.

### DIFF
--- a/src/seer/automation/agent/client.py
+++ b/src/seer/automation/agent/client.py
@@ -109,12 +109,14 @@ class OpenAiProvider:
         max_tokens: int | None = None,
         timeout: float | None = None,
         predicted_output: str | None = None,
+        reasoning_effort: str | None = None,
     ):
         message_dicts, tool_dicts = self._prep_message_and_tools(
             messages=messages,
             prompt=prompt,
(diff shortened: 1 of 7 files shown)


src/seer/automation/agent/client.py
```diff
@@ -985,7 +994,7 @@ def _prep_message_and_tools(
 
     @staticmethod
     def to_content(message: Message) -> Content:
-        if message.role == "tool_use":
+        if message.role == "tool_use" or (message.role == "assistant" and message.tool_calls):
```
jennmueng, line 997: Consider adding tests for cross-provider tool call conversion in tests/automation/agent/test_agent.py to verify consistent message formatting.

jennmueng (approved):

PR #1925, opened 2025-02-12
build(ci): Parallelize tests and optimize image building

jennmueng:
1. Parallelizes the tests using `pytest-split`
2. Uses a single build and push command for the docker image, then pulls it for every other job instead of calling Buildx every time
3. Points GoCD to look at the new parallelized tests check

**_Reduces total CI runtime from ~20 minutes to ~5 minutes._** This will also have the same effect on _deploy_ times!

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -113,7 +113,7 @@ jobs:
           fi
 
       - name: Restore test durations cache
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: .pytest_cache
           key: ${{ runner.os }}-pytest-split-cache-${{ hashFiles('requirements*.txt') }}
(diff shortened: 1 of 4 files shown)


roaga (approved):
rubber stamping but i may not be the best person to review this

.github/workflows/tests.yml
```diff
@@ -52,49 +112,53 @@ jobs:
...
-          pip install -r scripts/requirements.txt
-          make vcr-decrypt
+
       - name: Test with pytest
         run: |
-          make test
+          docker compose up -d test-db
+          docker compose run ${{ env.EXTRA_COMPOSE_TEST_OPTIONS }} app bash -c "\
+            pip install pytest-split && \
+            pytest --splits 8 --group ${{ matrix.group }} --cov . --cov-report=\"xml:.artifacts/coverage.xml\" -vv"
```
roaga, line 157: would pytest split help for running tests locally?

jennmueng: matrix would be annoying to do locally

jennmueng: but doable

PR #1929, opened 2025-02-12
chore(autofix): Repo-agnostic search tools

roaga:
Convert keyword search, file search, file search wildcard, and semantic file search so that they do not take a repo name as a parameter. They search across all repos provided and return results labeled by repo. Also enforces repo name as a required parameter for expand_document.


src/seer/automation/autofix/tools.py
```diff
@@ -196,113 +212,127 @@ def _separate_dirs_and_files(
...
+        repo_names = self._get_repo_names()
+        found_files = ""
 
-        self.context.event_manager.add_log(f"Searching for file `{filename}` in `{repo_name}`...")
+        for repo_name in repo_names:
+            repo_client = self.context.get_repo_client(
+                repo_name=repo_name, type=self.repo_client_type
+            )
+            all_paths = repo_client.get_index_file_set()
+            found = [path for path in all_paths if os.path.basename(path) == filename]
```
jennmueng, line 296:
```suggestion
            found = [path for path in all_paths if os.path.basename(path).lower() == filename.lower()]
```
Case insensitive comparison for filenames

tests/automation/autofix/test_autofix_tools.py
```diff
@@ -176,30 +234,36 @@ def test_semantic_file_search_not_found_no_contents(self, autofix_tools: BaseToo
...
+        expected = "Could not figure out which file matches what you were looking for. You'll have to try yourself."
+        assert result == expected
 
     def test_semantic_file_search_with_repo_name(self, autofix_tools: BaseTools):
+        # Instead of passing repo_name directly as an argument,
+        # set context.repos so that _get_repo_names() returns "owner/specific_repo"
+        dummy_repo = MagicMock(owner="owner", name="specific_repo")
+        dummy_repo.owner = "owner"
+        dummy_repo.name = "specific_repo"
+        autofix_tools.context.repos = [dummy_repo]
```
jennmueng, line 253: Existing unit tests only validate single repository scenarios. There's no test case verifying proper:

- Multi-repo search aggregation
- Cross-repo path disambiguation
- Add a test with 2+ repos containing overlapping filenames.

roaga: thanks deepseek

src/seer/automation/autofix/tools.py
```diff
@@ -44,27 +44,46 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.cleanup()
 
+    def _get_repo_names(self) -> list[str]:
+        if isinstance(self.context, AutofixContext):
+            return [f"{repo.owner}/{repo.name}" for repo in self.context.repos]
```
jennmueng, line 49: `RepoDefinition` has `.full_name` that gives the full name
```suggestion
            return [repo.full_name for repo in self.context.repos]
```

src/seer/automation/autofix/tools.py
```diff
@@ -44,27 +44,46 @@ def __enter__(self):
...
+        elif isinstance(self.context, CodegenContext):
+            return [f"{self.context.repo.owner}/{self.context.repo.name}"]
```
jennmueng, line 51:
```suggestion
            return [self.context.repo.full_name]
```

src/seer/automation/autofix/tools.py
```diff
@@ -363,10 +393,10 @@ def get_tools(self):
                     {
                         "name": "repo_name",
                         "type": "string",
-                        "description": "Optional name of the repository to search in if you know it.",
+                        "description": "Name of the repository containing the file.",
                     },
                 ],
-                required=["file_path"],
+                required=["file_path", "repo_name"],
             ),
```
jennmueng, line 400: I think we can still keep the repo name optional in expand document, especially for single repo cases, in multiple repo cases maybe we should try to match a filename and only throw an error back at the agent when there are path conflicts or path doesn't exist?

tests/automation/autofix/test_autofix_tools.py
jennmueng: 🐳

src/seer/automation/autofix/tools.py
roaga: i think it's probably good to force it learn the repo name because one common error state is that it outputs code changes without a repo name. Also, since all the other tools return results with repo name indicated, it should learn it pretty easily.

jennmueng: ok

jennmueng (approved):

PR #1930, opened 2025-02-13
trigger container rebuild

jennmueng:


requirements.txt
```diff
@@ -210,7 +210,7 @@ google-cloud-core==2.4.1
```
jennmueng, line 209: Verify the correct latest version of googleapis-common-protos. Based on the changelog, it should likely be 1.66.0 instead of 1.67.0.
*Suggested Fix*:
```suggestion
googleapis-common-protos==1.66.0
```

jennmueng, line 209: Check if `openai.NotGiven()` exists in v1.62.0 and confirm there are no breaking changes in the API surface used by the client code.

jennmueng, line 209: Check ChromaDB's requirements to ensure it supports PostHog 3.13.0. Look for any breaking changes in PostHog's 3.13.0 release notes that might affect ChromaDB integration.

jennmueng, line 209: Verify the 1.2.0 SDK maintains this API pattern and that the authentication handling (particularly `vertexai=True`) remains valid.

docker-compose.yml
```diff
@@ -36,6 +36,7 @@ services:
```
jennmueng, line 35: Consider removing this mount unless the application explicitly needs access to README.md during operation. Container images should generally only include runtime-essential files.

jennmueng (commented):
This PR updates several Python dependencies to newer versions and adds a README.md volume mount to the Docker container. The dependency updates appear to be routine version bumps, but some require compatibility checks. The Docker change adds documentation to the container but may not be necessary for runtime.

PR #1938, opened 2025-02-15
fix(autofix): Add fallback for code application step

roaga:
Occasionally, the apply model would not complete its response with a `</updated_code>` tag or it would be truncated for some other reason, causing the "new content" to be empty and cause the whole file to be deleted.

Now if there is no new content from the apply model (meaning something went wrong), fall back to asking Gemini Flash to directly write a unified diff. (I tested always just writing a diff, but the fix accuracy went down, so just using it as a fallback).

### DIFF
--- a/src/seer/automation/autofix/components/coding/component.py
+++ b/src/seer/automation/autofix/components/coding/component.py
@@ -38,7 +38,7 @@ from seer.automation.component import BaseComponent
 from seer.automation.models import FileChange
 from seer.automation.utils import escape_multi_xml, extract_text_inside_tags
 from seer.configuration import AppConfig
-from seer.dependency_injection import Module, inject, injected
+from seer.dependency_injection import copy_modules_initializer, inject, injected
 
 logger = logging.getLogger(__name__)
 
@@ -342,17 +342,6 @@ class CodingComponent(BaseComponent[CodingRequest, CodingOutput]):
         @observe(name="Process Change Task")
         @ai_track(description="Process Change Task")
         def process_task(task: CodeChangeXml, llm_client: LlmClient, app_config: AppConfig):
-            module = Module()
-            module.enable()
-
-            @module.provider
-            def provide_llm_client() -> LlmClient:
-                return llm_client
-
-            @module.provider
-            def provide_app_config() -> AppConfig:
-                return app_config
-
             repo_client = self.context.get_repo_client(task.repo_name)
             if task.type == "file_change":
                 file_content, _ = repo_client.get_file_content(task.file_path, autocorrect=True)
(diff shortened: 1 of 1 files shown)


src/seer/automation/autofix/components/coding/component.py
```diff
@@ -289,19 +334,29 @@ def invoke(self, request: CodingRequest) -> CodingOutput | None:
...
+            module.enable()
+
+            @module.provider
+            def provide_llm_client() -> LlmClient:
+                return llm_client
+
+            @module.provider
+            def provide_app_config() -> AppConfig:
+                return app_config
+
```
roaga, line 352: we don't actually need this in this PR, but I noticed when trying Claude that it broke DI since the Claude client needs the app config. So putting this here to make future swapping easy

src/seer/automation/autofix/components/coding/component.py
```diff
@@ -115,28 +117,70 @@ def apply_code_suggestion_to_file(
...
+                    updated_lines,
+                    fromfile=f"a/{file_path}",
+                    tofile=f"b/{file_path}",
+                )
             )
+            return diff
+
+        # fallback to writing a unified diff directly
+        fallback_prompt = textwrap.dedent(
+            """
```
jennmueng, line 147:
```suggestion
            """\
```

src/seer/automation/autofix/components/coding/component.py
```diff
@@ -115,28 +117,70 @@ def apply_code_suggestion_to_file(
...
+
+            @@ -1,3 +1,3 @@
+
+            x = 1
+            y = 2
+            -for i in range(10):
+            -    print(x + y)
+            +for i in range(20):
+            +    print(x * y)
+            </example>
```
jennmueng, line 165:
```suggestion
            <example>
            --- a/path/to/file.py
            +++ b/path/to/file.py
            @@ -1,3 +1,3 @@
            x = 1
            y = 2
            -for i in range(10):
            -    print(x + y)
            +for i in range(20):
            +    print(x * y)
            </example>
```

jennmueng (approved):

src/seer/automation/autofix/components/coding/component.py
```diff
@@ -22,6 +22,7 @@
```
jennmueng, line 21: The example diff format contains extra newlines after the ---/+++ lines that don't match actual unified diff syntax. Consider using a strictly correct example to guide the model's output better.

jennmueng, line 21: Consider adding basic validation of the generated diff format. Malformed diffs from the fallback could lead to incorrect file modifications.

jennmueng, line 21: Diff generation requires precise outputs. Consider using a lower temperature (e.g., 0.2) for the Gemini fallback to reduce randomness.

jennmueng, line 21: Consider normalizing line endings before diff generation to prevent issues with mixed line endings in the codebase.

jennmueng, line 21: Please confirm that the LLM client implementations (especially OpenAI/Gemini) are thread-safe when used across multiple threads concurrently.

jennmueng (commented):
This PR adds a fallback mechanism when code application via GPT models fails, using Gemini Flash to generate unified diffs directly. It also addresses dependency injection in multi-threaded environments and modifies prompt structures.

PR #1946, opened 2025-02-18
chore(autofix): Prettier file search log

jennmueng:
Add quotes around the query


roaga (changes requested):
I like it better without quotes

roaga (approved):

PR #1954, opened 2025-02-19
fix(autofix): Add retry for XML formatting in coding step

roaga:
If the coding output XML can't be parsed, ask the agent to fix it (up to one time).
Also update some prompts.

### DIFF
--- a/src/seer/automation/autofix/components/solution/prompts.py
+++ b/src/seer/automation/autofix/components/solution/prompts.py
@@ -152,12 +152,12 @@ class SolutionPrompts:
             Given the root cause of the issue:
             {root_cause_str}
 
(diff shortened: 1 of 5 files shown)


src/seer/automation/autofix/components/solution/prompts.py
```diff
@@ -152,12 +152,12 @@ def solution_formatter_msg(root_cause: RootCauseAnalysisItem | str):
             Given the root cause of the issue:
             {root_cause_str}
 
-            Based on the discussed plan, write the sequence of code changes needed to fix the issue.
+            Based on the discussed plan, write the steps needed to fix the issue.
```
jennmueng, line 155: this will def conflict lol but i guess ill deal with it

jennmueng (approved):

PR #1959, opened 2025-02-19
chore(anomaly-detection): Add prophet detector

ram-senth:


aayush-se (approved):
Changes look good.

I think PR should be renamed to "Add Prophet Detector Models" since adding the rest of the detector logic (pre-process, fit, predict, etc) still needs to be done in a separate PR. 

Also need to add tests too

src/seer/anomaly_detection/detectors/prophet_anomaly_detector.py
```diff
@@ -0,0 +1,206 @@
...
+        Initializes:
+            data: training dataset
+            test: test dataset
+            bc_lambda: box-cox lambda used to undo log transform
+
+        """
+        # TODO: Need to figure the right way of smoothing streaming data
+        # combined_y = pd.Series(
+        #     np.concatenate([train["y"].values, test_values]), dtype=np.float64
+        # )
```
aayush-se, line 56: nit: can remove leftover comments here and in the rest of this file

src/seer/anomaly_detection/detectors/prophet_anomaly_detector.py
```diff
@@ -0,0 +1,206 @@
...
+        Returns:
+            Test dataset with predictions added
+        """
+        future = self.model.make_future_dataframe(
+            periods=forecast_len, freq=f"{self.granularity}min"
+        )
+        forecast = self.model.predict(future)
+        forecast.index = forecast["ds"]
+        forecast.ds = pd.to_datetime(forecast.ds)
+        return forecast
```
aayush-se, line 152: forecast_len should be a field so it can be referenced with `self.forecast_len` throughout

src/seer/anomaly_detection/detectors/prophet_anomaly_detector.py
```diff
@@ -0,0 +1,206 @@
...
+from seer.anomaly_detection.models import AlgoConfig, Sensitivities
+from seer.dependency_injection import inject, injected
+
+
+class ProphetAnomalyDetector(BaseModel):
+    """
+    A detector that uses Prophet to detect anomalies in a time series.
+    """
+
+    pass
```
aayush-se, line 20: nit: remove

src/seer/anomaly_detection/detectors/prophet_anomaly_detector.py
```diff
@@ -0,0 +1,206 @@
...
+
```
aayush-se, line 21: nit: remove this placeholder

aayush-se (approved):
Looks good. Some leftover comments and nits + test coverage

src/seer/anomaly_detection/detectors/prophet_anomaly_detector.py
ram-senth: Don't think it is referred anywhere else in this file. Are you thinking of other places like storing in db? If so then it should be added to AlgoConfig?

src/seer/anomaly_detection/models/timeseries.py
```diff
@@ -10,6 +10,29 @@
 )
 
 
+class ProphetPrediction(BaseModel):
+    timestamps: npt.NDArray[np.float64] = Field(
```
aayush-se, line 14: nit: this should be singular `timestamp`

src/seer/anomaly_detection/models/timeseries.py
```diff
@@ -25,6 +48,11 @@ class TimeSeries(BaseModel):
         description="Anomalies identified in time series",
     )
 
+    prophet_prediction: Optional[ProphetPrediction] = Field(
```
aayush-se, line 51: this should be plural `prophet_predictions`

src/seer/anomaly_detection/detectors/prophet_anomaly_detector.py
aayush-se: You're right, it isn't referred to outside of this file. Though I think it would make intuitive sense to put it in the AlgoConfig

src/seer/anomaly_detection/models/timeseries.py
ram-senth: It is a list of timestamps. Still think it should be singular?

aayush-se: Ah sorry you're right was getting confused by another model

PR #1963, opened 2025-02-19
chore(evals): Improve final fix scoring prompt

jennmueng:
An issue with the prior prompt was that we were having it overfocus on scoring how well the predicted solution "matches" an expected output, rather than scoring how well the issue is fixed.

This prompt update:
1. Changes the concept of an "expected" solution to a "known" solution and prompts o3-mini to use it to inform what the issue was
2. The score is now on how well the predicted solution fixes the issue rather than asking it to score how well it "matches"
3. Prompt format adjustment to place context at the end for reasoning models.

(ignore the latency, i was running multiple evals at the same time so it slowed)
CleanShot 2025-02-19 at 15 22 39@2x

### DIFF
--- a/src/seer/automation/autofix/evaluations.py
+++ b/src/seer/automation/autofix/evaluations.py
@@ -260,32 +260,45 @@ def score_fix_single_it(
 
     prompt = textwrap.dedent(
         """\
+            <goal>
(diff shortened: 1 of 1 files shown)


roaga (approved):
Seems fine to me, but I hope we're not being too lenient or letting the scoring model interpret too much

PR #1966, opened 2025-02-20
feat(autofix): New solution step architecture

roaga:
Mirrors the root cause step with a gemini flash context gatherer and an o3 mini reasoner.

Doubles run cost and adds a minute of latency, but seems to make solutions more targeted and increases fix evals a lot (64% to 71%).


jennmueng (approved):
🚢

PR #1968, opened 2025-02-20
feat(autofix): Solution formatter outputs SolutionPlanStep

jennmueng:
Prompt adjustments to solution formatter and it's output format to be more aligned. This ensures the formatter actually outputs steps and doesn't confuse it for a timeline. And also do more code snippets

Before:
CleanShot 2025-02-20 at 11 21 05@2x
After:
CleanShot 2025-02-20 at 11 20 50@2x


src/seer/automation/autofix/components/solution/models.py
```diff
@@ -17,11 +17,16 @@ class RelevantCodeFile(BaseModel):
 class SolutionTimelineEvent(BaseModel):
     title: str
     code_snippet_and_analysis: str
-    timeline_item_type: (
-        Literal["internal_code", "external_system", "human_action"] | str
-    )  # TODO put back to literal only when not breaking anything
     relevant_code_file: RelevantCodeFile | None
-    is_new_event: bool
+    is_most_important_event: bool
+    event_type: Literal["internal_code"] = "internal_code"
```
roaga, line 22: won't this break backwards compatability

src/seer/automation/autofix/components/solution/models.py
```diff
@@ -34,4 +39,4 @@ class SolutionRequest(BaseComponentRequest):
 
 
 class SolutionOutput(BaseComponentOutput):
-    modified_timeline: list[SolutionTimelineEvent]
+    solution_steps: list[SolutionPlanStep]
```
roaga, line 42: won't this break the frontend

roaga (commented):
I don't understand the need for this change. Like sure conceptually, but it has no affect on evals and we're getting proper user-facing solution steps already with the latest change

src/seer/automation/autofix/components/solution/models.py
jennmueng: solution_steps is not what's sent to the frontend right, we convert this in event manager

roaga (approved):

PR #1969, opened 2025-02-20
feat(anomaly detection): Predict Using Prophet 

aayush-se:
Implements changes to include prophet predictions in the anomaly detection system:
- Created new tables and updated existing ORM to store Prophet related info
- Created new models to support Prophet Predictions and updated existing ones to include prophet as required
- Create async prediction Celery tasks which make future predictions and store them for use in prediction
- Update cleanup and other CRUD operations to include Prophet and Prophet History


src/migrations/versions/384911696acd_migration.py
```diff
@@ -0,0 +1,68 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "prophet_alert_time_series",
+        sa.Column("id", sa.BigInteger(), autoincrement=True, nullable=False),
+        sa.Column("dynamic_alert_id", sa.BigInteger(), nullable=False),
+        sa.Column("timestamp", sa.DateTime(), nullable=False),
+        sa.Column("yhat", sa.Float(), nullable=False),
+        sa.Column("yhat_lower", sa.Float(), nullable=False),
+        sa.Column("yhat_upper", sa.Float(), nullable=False),
+        sa.PrimaryKeyConstraint("id", "dynamic_alert_id"),
```
ram-senth, line 29: Think primary key should be just "id" it is the truely unique per row.

src/migrations/versions/384911696acd_migration.py
```diff
@@ -0,0 +1,68 @@
...
+    op.create_table(
+        "prophet_alert_time_series_history",
+        sa.Column("id", sa.BigInteger(), autoincrement=True, nullable=False),
+        sa.Column("alert_id", sa.BigInteger(), nullable=False),
+        sa.Column("timestamp", sa.DateTime(), nullable=False),
+        sa.Column("yhat", sa.Float(), nullable=False),
+        sa.Column("yhat_lower", sa.Float(), nullable=False),
+        sa.Column("yhat_upper", sa.Float(), nullable=False),
+        sa.Column("saved_at", sa.DateTime(), nullable=False),
+        sa.PrimaryKeyConstraint("id", "alert_id"),
```
ram-senth, line 48: Same here, either just id should be the primary key or alert_id + timestamp should be the primary key

src/seer/anomaly_detection/accessors.py
```diff
@@ -152,6 +159,16 @@ def _hydrate_alert(
...
                 num_old_points += 1
 
+        num_predictions_remaining = 0
+        for i, prediction in enumerate(db_alert.prophet_predictions):
+            prophet_timestamp = prediction.timestamp.timestamp()
+            prophet_timestamps[i] = prophet_timestamp
+            prophet_yhats[i] = prediction.yhat
+            prophet_yhat_lowers[i] = prediction.yhat_lower
+            prophet_yhat_uppers[i] = prediction.yhat_upper
+            if prophet_timestamp > datetime.now().timestamp():
```
ram-senth, line 169: minor optimization - call `datetime.now().timestamp()` outside the loop and store it in a variable?

src/seer/anomaly_detection/tasks.py
```diff
@@ -163,6 +207,69 @@ def _update_matrix_profiles(
...
+
+    prophet_detector = ProphetAnomalyDetector()
+
+    timestamps = np.array([0.0] * len(alert.timeseries))
+    values = np.array([0.0] * len(alert.timeseries))
+    for i, ts in enumerate(alert.timeseries):
+        timestamps[i] = ts.timestamp.timestamp()
+        values[i] = ts.value
+
+    forecast_len = random.randint(22, 28) * (60 // config.time_period)
```
ram-senth, line 219: Curious, why do we want randomness in the length of predictions?

src/migrations/versions/384911696acd_migration.py
ram-senth: There are four more floats in the prophet prediction dataframe - trend, additive_term, weekly and multiplicative_terms. Wondering, should we save them as well for troubleshooting?

src/seer/anomaly_detection/tasks.py
aayush-se: This was the "jitter" which we were discussing earlier. Basically, if we include some jitter in the length of predictions, then the amount of overlap when the subsequent predict is call should reduce since the check for that is if there are less than 12 hours remaining.

ram-senth: I was actually meaning that what ever time period we end up predicting for should cover the jitter we add in scheduling the task. Did not mean to add jitter in the prediction length. Sorry for not being very clear.

src/migrations/versions/384911696acd_migration.py
aayush-se: I was following the same format as the `DbDynamicAlertTimeSeries` schema which is why the primary key is such.

Screenshot 2025-02-25 at 1 43 26 PM

aayush-se: I can add those other floats in. Just wondering what troubleshooting they would serve with though since my understanding is that `yhat`, `yhat_lower`, and `yhat_upper` should provide enough information to display what prophet was predicting for the respective timeseries?

ram-senth: So DbDynamicAlertTimeSeries has just id as primary key. 
image
image

ram-senth: We should also add a created_at field with the datetime when the record is added to the db for record keeping.

aayush-se: Updated

src/migrations/versions/e0fcdc14251c_migration.py
```diff
@@ -0,0 +1,70 @@
...
+        sa.Column("created_at", sa.DateTime(), nullable=False),
+        sa.ForeignKeyConstraint(["dynamic_alert_id"], ["dynamic_alerts.id"], ondelete="CASCADE"),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("dynamic_alert_id", "timestamp"),
+    )
+    with op.batch_alter_table("prophet_alert_time_series", schema=None) as batch_op:
+        batch_op.create_index(
+            "ix_prophet_alert_time_series_alert_id_timestamp",
+            ["dynamic_alert_id", "timestamp"],
+            unique=False,
```
ram-senth, line 54: dynamic_alert_id + timestamp should be unique, right?

src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -157,6 +158,18 @@ def _online_detect(
                 },
             )
             raise ServerError("Invalid state")
+        if (
+            historic.data_purge_flag == TaskStatus.QUEUED
+            or historic.data_purge_flag == TaskStatus.PROCESSING
+        ):
+            logger.info(
+                "data_purge_flag",
```
ram-senth, line 166: Maybe "data_purge_flag_invalid" to make it a little more obvious that this is an unexpected state? Also, `logger.warn` instead of `logger.info`?

ram-senth: Also, I wouldn't worry about queued here, especially since we want the jitter to be up to 3600 seconds.

src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -240,22 +253,27 @@ def _online_detect(
...
+                or cleanup_predict_config.num_predictions_remaining
+                <= cleanup_predict_config.num_acceptable_predictions
             ):
                 alert_data_accessor.queue_data_purge_flag(historic.external_alert_id)
-                cleanup_timeseries.delay(
-                    historic.external_alert_id, cleanup_config.timestamp_threshold
+                cleanup_timeseries_and_predict.apply_async(
+                    (historic.external_alert_id, cleanup_predict_config.timestamp_threshold),
+                    countdown=random.randint(
+                        0, 120
```
ram-senth, line 271: I would make it a bigger window - 0 to 3600 might be better.

src/seer/anomaly_detection/tasks.py
```diff
@@ -163,6 +204,67 @@ def _update_matrix_profiles(
...
+    prophet_detector = ProphetAnomalyDetector()
+
+    timestamps = np.array([0.0] * len(alert.timeseries))
+    values = np.array([0.0] * len(alert.timeseries))
+    for i, ts in enumerate(alert.timeseries):
+        timestamps[i] = ts.timestamp.timestamp()
+        values[i] = ts.value
+
+    # Create 24 hours worth of predictions
+    forecast_len = 24 * (60 // config.time_period)
```
ram-senth, line 219: Minor, didn't you want to add the number of hours to predict to the algo_config class?

src/seer/anomaly_detection/anomaly_detection.py
aayush-se: 3600 seconds for each of the time windows should be fine right? Or would it make more sense to have it be [0, time window in seconds]?

ram-senth: Yes that is a better idea, use the time window.

src/migrations/versions/e0fcdc14251c_migration.py
aayush-se: Taking a look at previous migrations (`dynamic_alert_time_series` below), this seems to be for the index. The Prophet table does have the `UniqueConstraint` wrapper on it for the alert_id and timestamp. Also, testing locally I get an Error for violating the uniqueness constraint if I try to add a point with the same alert_id and timestamp to the table.

Screenshot 2025-02-25 at 5 01 11 PM
Screenshot 2025-02-25 at 5 01 48 PM

ram-senth: Ok verified that the table actually have it as unique.

ram-senth (approved):

PR #1978, opened 2025-02-21
fix(autofix): Better insight cards, question asking, plus more

roaga:
- better insight card prompts
- less dumb questions
- raise max iterations because gemini likes to _move_
- raise error when no coding output so step is retried properly 

Evals were stable if not slightly better at 71% fixed. I don't think this fixes the extra unit tests issue, however I did notice that those tests are introduced in the coding step, not solution step, at least in the cases I looked at.


src/seer/automation/autofix/components/root_cause/prompts.py
```diff
@@ -16,7 +16,7 @@ class RootCauseAnalysisPrompts:
     def format_system_msg(has_tools: bool = True):
         return textwrap.dedent(
             f"""\
-            You are an exceptional principal engineer that is amazing at researching bugs in codebases.
+            You are an exceptional AI system that is amazing at researching bugs in codebases.
```
aayush-se, line 19: I'm surprised that "AI system" would perform better than "principal/staff level engineer". I wonder why this is?

roaga: Saying that it was an engineer made it think that it should write code, deploy code, propose fixes, work with team members, write tests, and do a bunch of "best practice" stuff that we were not asking it to do directly. Meanwhile an AI system obeys every command :)

src/seer/automation/autofix/components/coding/component.py
```diff
@@ -361,11 +361,11 @@ def invoke(self, request: CodingRequest) -> CodingOutput | None:
                 cur.usage += agent.usage
 
             if not response:
-                return None
+                raise ValueError("No response from coding agent")
 
             code_changes_output = self._parse_code_changes_xml(response, agent)
             if not code_changes_output:
-                return None
+                raise ValueError("No code changes output from coding agent")
```
jennmueng, line 368: good call to just error here instead, worth looking elsewhere where we should be erroring instead of doing nothing

src/seer/automation/autofix/components/root_cause/prompts.py
jennmueng: I'm surprised this increases evals meanwhile when i tell it to write only code it does worse lmao

jennmueng (approved):

PR #1993, opened 2025-02-25
fix(autofix) Retry null Gemini structured outputs

kddubey:
Gemini can return a structured output where `completion.parsed is None`. These are handled by downstream uses of `generate_structured`, and retrying adds some latency. But it's probably worth it if it means getting an output.


src/seer/automation/autofix/components/root_cause/component.py
```diff
@@ -153,23 +153,14 @@ def invoke(
                     "Arranging data in a way that looks intentional..."
                 )
 
-                # Add retry logic for generating structured response
```
kddubey, line 156: quite sure this is the only place where we retry

roaga (approved):

PR #1995, opened 2025-02-25
Anomaly detection/combined scorer

ram-senth:


src/seer/anomaly_detection/accessors.py
```diff
@@ -298,6 +312,8 @@ def save_alert(
             )
             session.add(new_record)
             session.commit()
+            session.refresh(new_record, attribute_names=["id"])
+            return new_record.id
```
aayush-se, line 316: It seems like we don't need to refresh to get the id:

https://stackoverflow.com/questions/62396638/retrieve-id-after-commit

src/seer/anomaly_detection/accessors.py
```diff
@@ -459,3 +475,39 @@ def combine_anomalies(
...
+                        yhat=yhat,
+                        yhat_lower=yhat_lower,
+                        yhat_upper=yhat_upper,
+                    )
+                    session.add(prophet_prediction)
+            session.commit()
+
+        logger.info(
+            f"Stored {len(predictions.timestamps)} prophet predictions for alert {alert_id}"
+        )
```
aayush-se, line 513: Just remember to update this upon the landing of the new store changes

src/seer/anomaly_detection/detectors/anomaly_detectors.py
```diff
@@ -40,7 +41,8 @@ def detect(
         self,
         timeseries: TimeSeries,
         ad_config: AnomalyDetectionConfig,
-        algo_config: AlgoConfig,
+        algo_config: AlgoConfig = injected,
```
aayush-se, line 44: @inject decorator should be added because injecting config

src/seer/anomaly_detection/detectors/anomaly_scorer.py
```diff
@@ -0,0 +1,258 @@
...
+                    )
+                    if (
+                        mp_flag == "anomaly_higher_confidence"
+                        and prophet_flag == "anomaly_higher_confidence"
+                    ):
+                        flags.append("anomaly_higher_confidence")
+                    elif prophet_score >= 2.0:
+                        flags.append(prophet_flag)
+                    else:
+                        flags.append("none")
```
aayush-se, line 239: shouldn't the MP override also be included here? As in if the score is significantly greater than the threshold (I believe 3x was the sweet spot), then the MP flags should be used.

ram-senth: You will be able to add itas a follow-up PR? I haven't looked at that change that you made yet.

src/seer/anomaly_detection/accessors.py
ram-senth: Done

src/seer/anomaly_detection/accessors.py
ram-senth: Done

src/seer/anomaly_detection/detectors/anomaly_detectors.py
ram-senth: Done

src/seer/anomaly_detection/accessors.py
```diff
@@ -107,10 +118,11 @@ def _hydrate_alert(
...
         n_predictions = len(db_alert.prophet_predictions)
-        prophet_timestamps = np.array([0.0] * n_predictions)
-        prophet_yhats = np.array([0.0] * n_predictions)
-        prophet_yhat_lowers = np.array([0.0] * n_predictions)
-        prophet_yhat_uppers = np.array([0.0] * n_predictions)
+        prophet_timestamps = np.array([None] * n_predictions)
+        prophet_ys = np.array([None] * n_predictions)
+        prophet_yhats = np.array([None] * n_predictions)
+        prophet_yhat_lowers = np.array([None] * n_predictions)
+        prophet_yhat_uppers = np.array([None] * n_predictions)
```
aayush-se, line 125: performance nit: would it be better to use `np.full(n_predictions, None)`? This would just create the array directly rather than performing a conversion

ram-senth: Think this was something you added? I just changed the values to None instead of 0.0 as 0.0 can affect scores.

aayush-se: Yes you're right I did add these to pass down the prophet related info. Was reviewing the `_hydrate_alert` function prior to this so was thinking of potential optimizations. I can add this as a followup PR if needed

ram-senth: It is fine, I can make the changes. It is a minor one.

src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -81,12 +84,23 @@ def _batch_detect(
...
+
+        forecast_len = algo_config.prophet_forecast_len * (60 // config.time_period)
+        ts_internal = convert_external_ts_to_internal(timeseries)
+        prophet_df = prophet_detector.predict(
+            ts_internal.timestamps,
+            ts_internal.values,
+            forecast_len,
+            config.time_period,
+            config.sensitivity,
+        )
```
aayush-se, line 97: qq: Could we use the cached predictions in the db rather than re-running predict here?

aayush-se: and/or does forecast_len need to be > 0 in this case? Since specifically we just want the prophet predictions for what we currently have (history) and are not storing the future predictions in this function. I assume this would make this function slightly faster?

ram-senth: This method is called from both store_data and batch_predict calls. In the store_data path, the result from here is stored in DB. Agreed that the we do not need to predict for the batch_detection path but then that API is not being used actively in prod at this time. So I would not put it high on the list.

src/seer/anomaly_detection/detectors/anomaly_scorer.py
```diff
@@ -0,0 +1,273 @@
...
+            history_mp_dist=history_mp_dist,
+            ad_config=ad_config_for_mp,
+            window_size=window_size,
+            algo_config=algo_config,
+        )
+        if mp_flags_and_scores is None:
+            raise ServerError("No flags and scores from MP scorer")
+
+        if prophet_df is None or prophet_df.empty:
+            # logger.warning("The prophet_df is None or empty, skipping prophet scoring")
```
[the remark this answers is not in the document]
ram-senth: This will cause a bunch of warnings in the case of combo_detect at present, where a batch detection is followed by a bunch of stream detection calls. I will be adding this back once I update that in a followup pr.

src/seer/anomaly_detection/detectors/anomaly_scorer.py
```diff
@@ -0,0 +1,273 @@
...
+    ) -> AnomalyFlags:
+        if (
+            direction == "both"
+            or mp_flag == "none"
+            or mp_flag == prev_mp_flag
+            or prophet_flag == "none"
+        ):
+            return prophet_flag
+
+        if (direction == "up" and y >= yhat_upper) or (direction == "down" and y <= yhat_lower):
```
aayush-se, line 212: Should this just be strictly greater/less than? I noticed thats what it was in the original LocationDetectors too. Probably doesn't make a large difference otherwise though.

src/seer/anomaly_detection/detectors/anomaly_scorer.py
```diff
@@ -0,0 +1,273 @@
...
+
+        # todo: return prophet thresholds
+        def merge(timestamps, mp_flags, prophet_map):
+            flags = []
+            found = 0
+            previous_mp_flag: AnomalyFlags = "none"
+            for timestamp, mp_flag in zip(timestamps, mp_flags):
+                if pd.to_datetime(timestamp) in prophet_map["flag"]:
+                    found += 1
+                    pd_dt = pd.to_datetime(timestamp)
```
aayush-se, line 233: I believe it would be faster to convert the entire timestamp array to pd datetimes first then iterate over the converted values rather than converting each element within the loop

src/seer/anomaly_detection/detectors/anomaly_scorer.py
ram-senth: Doesn't make any noticeable difference.

src/seer/anomaly_detection/detectors/anomaly_scorer.py
```diff
@@ -0,0 +1,273 @@
...
+                        yhat_lower=prophet_map["yhat_lower"][pd_dt],
+                        yhat_upper=prophet_map["yhat_upper"][pd_dt],
+                        direction=ad_config.direction,
+                    )
+                    if (
+                        mp_flag == "anomaly_higher_confidence"
+                        and prophet_flag == "anomaly_higher_confidence"
+                    ):
+                        flags.append("anomaly_higher_confidence")
+                    elif prophet_score >= 2.0:
```
aayush-se, line 251: I noticed this threshold was >= 5.0 on [the data-analysis notebook](https://github.com/getsentry/data-analysis/blob/main/anomaly_detection/stumpy-noise-correction-plus-slope.ipynb) (I can't select the exact lines of code but if you cmd + f `merge_prophet_mp_results` its in that function). Is there a reason this was dropped to 2.0?

src/seer/anomaly_detection/detectors/anomaly_scorer.py
```diff
@@ -0,0 +1,273 @@
...
+                        flags.append(prophet_flag)
+                    else:
+                        flags.append("none")
+                else:
+                    flags.append(mp_flag)
+                previous_mp_flag = mp_flag
+            # todo: publish metrics for found/total
+            # if debug:
+            #     print(f"found {found} out of {len(timestamps)}")
```
aayush-se, line 260: nits: Should this just be added now as a log.info()? Also extra debug comment?

src/seer/anomaly_detection/detectors/anomaly_scorer.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def _merge_prophet_mp_results(
+        self,
+        timestamps: np.ndarray,
+        mp_flags_and_scores: FlagsAndScores,
+        prophet_predictions: pd.DataFrame,
+        ad_config: AnomalyDetectionConfig,
+    ) -> FlagsAndScores:
+
+        # todo: return prophet thresholds
```
aayush-se, line 225: qq: I assume adding this would be a followup PR to support the UI changes for the confidence interval?

src/seer/anomaly_detection/detectors/anomaly_scorer.py
ram-senth: For batch detection we are looking up at most 96 timestamps and for streaming we are looking up 1 timestamp. The prediction data can be up to 2688 long. So I do not think updating the entire DataFrame will be more efficient.

src/seer/app.py
```diff
@@ -376,7 +376,6 @@ def ready_check(app_config: AppConfig = injected):
     from seer.inference_models import models_loading_status
 
     status = models_loading_status()
-    logger.info(f"Model loading status: {status}")
```
aayush-se, line 379: Why is this removed?

src/seer/anomaly_detection/detectors/anomaly_scorer.py
ram-senth: Wasn't that part of your recommendation from your experiments for that one timeseries with an extended flatline at 0? Or am I misremembering it? Regardless, all my experiments in these past few days that we evaluated together and with Tillman has been using 2.0.

aayush-se (approved):
Looks great! No structural changes just had a couple of nits and questions.

src/seer/anomaly_detection/detectors/anomaly_scorer.py
ram-senth: Actually I'm actively working on this. I enabled this log and noticed that there is a difference in how timestamps are treated by the code that retrieves data from DB, resulting in stream computations not finding the datapoint. Am trouble shooting it.

src/seer/anomaly_detection/detectors/anomaly_scorer.py
ram-senth: Yes

src/seer/app.py
ram-senth: This was causing a log entry to be published every second, increasing the noise in logs.

aayush-se: Just asked jenn/rohan about this. It isnt used for autofix, but may be helpful for grouping? 

Would it make more sense to push this log to a few lines below within the if statements (FAILED/DONE checks on lines 384-387)?

src/seer/anomaly_detection/detectors/anomaly_scorer.py
aayush-se: FWIW it may be the discrepancy in the timestamps stored in DB (seconds) vs the pd timestamps which I believe defaults to 'ns': https://pandas.pydata.org/docs/reference/api/pandas.to_datetime.html

src/seer/app.py
ram-senth: I do not think there is much value in this log. Let me know if it is helpful for grouping.

aayush-se: Just reviewed the file this log can probably just be removed.

aayush-se (approved):
Changes look good

PR #2004, opened 2025-02-26
fix(relevant-warnings): Fix serialization on overwatch payload

suejung-sentry:
The overwatch payload expects the json serialization to exclude spaces e.g. like `{"run_id":"test-123"}` not `{"run_id": "test-123"` (note the space before "test-123") and since the exact string needs to match when signed with the shared secret, need to add this extra serialization rule


kddubey (approved):
👍

PR #2005, opened 2025-02-26
fix(autofix): Timeout + retry on stream first token and hanging

jennmueng:
Autofix run timeouts turns out happens (so far as we know) on 2 reasons:
1. A deploy goes out and kills the run
2. Streams hanging, especially from anthropic

This PR fixes the latter stream hanging issue by introducing a time to first token timeout and a stream inactivity timeout.

### DIFF
--- a/src/seer/automation/agent/client.py
+++ b/src/seer/automation/agent/client.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import re
+import time
 from dataclasses import dataclass
 from typing import Any, ClassVar, Iterable, Iterator, Type, Union, cast
 
@@ -36,6 +37,9 @@ from seer.automation.agent.models import (
     LlmProviderType,
     LlmRefusalError,
     LlmResponseMetadata,
+    LlmStreamFirstTokenTimeoutError,
+    LlmStreamInactivityTimeoutError,
+    LlmStreamTimeoutError,
     Message,
     StructuredOutputType,
     ToolCall,
@@ -96,7 +100,9 @@ class OpenAiProvider:
 
     @staticmethod
     def is_completion_exception_retryable(exception: Exception) -> bool:
-        return isinstance(exception, openai.InternalServerError)
+        return isinstance(exception, openai.InternalServerError) or isinstance(
+            exception, LlmStreamTimeoutError
+        )
 
     def generate_text(
         self,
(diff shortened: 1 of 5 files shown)


src/seer/automation/agent/client.py
```diff
@@ -1264,6 +1272,8 @@ def generate_text_stream(
         run_name: str | None = None,
         timeout: float | None = None,
         reasoning_effort: str | None = None,
+        first_token_timeout: float = 40.0,  # Time to first token timeout
```
roaga, line 1275: is this timeout for first token enough for a model like o3-mini-high? It might take a while for it to return something

src/seer/automation/agent/client.py
```diff
@@ -1313,6 +1324,34 @@ def generate_text_stream(
...
+
+                if current_time - last_yield_time > timeout_to_use:
+                    if first_token_received:
+                        raise LlmStreamInactivityTimeoutError(
+                            f"Stream inactivity timeout after {timeout_to_use} seconds"
+                        )
+                    else:
+                        raise LlmStreamFirstTokenTimeoutError(
+                            f"Stream time to first token timeout after {timeout_to_use} seconds"
+                        )
```
roaga, line 1348: when this error is raised, it's this one LLM call that is retried, right? Not the whole step?

tests/automation/agent/test_client.py
```diff
@@ -827,3 +832,198 @@ def test_encode(self, model: GoogleProviderEmbeddings):
         for text, embedding in zip(texts, embeddings, strict=True):
             embedding_expected = model.encode([text])[0]
             assert np.allclose(embedding, embedding_expected, atol=1e-4)
+
+
+@pytest.mark.parametrize(
+    "provider_class,model_name",
+    [
+        (OpenAiProvider, "gpt-3.5-turbo"),
```
roaga, line 840: i haven't heard that name in years

src/seer/automation/agent/client.py
jennmueng: yes

src/seer/automation/agent/client.py
jennmueng: hmm I forgot that openai doesn't start streaming until it's done thinking... let's see if we can pass a special case for the o models

tests/automation/agent/test_client.py
jennmueng: claude 3.7 remembers its old friend fondly i guess

PR #2017, opened 2025-02-27
chore(anomaly detection): Remove Fixed Window Logic

aayush-se:
**Draft PR waiting for Combined Logic to be implemented then will rebase these changes on top of that**

- Comment out calculations of the fixed window for the Dynamic Window
- Ensures that we only use the SuSS window for the MP calculations
- Enables simple rollback to using Dynamic Window logic again in the future


aayush-se:
Completed as part of combined scorer PR

PR #2022, opened 2025-02-27
chore(seer): Use Migration URL

aayush-se:
- Uses `DATABASE_MIGRATION_URL` if present which includes the proper permissions for creating migrations
- Defaults to the standard `DATABASE_URL` otherwise


jennmueng (approved):
I'm not 100% sure whether this is gonna work either but lets merge and try

PR #2027, opened 2025-02-27
fix(relevant-warnings) Send repo name for fetching issues

kddubey:
corresponding change for https://github.com/getsentry/sentry/pull/86011

these PRs can be deployed independently b/c `repo_full_name` is optional in the RPC method


kddubey:
looks like they've figured it out

PR #2031, opened 2025-02-28
fix(anomaly detection): Prophet Validation Issue

aayush-se:
- Fix validation issue upon adding new prophet entries
- Ensure current timestamp does not get overwritten

### DIFF
--- a/src/seer/anomaly_detection/tasks.py
+++ b/src/seer/anomaly_detection/tasks.py
@@ -250,21 +250,23 @@ def _store_prophet_predictions(alert: DbDynamicAlert, predictions: ProphetPredic
 
         session.query(DbProphetAlertTimeSeries).filter(
             DbProphetAlertTimeSeries.dynamic_alert_id == alert.id,
-            DbProphetAlertTimeSeries.timestamp >= min_timestamp,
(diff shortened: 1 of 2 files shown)


kddubey (approved):

trillville (approved):

src/seer/anomaly_detection/tasks.py
```diff
@@ -250,21 +250,23 @@ def _store_prophet_predictions(alert: DbDynamicAlert, predictions: ProphetPredic
...
-            prophet_prediction = DbProphetAlertTimeSeries(
-                dynamic_alert_id=alert.id,
-                timestamp=datetime.fromtimestamp(timestamp),
-                yhat=yhat,
-                yhat_lower=yhat_lower,
-                yhat_upper=yhat_upper,
-            )
-            session.add(prophet_prediction)
+            timestamp = datetime.fromtimestamp(timestamp)
+            if timestamp > min_timestamp:
```
ram-senth, line 261: If I read this right, the prophet table can have a mix of predictions - some records from a past prediction and some from current. I was hoping to avoid that troubleshooting will be a nightmare, especially around some big nomaly that is dropping off the 28 day window in the current pruning run. We should discuss this some more.

PR #2032, opened 2025-02-28
fix(anomaly detection): Change Store Query to Upsert

aayush-se:
- Changes the store query to be an upsert so we can include all the new predictions
- Update test to ensure timestamps do not include seconds as in production

### DIFF
--- a/src/seer/anomaly_detection/tasks.py
+++ b/src/seer/anomaly_detection/tasks.py
@@ -1,13 +1,13 @@
 import logging
 from datetime import datetime, timedelta
 from operator import and_, or_
-from sqlite3 import IntegrityError
 from typing import List
 
 import numpy as np
 import sentry_sdk
 import stumpy  # type: ignore # mypy throws "missing library stubs"
 from sqlalchemy import delete
+from sqlalchemy.dialects.postgresql import insert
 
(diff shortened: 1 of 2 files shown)


src/seer/anomaly_detection/tasks.py
```diff
@@ -163,7 +164,13 @@ def _save_timeseries_history(
                 yhat_upper=prophet_ts.yhat_upper,
                 saved_at=datetime.now(),
             )
-            session.add(prophet_history_record)
+            try:
+                session.add(prophet_history_record)
+                session.flush()
+            except IntegrityError:
+                # Roll back and skip the record as it already exists
+                session.rollback()
```
ram-senth, line 172: Did you get a chance to test this rollback? If I understand it right, all added until this point will be rolled back. What are the situations when a duplicate record is added? Is it ok to assume that data in the table is the correct and discard the new data?

aayush-se: Thinking about it, there should not be any cases where a duplicate record is added because the history from the prophet prediction shouldn't include that timepoint because we are generating it on new data.

ram-senth (approved):

PR #2037, opened 2025-02-28
feat(issue-summary) gpt-4o-mini -> gemini-2.0-flash-lite

kddubey:
Removes issue summary dependence on OpenAI. p50 latency is 0.1 sec higher, i.e., still slightly slower than 4o-mini + embeddings. We prompt engineer instead of embedding b/c flash-lite follows instructions.

This PR will cause the possible cause to always show while mostly retaining its current performance.

Anecdotal looks slightly better. The "subtly phrase your uncertainty" part doesn't cause it to hedge often.

In a follow-up PR in sentry, I'll change the frontend to show the possible cause loading. This PR can be deployed independently of that, b/c the frontend will show the possible cause if scores are null (tested locally as well)


src/seer/automation/summarize/issue.py
```diff
@@ -106,34 +80,54 @@ def summarize_issue(
         Your #1 goal is to help our engineers immediately understand what's going on in the main issue.
 
         Write a concise report that sets the scene for the issue. Know that our engineers can already see all the same details on the issue that were provided to you, so DO NOT repeat them. Instead extract the holistic insights that would take time to figure out otherwise.
-        Please write under 20 words per section, while cramming as much detail as possible, to make the report super easy to skim. Please bold important insights and unique terms by surrounding them with double asterisks (**like this**). The structure you should follow is:
+        Please write under 15 words per section, while cramming as much detail as possible, to make the report super easy to skim. Please bold important insights and unique terms by surrounding them with double asterisks (**like this**). The structure you should follow is:
```
kddubey, line 83: 4o-mini outputs 15 when you tell it to output 20. To keep behavior the same, have Gemini output 15. I checked that performance doesn't change w/ 20 for Gemini. Anecdotally, the 15-word Gemini is a bit shorter and sweeter

src/seer/automation/summarize/issue.py
```diff
@@ -34,35 +30,13 @@ class IssueSummaryWithScores(IssueSummary):
...
-        possible_cause_confidence=possible_cause_confidence.item(),
-        possible_cause_novelty=possible_cause_novelty,
-    )
+class IssueSummaryForLlmToGenerate(BaseModel):
+    whats_wrong: str
+    session_related_issues: str
+    possible_cause: str
+    possible_cause_novelty_score: float
+    possible_cause_confidence_score: float
+    title: str
```
kddubey, line 39: intentionally moved title to bottom so that it has above sections in context (pretty sure gemini respects field order). though we don't use title/headline in the front end

src/seer/automation/summarize/issue.py
kddubey: here was 4o-mini's distribution for # words when asked for 20 words in possible cause

openai_possible_cause_20

here is gemini's when asked for 15

gemini_possible_cause_15

src/seer/automation/summarize/issue.py
```diff
@@ -106,34 +80,54 @@ def summarize_issue(
...
-        insights from the application session issues, if relevant to this issue [return empty string if none]
+        Insights from the application session issues, if relevant to this issue [you must return an empty string here if there are no relevant session related issues]
 
         ###### Possible cause [optional]
-        guess as to the cause, maybe show if there's clear smoking bullet [return empty string if none]
-
-        Also provide a concise title for the report that summarizes everything you write above.
+        Guess as to the cause, maybe show if there's clear smoking bullet. [return empty string if none]
+        The guess should be somewhat novel compared to the `whats_wrong` section.
+        If you're not sure about the guess, express this uncertainty subtly in your guess, e.g., by using words like "possible", "perhaps", "may be", etc.
```
kddubey, line 94: I should ablate this line, one sec. I wanna see if it matters for the confidence score. Will update

roaga:
This LGTM, but why do you want to change the frontend to show the Possible Cause loading? Isn't there still a possibility that we don't show it?

roaga (approved):

jennmueng (approved):
lgtm, seconding @roaga's q about why do we need to show the frontend loading

roaga:
Based on Kush's offline message, we're not dropping any causes anymore, so it makes sense to show the cause loading then

trillville (approved):
🔥

src/seer/automation/summarize/issue.py
kddubey: @roaga @jennmueng it does matter. Accuracy is similar, but the confidence score is significantly worse w/o this line.

w/o the `"If you're not sure..."` line, 1 - AUC = `0.65`:

image

w/ the `"If you're not sure..."`, 1 - AUC = `0.73`:

image

Repeated both settings a couple of times and this comparison is quite stable.

Lesson learned for confidence scores: first ask it to phrase any uncertainty, then generate the score.

PR #2044, opened 2025-02-28
feat(autofix): Add tools for commit history

roaga:
Adds two new tools:
- Explain File: given a file, show the descriptions of the 30 most recent commits as well as a tree of what other files were touched in the same commit.
- View Diff: given a 7-char commit sha and a specific file, view the diff of the change made.

Both tools typically complete in 2-5s. I think the latency increase in the evals is because the agent decides to explore more files after seeing the connected files, which also leads to better solutions.

Eval results:
- Root cause stable
- Solution correctness: 55% -> 62%
- Code conciseness: 91% -> 95%
- Code correctness: 86% -> 88%

### DIFF
--- a/src/seer/automation/autofix/evaluations.py
+++ b/src/seer/automation/autofix/evaluations.py
@@ -67,7 +67,7 @@ class DatasetItemDict(TypedDict):
     expected_output: ExpectedOutputDict
 
 
-@observe(name="Sync run evaluation on item (JENN)")
+@observe(name="Sync run evaluation on item")
 def sync_run_evaluation_on_item(item: DatasetItemClient) -> AutofixContinuation:
     run_id = None
 
--- a/src/seer/automation/codebase/repo_client.py
+++ b/src/seer/automation/codebase/repo_client.py
@@ -4,6 +4,8 @@ import os
 import shutil
 import tarfile
 import tempfile
+import textwrap
+from concurrent.futures import ThreadPoolExecutor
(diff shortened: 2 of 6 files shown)


src/seer/automation/autofix/evaluations.py
```diff
@@ -67,7 +67,7 @@ class DatasetItemDict(TypedDict):
     expected_output: ExpectedOutputDict
 
 
-@observe(name="Sync run evaluation on item (JENN)")
+@observe(name="Sync run evaluation on item")
```
jennmueng, line 70: oops

src/seer/automation/codebase/repo_client.py
```diff
@@ -378,6 +401,149 @@ def get_valid_file_paths(self, sha: str | None = None, files_only=False) -> set[
...
+                # Determine if this is the last item at this level
+                is_last_item = i == len(items) - 1 or (i == len(items) - 2 and "__status__" in node)
+
+                # Create the appropriate prefix for this line
+                if is_root:
+                    current_prefix = ""
+                    next_prefix = ""
+                else:
+                    current_prefix = prefix + ("└── " if is_last else "├── ")
+                    next_prefix = prefix + ("    " if is_last else "│   ")
```
jennmueng, line 514: I've noticed in my pr reviewer w/ gemini flash these structures work better than a list of files too

jennmueng (approved):

PR #2071, opened 2025-03-04
feat(autofix): Generate summaries for RC and Solution timelines

jennmueng:
Has the formatter step for Root Cause and Summary return a 1 liner summary of the root cause and solution. Places it in an optional `description` field for both steps, and should be backwards compatible.

Does not meaningfully affect eval scores.
Root Cause stable, but helpfulness 0.8969 -> 0.9196
Solution seems stable but is fixed 0.5670 -> 0.5804, score 0.6524 -> 0.6454
Coding dropped, even tho it wasn't touched? conciseness 0.9622  -> 0.9396, correctness 0.9304 -> 0.8983


src/seer/automation/autofix/steps/solution_step.py
```diff
@@ -78,7 +78,7 @@ def _invoke(self):
                 root_cause_and_fix=root_cause_and_fix,
                 event_details=event_details,
                 summary=summary,
-                instruction=state.request.instruction,
+                original_instruction=state.request.instruction,
```
jennmueng, line 81: how long was this bug here lol, my local pylance caught this immediately...we should switch soon...

roaga (approved):

PR #2078, opened 2025-03-05
Anomaly detection/remove dynamic window

aayush-se:


src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -206,64 +206,64 @@ def _online_detect(
...
+        #         use_suss_window = True
+
+        #     anomalies.use_suss.append(use_suss_window)
+
+        # else:
+        #     anomalies.use_suss.append(True)
+        anomalies.use_suss.append(True)
+
+        num_anomlies = len(streamed_anomalies.flags)
+        streamed_anomalies_online = alert_data_accessor.combine_anomalies(
```
ram-senth, line 257: minor - why did we rename this from streamed_anomalies to streamed_anomalies_online?

ram-senth (approved):
Just one minor comment.

ram-senth:
One test is failing.

src/seer/anomaly_detection/anomaly_detection.py
aayush-se: This was because mypy threw an error when the `MPTimeSeriesAnomaliesSingleWindow` object was reassigned to a `MPTimeSeriesAnomalies` so had to assign the former to 'streamed_anomalies' and the latter to 'streamed_anomalies_online'

PR #2098, opened 2025-03-06
feat(autofix): Coding follows a solution from UI

jennmueng:
Le coding agent will follow the solution passed from the UI, and we also store which is active or not

Key change is on the update call, the frontend can pass a solution to be set

Custom solution is removed on frontend, keeping it here for backwards compat

### DIFF
--- a/src/seer/automation/autofix/components/solution/models.py
+++ b/src/seer/automation/autofix/components/solution/models.py
@@ -1,6 +1,6 @@
 from typing import Literal
 
-from pydantic import BaseModel, ConfigDict
+from pydantic import BaseModel, ConfigDict, model_validator
 
 from seer.automation.agent.models import Message
 from seer.automation.autofix.components.root_cause.models import RootCauseAnalysisItem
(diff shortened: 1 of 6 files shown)


src/seer/automation/autofix/components/solution/models.py
```diff
@@ -16,13 +16,21 @@ class RelevantCodeFile(BaseModel):
 
 class SolutionTimelineEvent(BaseModel):
     title: str
-    code_snippet_and_analysis: str
-    relevant_code_file: RelevantCodeFile | None
+    code_snippet_and_analysis: str | None = None
+    relevant_code_file: RelevantCodeFile | None = None
```
jennmueng, line 20: human instruction will only have title

src/seer/automation/autofix/components/solution/models.py
```diff
@@ -16,13 +16,21 @@ class RelevantCodeFile(BaseModel):
...
     is_most_important_event: bool = False
-    event_type: Literal["internal_code"] | str = "internal_code"
+    timeline_item_type: Literal["internal_code", "human_instruction"] = "internal_code"
```
jennmueng, line 22: it was `timeline_item_type` on the frontend, and we were sending event_type, butit works cuz everything was internal code anyways lol

roaga (approved):

PR #2100, opened 2025-03-06
feat: Store Bot Comment Embeddings from Closed PRs

RulaKhaled:
- [x] Add tests

This PR adds functionality to collect and store bot comment patterns when PRs are closed. We analyze user reactions to identify good/bad patterns and store them with embeddings for future similarity matching.

# Implementation Details
1. PR Closed Handler
```
class PrClosedStep(CodegenStep):
    """Processes closed PRs to collect comment patterns"""
```

- Triggers when a PR is closed
- Fetches all review comments from the PR
- Filters for bot comments only (codecov-ai-reviewer[bot])

2. Comment Analysis

- Counts 👍 and 👎 reactions on each comment
- Determines if it's a good pattern (more 👍) or bad pattern (more 👎)

3. Embedding Generation
```
# Generate embeddings for similarity matching
model = GoogleProviderEmbeddings.model(
    "text-embedding-005", 
    task_type="CODE_RETRIEVAL_QUERY"
)
embedding = model.encode(comment.body)[0]  # 768-dimensional vector
```
4. Database Storage
```
class DbReviewCommentEmbedding(Base):
    """Store PR review comments with their embeddings for pattern matching per organization"""
    __tablename__ = "review_comments_embedding"
```

- Stores comment text, metadata, and embeddings
- Uses pgvector for efficient similarity search
- Handles duplicates with on_conflict_do_nothing

# Example Flow

1. PR #123 is closed
2. System finds bot comment
3. Comment has 3 👍 and 1 👎 → marked as good pattern
4. Generates embedding for similarity matching
5. Stores in DB with metadata (PR link, file location, etc.)


src/seer/automation/codegen/pr_closed_step.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Handles comment analysis logic
+    """
+
+    def __init__(self, bot_username: str = "codecov-ai-reviewer[bot]"):
+        self.bot_username = bot_username
+
+    def is_bot_comment(self, comment: PullRequestComment) -> bool:
+        """Check if comment is authored by bot"""
+        return comment.user.login == self.bot_username
```
jennmueng, line 38: is it possible to use user ids here instead of usernames in case we ever change the bot name?

src/seer/automation/codegen/pr_closed_step.py
```diff
@@ -0,0 +1,154 @@
...
+            is_good_pattern, is_bad_pattern = self.analyzer.analyze_reactions(comment)
+
+            logger.info(
+                f"Processing bot comment id {comment.id} on PR {pr.url}: "
+                f"good_pattern={is_good_pattern}, "
+                f"bad_pattern={is_bad_pattern}"
+            )
+
+            model = GoogleProviderEmbeddings.model(
+                "text-embedding-005", task_type="CODE_RETRIEVAL_QUERY"
```
jennmueng, line 94: [SEMANTIC_SIMILARITY](https://cloud.google.com/vertex-ai/generative-ai/docs/embeddings/task-types#assess_text_similarity) is probably a better task type for this use case given we're comparing review comments. The `CODE_RETRIEVAL_QUERY` looks to be for code chunk retrieval and you need to store what you're retrieving as `RETRIEVAL_DOCUMENT`

src/seer/automation/codebase/repo_client.py
```diff
@@ -122,11 +122,31 @@ def get_codecov_pr_review_app_credentials(
     return app_id, private_key
 
 
+@inject
+def get_codecov_pr_closed_app_credentials(
```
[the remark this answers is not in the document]
jennmueng: Yeah, do we really need a second bot just to read the PR closed?

src/seer/db.py
```diff
@@ -467,3 +467,35 @@ class DbProphetAlertTimeSeriesHistory(Base):
...
+    __table_args__ = (
+        UniqueConstraint("provider", "pr_id", "repo", "owner"),
+        Index("ix_review_comments_repo_owner_pr", "owner", "repo", "pr_id"),
+        Index(
+            "ix_review_comments_embedding_hnsw",
+            "embedding",
+            postgresql_using="hnsw",
+            postgresql_with={"m": 16, "ef_construction": 200},
+            postgresql_ops={"embedding": "vector_cosine_ops"},
+        ),
```
jennmueng, line 499: Assuming we'll just be comparing this on a per-repo basis, we probably don't need the hnsw index; if we're expecting during query time to compare against <1000 rows then we don't need it.

src/seer/automation/codegen/tasks.py
```diff
@@ -99,6 +114,25 @@ def codegen_unittest(request: CodegenBaseRequest, app_config: AppConfig = inject
...
+
+    pr_closed_request = PrClosedStepRequest(
+        run_id=cur_state.run_id,
+        pr_id=request.pr_id,
+        repo_definition=request.repo,
+    )
+
+    PrClosedStep.get_signature(
+        pr_closed_request, queue=app_config.CELERY_WORKER_QUEUE
+    ).apply_async()
```
jennmueng, line 131: This should have a very short runtime right, do we need to make it an async celery task? If we just do it sync, whatever's calling this can handle erroring out appropriately

src/seer/automation/codebase/repo_client.py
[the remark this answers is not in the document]
RulaKhaled: i agree! i assumed it's a codebase pattern since we have different fns for unit_tests and pr_review steps

src/seer/automation/codegen/pr_closed_step.py
RulaKhaled: for sure

src/seer/db.py
RulaKhaled: i mainly added HNSW index because it provides a standardized way to perform vector similarity searches. But to your point:

>  if we're expecting during query time to compare against <1000 rows then we don't need it

i think it's good to future-proof the app if some repos accumulate more comments over time (and tbh the index overhead is negligible). But anyway you would def know the codebase better, so let me know if you still encourage removing it!

src/seer/automation/codegen/pr_closed_step.py
RulaKhaled: definitely, thanks!

src/seer/db.py
jennmueng: Yep, we can keep it, it's not a major problem

jennmueng (approved):

PR #2105, opened 2025-03-06
feat(anomaly detection): MP Override

aayush-se:
- Adjusts the scoring algorithm to use the MP Flags if there is strong confidence (mp_dist > 2 * threshold) even if prophet does not indicate an anomaly. This is an incremental change.

Before MP Override:
Screenshot 2025-03-05 at 9 31 41 PM

After MP Override:
Screenshot 2025-03-05 at 9 54 36 PM

If we look at the end of the big spike in this example, because the Prophet CI begins to widen after, originally we stop the alert earlier than we should (not good in stream case where we are not sure how the timeseries will behave next). By adding the override, we extend the anomaly just a little more and ensure we are fully confident before marking it as resolved. 

Overall in the other examples, only 1 of them produced a 1 timestep false positive while the rest remained the same as intended.

**CAVEAT:**
An example like this still needs to be dug deeper into because this spike at Nov 15th should be called out, but the matrix profile does not go high enough for it to trigger and Prophet scores are not high enough for it to override either due to the confidence intervals being so large. On `high` sensitivity though this should be captured due to confidence intervals shrinking.
image

### DIFF
--- a/src/seer/anomaly_detection/accessors.py
+++ b/src/seer/anomaly_detection/accessors.py
@@ -13,6 +13,7 @@ from sqlalchemy.dialects.postgresql import insert
 
 from seer.anomaly_detection.models import (
     AlgoConfig,
+    ConfidenceLevel,
     DynamicAlert,
     MPTimeSeries,
     MPTimeSeriesAnomalies,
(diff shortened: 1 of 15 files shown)


ram-senth:
Can we add a enumeration for the new confidence_level values instead of passing around strings?

jennmueng (approved):

jennmueng:
oops like an idiot i hit approve on the wrong tab LOL ignore me

ram-senth (approved):

PR #2125, opened 2025-03-08
chore(anomaly_detection): Enable prophet in combo detection

ram-senth:


src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -296,12 +310,38 @@ def _online_detect(
...
 
+    def _adjust_time_budget_for_combo_detection(
+        self, time_budget_ms: int | None, num_rows: int, num_rows_per_day: int
+    ):
+        if time_budget_ms is not None:
+            if num_rows > num_rows_per_day:
+                time_budget_ms = time_budget_ms // (num_rows // num_rows_per_day)
+            else:
+                time_budget_ms = time_budget_ms // 2
+        return time_budget_ms
```
aayush-se, line 321: Should this default to a valid if `time_budget_ms` is None? I don't see a check elsewhere in the PR to handle for a None value.

aayush-se (approved):
Looks good. Just one minor comment

src/seer/anomaly_detection/anomaly_detection.py
ram-senth: As we discussed offline, this PR does not modify pre-existing logic of actually budgeting time. So the null check is not part of it. However, there is gaps in how timebudget is being used with the new prophet+mp logic that needs revisiting. I will add anther ticket and followup on it separately.

PR #2128, opened 2025-03-10
fix(relevant-warnings) Read PR instead of commit

kddubey:
[context in slack](https://sentry.slack.com/archives/C08B43VDQVD/p1741312266930079)

### DIFF
--- a/src/seer/automation/codegen/models.py
+++ b/src/seer/automation/codegen/models.py
@@ -150,11 +150,19 @@ class CodegenRelevantWarningsStateResponse(BaseModel):
 
 
 class PrFile(BaseModel):
-    # https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28
     filename: str
     patch: str
     status: Literal["added", "removed", "modified", "renamed", "copied", "changed", "unchanged"]
     changes: int
+    sha: str
+
+    def __hash__(self) -> int:
+        return hash((self.filename, self.sha))
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, PrFile):
+            return False
+        return self.filename == other.filename and self.sha == other.sha
 
 
 class FilterWarningsRequest(BaseComponentRequest):
--- a/src/seer/automation/codegen/models.py
+++ b/src/seer/automation/codegen/models.py
@@ -156,14 +156,6 @@ class PrFile(BaseModel):
     changes: int
     sha: str
 
-    def __hash__(self) -> int:
-        return hash((self.filename, self.sha))
-
-    def __eq__(self, other: object) -> bool:
-        if not isinstance(other, PrFile):
-            return False
-        return self.filename == other.filename and self.sha == other.sha
-
 
 class FilterWarningsRequest(BaseComponentRequest):
     warnings: list[StaticAnalysisWarning]
--- a/src/seer/automation/codegen/relevant_warnings_component.py
+++ b/src/seer/automation/codegen/relevant_warnings_component.py
@@ -110,7 +110,7 @@ class FilterWarningsComponent(BaseComponent[FilterWarningsRequest, FilterWarning
 def _fetch_issues_for_pr_file_cache_key(
     organization_id: int, provider: str, external_id: str, pr_file: PrFile, *args
 ) -> tuple[str]:
-    return hashkey(organization_id, provider, external_id, pr_file)
+    return hashkey(organization_id, provider, external_id, pr_file.filename, pr_file.sha)
 
 
 @cached(cache=LRUCache(maxsize=4096), key=_fetch_issues_for_pr_file_cache_key)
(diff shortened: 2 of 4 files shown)


src/seer/automation/codegen/models.py
```diff
@@ -150,11 +150,11 @@ class CodegenRelevantWarningsStateResponse(BaseModel):
 
 
 class PrFile(BaseModel):
-    # https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28
     filename: str
     patch: str
     status: Literal["added", "removed", "modified", "renamed", "copied", "changed", "unchanged"]
     changes: int
+    sha: str
```
kddubey, line 157: this is indeed the blob SHA

src/seer/automation/codegen/relevant_warnings_step.py
```diff
@@ -117,12 +117,16 @@ def _invoke(self, **kwargs):
...
-        commit = repo_client.repo.get_commit(self.request.commit_sha)
+        pr_files = repo_client.repo.get_pull(self.request.pr_id).get_files()
         pr_files = [
             PrFile(
-                filename=file.filename, patch=file.patch, status=file.status, changes=file.changes
+                filename=file.filename,
+                patch=file.patch,
+                status=file.status,
+                changes=file.changes,
+                sha=file.sha,
```
kddubey, line 127: this is indeed the blob SHA

jennmueng: unintuitive lol

jennmueng (approved):

src/seer/automation/codegen/relevant_warnings_component.py
```diff
@@ -109,14 +107,48 @@ def invoke(self, request: FilterWarningsRequest) -> FilterWarningsOutput:
...
+        run_id=run_id,
+    )
+    if pr_filename_to_issues is None:
+        logger.exception(
+            "Something went wrong with the issue-fetching RPC call",
+            extra={"file": pr_file.filename},
+        )
+        return []
+    assert list(pr_filename_to_issues.keys()) == [pr_file.filename]
+    return list(pr_filename_to_issues.values())[0]
```
suejung-sentry, line 142: [note from suejung - IGNORE ME - I am testing out some formats for a comment]

⚠️ Potential crash in `_fetch_issues_for_pr_file`: the code assumes the RPC response always includes the expected filename and at least one issue. If not, it could crash in two ways:  
- The `assert` might fail if the filename isn’t in the response.  
- Accessing `list(...)[0]` will crash if the list is empty.

Suggest replacing the `assert` with a safe conditional and checking for an empty list before accessing elements.

<details>
<summary>Why this could crash</summary>

This function relies on an RPC call to return a mapping like `{filename: [list of issues]}`.

However, if the RPC doesn’t return the expected filename (e.g. the file has no issues or the RPC skips it), then:

1. The line  
   ```python
   assert list(pr_filename_to_issues.keys()) == [pr_file.filename]
   ```  
   will raise an `AssertionError`.

2. Even if the dict is present but empty (e.g. `{}`), the next line  
   ```python
   return list(pr_filename_to_issues.values())[0]
   ```  
   will raise an `IndexError` because it's trying to access the first element of an empty list.

Although the RPC client handles some failure cases by returning `None`, it doesn’t protect against these more subtle edge cases. The backend service sometimes omits filenames entirely if there are no issues found, and the current code doesn’t account for that.

**Bottom line:** this logic assumes more than the RPC guarantees. Let’s make it defensive.

</details>

suejung-sentry (commented):
[note from suejung - IGNORE ME - I am testing out some formats for a comment]

## Potential Errors Detected

### Item 1

⚠️ Potential crash in `_fetch_issues_for_pr_file`  
The code assumes the RPC response always includes the expected filename and at least one issue. If not, it could crash in two ways:

- The `assert` might fail if the filename isn’t in the response.  
- Accessing `list(...)[0]` will crash if the list is empty.

**🛠 Suggestion:** Replace the `assert` with a safe conditional, and check that the list is non-empty before accessing elements.

<details>
<summary>Why this could crash</summary>

This function relies on an RPC call to return a mapping like `{filename: [list of issues]}`.

However, if the RPC doesn’t return the expected filename (e.g. the file has no issues or the RPC skips it), then:

1. The line  
   ```python
   assert list(pr_filename_to_issues.keys()) == [pr_file.filename]
will raise an AssertionError.

Even if the dict is present but empty (e.g. {}), the next line
```
return list(pr_filename_to_issues.values())[0]
```
will raise an IndexError because it's trying to access the first element of an empty list.

Although the RPC client handles some failure cases by returning None, it doesn’t protect against these more subtle edge cases. The backend service sometimes omits filenames entirely if there are no issues found, and the current code doesn’t account for that.

Bottom line:
This logic assumes more than the RPC guarantees. Let’s make it defensive to avoid crashes on missing or empty responses.

</details>

PR #2140, opened 2025-03-11
feat(workflows): Seer Endpoint for Comparison workflows

Mitan:
Adds a new endpoint /v1/workflows/compare-cohorts that returns the Comparison workflows results based on the [contract](https://www.notion.so/sentry/Seer-Endpoint-API-contract-1ab8b10e4b5d80889a86ca296f8d30c8)


src/seer/workflows/common/constants.py
```diff
@@ -0,0 +1,14 @@
+# Constants used across the workflow
+
+# Attribute used to represent missing values
+EMPTY_VALUE_ATTRIBUTE = ""
```
ram-senth, line 4: Would recommend using a config class with these fields. Makes it more configurable and also it will be easier to test the logic with different settings.

src/seer/workflows/compare/models.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+# Input models
+class StatsAttributeBucket(BaseModel):
+    label: str
+    value: float
+
+
+class StatsAttribute(BaseModel):
+    attributeName: str
```
ram-senth, line 20: Recommend staying consistent with camel case naming convention and avoid snake case.

src/seer/workflows/compare/processor.py
```diff
@@ -0,0 +1,137 @@
...
+
+import pandas as pd
+
+from seer.workflows.common.constants import DEFAULT_ALPHA, EMPTY_VALUE_ATTRIBUTE
+from seer.workflows.compare.models import CompareCohortsRequest, StatsCohort
+from seer.workflows.exceptions import DataProcessingError
+
+
+@dataclass
+class DataProcessor:
```
ram-senth, line 11: It seems like DataProcessor class is more for processing data than actually storing it. Do we need the `@dataclass` annotation? Also, if you decide to add a new config class then this class should be initialized with instance of the config class. If you feel adventurous then you can also look at our dependency injection for injecting the config.

src/seer/workflows/compare/models.py
```diff
@@ -0,0 +1,55 @@
...
+    baseline: StatsCohort
+    selection: StatsCohort
+    options: Options | None = None
+
+
+# Output models
+class AttributeResult(BaseModel):
+    attributeName: str
+    attributeValues: List[str]
+    attributeScore: float
```
roaga, line 50: I read the contract and was a bit confused by the response structure. What is the meaning behind the list of values returned here vs the ones in the request? Isn't it also a little confusing to conflate "label" and "value" for the same value?

Similar question, what's the score?

src/seer/workflows/compare/processor.py
```diff
@@ -0,0 +1,137 @@
...
+    Processes cohort data by normalizing and transforming attribute distributions.
+
+    Attributes:
+        empty_value_attribute (str): Label used for missing values in distributions
+        alpha (float): Smoothing parameter for Laplace smoothing
+    """
+
+    empty_value_attribute: str = EMPTY_VALUE_ATTRIBUTE
+    alpha: float = DEFAULT_ALPHA
+
```
roaga, line 22: do you think these constants should be defined in a single file or defined in the classes/files that use them?

src/seer/workflows/compare/processor.py
```diff
@@ -0,0 +1,137 @@
...
+                            item.label: item.value / data.total_count for item in attr.buckets
+                        },
+                    }
+                    for attr in data.attributeDistributions.attributes
+                ]
+            )
+            df["distribution"] = df["distribution"].apply(lambda x: self.add_unseen_value(x))
+            return df
+        except Exception as e:
+            raise DataProcessingError(f"Failed to preprocess cohort data: {str(e)}") from e
```
roaga, line 50: this try except is probably a good idea in practice, but what kinds of exceptions are you expecting here? It's just some math

src/seer/workflows/compare/processor.py
```diff
@@ -0,0 +1,137 @@
...
+        try:
+            # reindex distribution to include all keys, filling missing values with 0
+            distribution = distribution.reindex(all_keys, fill_value=0)
+
+            # perform lapalce smoothing of the distribution
+            # Add alpha to all values and renormalize
+            distribution = distribution + self.alpha
+            return dict(distribution / distribution.sum())
+        except Exception as e:
+            raise DataProcessingError(f"Failed to transform distribution: {str(e)}") from e
```
roaga, line 94: same question here about the try except

src/seer/workflows/compare/service.py
```diff
@@ -0,0 +1,96 @@
...
+    This service orchestrates the process of comparing two cohorts by:
+    1. Processing and normalizing their attribute distributions
+    2. Computing comparison metrics (KL divergence and entropy)
+    3. Ranking attributes by their differences
+
+    The service uses dependency injection for its processor and scorer components,
+    allowing for flexible configuration and easier testing.
+    """
+
+    _instance = None  # Class-level instance for singleton pattern
```
roaga, line 24: why is a singleton necessary? Initialization doesn't seem that heavy to me if I understand it correctly

ram-senth (commented):
Have added a few suggestions around the coding conventions. Will look at the logic and may add some more comments later.

src/seer/workflows/compare/service.py
```diff
@@ -0,0 +1,96 @@
...
+                  ranked by their RRF scores
+
+        Process:
+            1. Ensures options are set (uses defaults if none provided)
+            2. Processes cohort data to normalize distributions
+            3. Computes comparison metrics and ranks attributes
+            4. Extracts top K attributes and their most distinctive values
+        """
+        if request.options is None:
+            request.options = Options()
```
roaga, line 62: would just using a default value in the request model be cleaner?

src/seer/workflows/compare/processor.py
```diff
@@ -0,0 +1,137 @@
...
+                - attribute_name: Name of the attribute
+                - distribution: Dictionary mapping labels to normalized values
+        """
+        try:
+            df = pd.DataFrame(
+                [
+                    {
+                        "attribute_name": attr.attributeName,
+                        "distribution": {
+                            item.label: item.value / data.total_count for item in attr.buckets
```
aayush-se, line 41: Is there potential for nan/null values to occur here? If so then should handle it here or elsewhere in the pre-processor

src/seer/workflows/compare/processor.py
```diff
@@ -0,0 +1,137 @@
...
+            - Adds alpha to all values and renormalizes to maintain sum = 1
+        """
+        try:
+            # reindex distribution to include all keys, filling missing values with 0
+            distribution = distribution.reindex(all_keys, fill_value=0)
+
+            # perform lapalce smoothing of the distribution
+            # Add alpha to all values and renormalize
+            distribution = distribution + self.alpha
+            return dict(distribution / distribution.sum())
```
aayush-se, line 92: similar concern for nan/null values here if for some reason all possible keys are missing and filled with 0 values

src/seer/workflows/compare/processor.py
Mitan: by right, the pydantic model wouldn't allow it to be null
https://github.com/getsentry/seer/blob/3435c14cac7e92c495df805861de94d6c93e5aad/src/seer/workflows/compare/models.py#L14

src/seer/workflows/compare/models.py
Mitan: The list of values in request is the full set of values, unordered (for example, all possible browsers)
The list of values in response are top top_K suspicious browsers, ordered. 
Label here is the attribute value like "Chrome", and value is it's count. Is the naming confusing? I might change it if it is.
The scores are the "suspiciousness" score of the attribute, requested by the user to do some thresholding logic in the UI

src/seer/workflows/compare/processor.py
Mitan: I might move them into a config class, like Ram suggested

src/seer/workflows/compare/models.py
roaga: Yeah I think the naming was a bit confusing. For example, in the request, "Chrome" is a "label", but in the result, it's a "value". If you could capture the explanation you gave here in all the names and/or in comments, that would probably be helpful.

src/seer/workflows/compare/processor.py
Mitan: since it's pandas, it can throw some weird shit :)
what is sop in the team then? just do one big try catch? or don't catch anything and resolve uncaught exceptions?

src/seer/workflows/compare/service.py
Mitan: it was more the future use if we are planning to add some pre-computing, but you are right, for now i can remove it

src/seer/workflows/common/constants.py
Mitan: thanks, good point!

src/seer/workflows/compare/service.py
Mitan: good point, thanks!

src/seer/workflows/compare/models.py
Mitan: is there any preference within the team for either? saw both cases appear in the codebase 👀

src/seer/workflows/compare/processor.py
Mitan: does this address your comment? or do you mean smth else?
https://github.com/getsentry/seer/blob/3435c14cac7e92c495df805861de94d6c93e5aad/src/seer/workflows/compare/models.py#L14-L16

src/seer/workflows/compare/processor.py
Mitan: thanks, let me think about that!

src/seer/workflows/compare/models.py
Mitan: addressed in 13d4e069a128a7ac62b6a2d5541974cc00712a8f

src/seer/workflows/compare/service.py
Mitan: addressed in 1434edd7c20164d0bf277bd0efe88d911e37e8d5

src/seer/workflows/compare/service.py
Mitan: addressed in d11e6e1d18b4efe7d4a89a7bb79439b7a01d5858

src/seer/workflows/common/constants.py
Mitan: adressed in af3c557295df5aa98b2fe3816f6f40ebf6256884
I made the config part of the request with optional params.
This way the whole flow is stateless and we can support the configs changes from multiple use-cases at the same time. Pls let me know your thoughts.

src/seer/workflows/compare/processor.py
Mitan: See comment above

src/seer/workflows/compare/processor.py
Mitan: I would like to think a bit more about it, so will create a separate ticket for this

tests/workflows/test_processor.py
```diff
@@ -0,0 +1,130 @@
...
+    result = processor.prepareCohortsData(request)
+
+    # Check the structure of the result
+    assert "attributeName" in result.columns
+    assert "distributionBaseline" in result.columns
+    assert "distributionSelection" in result.columns
+
+    # Check that distributions are properly normalized
+    assert all(sum(d.values()) == pytest.approx(1.0) for d in result["distributionBaseline"])
+    assert all(sum(d.values()) == pytest.approx(1.0) for d in result["distributionSelection"])
```
kddubey, line 130: prolly good to add a test for the values of the probabilities for each `attributeValue`

kddubey (commented):
Big TY for docstrings!

Mostly minor feedback. Marked potentially important comments w/ (**)

src/seer/workflows/compare/processor.py
```diff
@@ -0,0 +1,130 @@
...
+        baseline = self._preprocessCohort(request.baseline, config)
+        selection = self._preprocessCohort(request.selection, config)
+
+        dataset = baseline.merge(
+            selection, on="attributeName", how="inner", suffixes=("Baseline", "Selection")
+        )
+        # identify common keys which appear in both distributions
+        dataset["commonKeys"] = dataset.apply(
+            lambda row: set(row["distributionBaseline"].keys())
+            | set(row["distributionSelection"].keys()),
```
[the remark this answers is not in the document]
Mitan: here | is actually intended, because these are the values of individual attribute, and if some new value appears only in the outlier (like "unknown" broswer), it can be valuable info, so we want to keep it and take union of keys instead of intersection

Thanks,  i actually didn't know that apply is slower, but here it shouldn't make much difference since the number of attributes shouldn't be too big

src/seer/workflows/compare/processor.py
```diff
@@ -0,0 +1,130 @@
...
+            3. Identifies common keys across distributions
+            4. Applies Laplace smoothing to both distributions
+            5. Cleans up intermediate calculation columns
+        """
+        config = request.config
+        baseline = self._preprocessCohort(request.baseline, config)
+        selection = self._preprocessCohort(request.selection, config)
+
+        dataset = baseline.merge(
+            selection, on="attributeName", how="inner", suffixes=("Baseline", "Selection")
```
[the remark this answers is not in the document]
Mitan: The approach is different for attribute names and attribute values. If the whole attribute is missing in one cohort (like there is no browser at all in the selection), we drop it. And that's what inner merge is doing.

But if there are some individual values which appear only in one cohort (like there is chrome only in baseline, but not in selection), we keep such values and do the union over attributes

src/seer/workflows/compare/scorer.py
```diff
@@ -0,0 +1,144 @@
...
+class CohortsMetricsScorer:
+    """
+    Scores cohort comparisons using multiple metrics including KL divergence and entropy.
+
+    This class implements a reciprocal rank fusion (RRF) approach to combine multiple
+    scoring metrics into a final ranking score.
+    """
+
+    @staticmethod
+    def _klMetricLambda(baseline: pd.Series, selection: pd.Series) -> pd.Series:
```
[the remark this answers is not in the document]
Mitan: thanks, yes, it is! interesting that mypy doesn't catch it

src/seer/workflows/compare/scorer.py
```diff
@@ -0,0 +1,144 @@
...
+            baseline: Probability distribution of the baseline cohort
+            selection: Probability distribution of the selection cohort
+
+        Returns:
+            Series containing the KL divergence values for each attribute value
+
+        Notes:
+            Uses relative entropy to measure how selection distribution differs from baseline
+        """
+        return rel_entr(pd.Series(baseline), pd.Series(selection))
```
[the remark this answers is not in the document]
Mitan: at these points the set of keys is the same for both baseline and selection due to this
https://github.com/getsentry/seer/blob/28c20ae1bc0c07e7ceb9d05ab523917c482d9489/src/seer/workflows/compare/processor.py#L115-L118

And looks like it's smart enough to do the matching!

src/seer/workflows/compare/processor.py
```diff
@@ -0,0 +1,130 @@
...
+                            for item in attr.buckets
+                        },
+                    }
+                    for attr in data.attributeDistributions.attributes
+                ]
+            )
+            df["distribution"] = df["distribution"].apply(lambda x: self._addUnseenValue(x, config))
+            return df
+        except Exception as e:
+            raise DataProcessingError(f"Failed to preprocess cohort data: {str(e)}") from e
```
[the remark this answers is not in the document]
Mitan: thanks, I am planning to have a separate ticket to figure out exception handling based on Rohan's comment #2159

tests/workflows/test_processor.py
Mitan: thanks, edited the test

src/seer/workflows/compare/service.py
```diff
@@ -0,0 +1,78 @@
...
+        scoredDataset = self.scorer.computeMetrics(dataset, request.config)
+        results = [
+            {
+                "attributeName": row["attributeName"],
+                "attributeValues": list(row["distributionSelection"].keys())[
+                    : request.config.topKBuckets
+                ],
+                "attributeScore": row["rrfScore"],
+            }
+            for _, row in scoredDataset.head(request.config.topKAttributes).iterrows()
```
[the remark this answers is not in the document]
Mitan: that's an interesting point, i tried to test it locally, and it seems to work as expected, but to be safe i added reseting of index, as an alternative safety measure here:
https://github.com/getsentry/seer/blob/af783fa6d04fdc6cc6a31254bb5bc7c625065612/src/seer/workflows/compare/scorer.py#L142

kddubey (approved):
great work! 

last small thing would be to try use snake-case when you can / outside the pydantic models which interact w/ sentry since I assume those use camel case. also, sentry has [some utilities](https://github.com/getsentry/sentry/blob/master/src/sentry/api/serializers/rest_framework/base.py) ([example](https://github.com/getsentry/sentry/blob/3aeac32b994616872218adf1bdc2587cda8e14b6/src/sentry/api/endpoints/group_ai_summary.py#L210)) for converting b/t cases if you wanna use those on sentry's side

src/seer/workflows/compare/processor.py
```diff
@@ -0,0 +1,130 @@
...
+            - Adds alpha to all values and renormalizes to maintain sum = 1
+        """
+        try:
+            # reindex distribution to include all keys, filling missing values with 0
+            distribution = distribution.reindex(allKeys, fill_value=0)
+
+            # perform lapalce smoothing of the distribution
+            # Add alpha to all values and renormalize
+            distribution = distribution + config.alphaLaplace
+            return dict(distribution / distribution.sum())
```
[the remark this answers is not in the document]
ram-senth: @kddubey wondering, what made you decide `distribution` here probabilities? From what I see in the code, it seems to be the counts.

src/seer/workflows/compare/models.py
ram-senth: @Mitan my bad, should have responded earlier. Think majority of our code uses snake case. See that you have adopted camel case. I'm fine with either one, we just need to standardize on one I guess. Worth asking this q in our channel to see if others have any preference.

src/seer/workflows/compare/processor.py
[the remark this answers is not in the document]
kddubey: `_preprocessCohort` is called before `_transformDistribution` in `prepareCohortsData`. docstring also says `distribution: Series containing the probability distribution`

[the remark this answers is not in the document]
ram-senth: My bad, missed that call to _preprocessCohort. Thanks.

PR #2145, opened 2025-03-12
chore(ci): try overwatch without libssl

giovanni-guidini:
Technically on https://github.com/codecov/overwatch/commit/ee056fa94dbc219331ae64103f38fc61e56c6dfa we shouldn't have to explicitly download OpenSSL anymore.

The overwatch CLI should be able to use a rust library that, hopefully, goes with it.

I've seen it work locally and for the overwatch project itself, but I'm not yet confident the requirement can be dropped. So I'm testing it out :E

### DIFF
--- a/.github/workflows/relevant-warnings.yml
+++ b/.github/workflows/relevant-warnings.yml
@@ -30,12 +30,6 @@ jobs:
           pip install mypy==1.15.0
           pip install ruff==0.3.3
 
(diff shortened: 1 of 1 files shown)


giovanni-guidini:
well I had my suspicions and they seemed to have been correct.
Thanks for lending the CI 🙏

PR #2146, opened 2025-03-12
fix(autofix): Stronger autocorrect for repo names

roaga:
If there is only one available repo, we autocorrect to that. Should alleviate some of our parsing errors in the coding step.

Also add some more errors to retry on in the LLM client.

### DIFF
--- a/src/seer/automation/agent/client.py
+++ b/src/seer/automation/agent/client.py
@@ -455,6 +455,7 @@ class AnthropicProvider:
         retryable_errors = (
             "overloaded_error",
             "Internal server error",
+            "not_found_error",
         )
         return (
             isinstance(exception, anthropic.AnthropicError)
@@ -839,6 +840,7 @@ class GeminiProvider:
             # https://sentry.sentry.io/issues/6301072208
             "TLS/SSL connection has been closed",
             "Max retries exceeded with url",
+            "Internal error",
         )
         return (
             isinstance(exception, ClientError)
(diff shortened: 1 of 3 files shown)


src/seer/automation/agent/client.py
```diff
@@ -455,6 +455,7 @@ def is_completion_exception_retryable(exception: Exception) -> bool:
         retryable_errors = (
             "overloaded_error",
             "Internal server error",
+            "not_found_error",
```
jennmueng, line 458: can you shoot me the Sentry issue this one is for?

src/seer/automation/autofix/autofix_context.py
```diff
@@ -131,6 +131,8 @@ def get_repo_client(
 
     def autocorrect_repo_name(self, repo_name: str) -> str | None:
         repo_names = [repo.full_name for repo in self.repos]
+        if len(repo_names) == 1:
+            return repo_names[0]
```
jennmueng, line 135: Hmm, would this be an issue for a case such as when the LLM is actually searching for the `getsentry` repo when we only have the `sentry` repo, should the correct feedback loop for the LLM be to just error and say we don't have access to it rather than autocorrecting to `sentry`?

src/seer/automation/agent/client.py
roaga: https://sentry.sentry.io/issues/6369703164

src/seer/automation/autofix/autofix_context.py
roaga: Fair point, changed my approach. Trying to fix https://sentry.sentry.io/issues/6363508439

src/seer/automation/agent/client.py
jennmueng: bruh what kind of error is that lmfao but ok

jennmueng (approved):

PR #2153, opened 2025-03-12
feat(issue_summary): Score fixability via issue summary

jennmueng:
Adds a new endpoint `/v1/automation/summarize/fixability` that returns the fixability score of an issue summary. This is a function that's called separately in anticipation for fixability being handled async, for now it's called in sync for people with the feature flag

We store these scores and the versioning etc on the root db model for ease of migrations


src/seer/automation/summarize/models.py
```diff
@@ -14,9 +14,16 @@ class SummarizeIssueRequest(BaseModel):
...
+
+
 class SummarizeIssueScores(BaseModel):
-    possible_cause_confidence: float
-    possible_cause_novelty: float
+    possible_cause_confidence: float | None = None
+    possible_cause_novelty: float | None = None
+    fixability_score: float | None = None
+    fixability_score_version: int | None = None
+    is_fixable: bool | None = None
```
roaga, line 26: the pattern for the other scores is that we don't define the threshold or "make decisions" in Seer, but rather just give the raw score and let the user (sentry frontend or backend?) decide what to do with it

should we follow that same pattern here?

src/seer/automation/utils.py
```diff
@@ -300,3 +301,10 @@ def batch_texts_by_token_count(
     if batch:
         # The last batch didn't hit max_tokens. It needs to be yielded.
         yield batch
+
+
+def cosine_similarity(embeddings_a, embeddings_b):
```
roaga, line 306: type hints would be nice, but oh well i don't care enough

roaga: also maybe this should be in `embeddings.py`? idk

tests/test_seer.py
```diff
@@ -511,6 +516,60 @@ def test_summarize_issue_endpoint_internal_error(self, mock_run_summarize_issue)
         assert response.status_code == 500  # InternalServerError
         mock_run_summarize_issue.assert_called_once_with(test_data)
 
+    @mock.patch("seer.app.run_fixability_score")
```
roaga, line 519: mocks are so much nicer than vcrs

(only partial saracasm)

src/seer/automation/summarize/models.py
```diff
@@ -14,9 +14,16 @@ class SummarizeIssueRequest(BaseModel):
...
+    group_id: int
+
+
 class SummarizeIssueScores(BaseModel):
-    possible_cause_confidence: float
-    possible_cause_novelty: float
+    possible_cause_confidence: float | None = None
+    possible_cause_novelty: float | None = None
+    fixability_score: float | None = None
+    fixability_score_version: int | None = None
```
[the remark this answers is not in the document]
jennmueng: Expecting to change the score and threshold often, so having a version just makes it easy to distinguish & also easy to run migrations on

src/seer/automation/summarize/issue.py
```diff
@@ -159,13 +184,62 @@ def run_summarize_issue(request: SummarizeIssueRequest):
...
+
+
+@observe(name="Get Fixability Score")
+def run_fixability_score(request: GetFixabilityScoreRequest) -> SummarizeIssueResponse:
+    with Session() as session:
+        db_state = session.get(DbIssueSummary, request.group_id)
+        if not db_state:
+            raise ValueError(f"No issue summary found for group_id: {request.group_id}")
+        issue_summary = IssueSummaryWithScores.from_db_state(db_state)
+
```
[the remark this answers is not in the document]
jennmueng: yep but we don't need to use this there if we don't need the scores, but ill update it anyways

roaga:
Also when you say it's called synchronously, you'll just be calling this new endpoint from `sentry` backend right? I don't think I saw it being called here directly

jennmueng:
> Also when you say it's called synchronously, you'll just be calling this new endpoint from `sentry` backend right? I don't think I saw it being called here directly

https://github.com/getsentry/sentry/pull/87050

roaga (approved):

PR #2160, opened 2025-03-13
feat(autofix): New testing flow

roaga:
Always adds a step to add a unit test at the end of the solution timeline, but disables it by default. Adds support for triggering "testing mode" if that event is present in the solution sent in the payload. I tested it and it follows instructions nicely.

### DIFF
--- a/src/seer/automation/autofix/event_manager.py
+++ b/src/seer/automation/autofix/event_manager.py
@@ -193,7 +193,7 @@ class AutofixEventManager:
                 )
                 for solution_step in solution_output.solution_steps
             ]
-            solution_step.solution.append(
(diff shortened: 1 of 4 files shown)


src/seer/automation/autofix/event_manager.py
```diff
@@ -193,6 +193,16 @@ def send_solution_result(self, solution_output: SolutionOutput):
                 )
                 for solution_step in solution_output.solution_steps
             ]
+            solution_step.solution.append(  # type: ignore[union-attr]
+                SolutionTimelineEvent(
```
roaga, line 197: mypy needs to go

jennmueng (approved):

PR #2162, opened 2025-03-14
bugfix(anomaly detection): fix for combined detection call from UI timing out

ram-senth:
This PR includes:
* reducing the number of days use stream prediction to 3 days
* changes the number of prophet predictions to cover all 3 days to reduce the number of times we run Prophet prediction
* removes trace on methods that are called for each point in the dataset to avoid huge traces


aayush-se (approved):
looks good

PR #2165, opened 2025-03-14
:alien: add code snippet to StaticAnalysisWarning

giovanni-guidini:
From https://github.com/codecov/overwatch/pull/239 we will be able to have code snippets for the warnings. The hope is that they will help provide better suggestions from the LLMs.

Obviously depends on that PR being merged.
I did spent more time than I expected fixing the formatting, not sure if it was all that worth it. Do LLMs even care?

### DIFF
--- a/src/seer/automation/codebase/models.py
+++ b/src/seer/automation/codebase/models.py
@@ -128,19 +128,40 @@ class StaticAnalysisWarning(BaseModel):
     encoded_location: str
     rule_id: int | None = None
     rule: StaticAnalysisRule | None = None
+    code_snippet: str | None = None
     # TODO: project info necessary for seer?
+
+    def _try_get_language(self) -> str | None:
+        if ".py" in self.encoded_location:
+            return "python"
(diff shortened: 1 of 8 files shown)


src/seer/automation/codebase/models.py
```diff
@@ -128,19 +128,40 @@ class StaticAnalysisWarning(BaseModel):
     encoded_location: str
     rule_id: int | None = None
     rule: StaticAnalysisRule | None = None
+    code_snippet: str | None = None
```
suejung-sentry, line 131: We had this field on the `StaticAnalysisWarning` model already called `encoded_code_snippet`, could we use that key? Technically not much is encoded about it, true

giovanni-guidini: Ah good point. If in the request the field name is `encoded_code_snippet` it could break parsing of the request.

Thanks for pointing it out, I'll use the same name.

suejung-sentry (approved):
thanks for adding this improvement!

PR #2166, opened 2025-03-14
chore(anomaly detection): too many instances of test is run in parallel during deployment, skipping test for now

ram-senth:
Seems like there is some issue with how test parallelization is setup. One test is getting executed 100+ times in parallel and this seems to make the tests run very slow, resulting in a timeout.

### DIFF
--- a/tests/seer/anomaly_detection/test_anomaly_detection.py
+++ b/tests/seer/anomaly_detection/test_anomaly_detection.py
@@ -535,42 +535,43 @@ class TestAnomalyDetection(unittest.TestCase):
             assert len(response.timeseries) == n
             assert isinstance(response.timeseries[0], TimeSeriesPoint)
 
-    def test_detect_anomalies_combo_large_current(self):
-        config = AnomalyDetectionConfig(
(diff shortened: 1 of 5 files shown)


tests/seer/anomaly_detection/test_anomaly_detection.py
```diff
@@ -535,42 +535,43 @@ def test_detect_anomalies_combo(self):
...
+    #     context = TimeSeriesWithHistory(history=ts_history, current=ts_current)
+
+    #     request = DetectAnomaliesRequest(
+    #         organization_id=1, project_id=1, config=config, context=context
+    #     )
+
+    #     response = AnomalyDetection().detect_anomalies(request=request, time_budget_ms=5000)
+
+    #     assert isinstance(response, DetectAnomaliesResponse)
+    #     assert isinstance(response.timeseries, list)
```
aayush-se, line 571: I think you can just pass in a `@pytest.mark.skip(reason="<reason>")` to mark it rather than commenting it out.
https://stackoverflow.com/questions/38442897/how-do-i-disable-a-test-using-pytest

aayush-se (approved):

PR #2175, opened 2025-03-14
fix(autofix): Return existing PR + logging

jennmueng:
After looking at the trace of a lot of the errors, [a lot of them are coming from the non-draft pr line](https://sentry.sentry.io/issues/6361334163/events/f3e7f38368fa449aba774b4fa940d9fa/?project=6178942&sort=date&statsPeriod=14d). And apart from that the errors don't have a stacktrace, this attempts to better logging there so we can debug what the actual issue is.

Also, if a PR already exists for a branch, we should just return it and link to it on Autofix cuz u can only have 1 PR for a branch...

### DIFF
--- a/src/seer/automation/codebase/repo_client.py
+++ b/src/seer/automation/codebase/repo_client.py
@@ -21,8 +21,8 @@ from github import (
     UnknownObjectException,
 )
 from github.GitRef import GitRef
-from github.Repository import Repository
 from github.PullRequest import PullRequest
+from github.Repository import Repository
 
 from seer.automation.autofix.utils import generate_random_string, sanitize_branch_name
 from seer.automation.codebase.models import GithubPrReviewComment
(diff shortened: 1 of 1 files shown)


src/seer/automation/codebase/repo_client.py
```diff
@@ -690,7 +704,7 @@ def create_pr_from_branch(
                 draft=True,
             )
         except GithubException as e:
-            if e.status == 422:
+            if e.status == 422 and "Draft pull requests are not supported" in str(e):
```
roaga, line 707: is that the actual error message?

roaga (approved):

src/seer/automation/codebase/repo_client.py
jennmueng: yes from the issue

PR #2176, opened 2025-03-15
fix(autofix) Constrain repo name in semantic search

kddubey:
noticed we [`autocorrect_repo_name`](https://github.com/getsentry/seer/blob/ecca1fd00d46cfdd668bc55b71614921c692fa80/src/seer/automation/autofix/autofix_context.py#L132) but structured output accurately works w/ dynamically created `Literal`s (have used this successfully for classification tasks, [like this one](https://github.com/kddubey/blog/blob/main/llm_api/batch/banking77.ipynb))

also adds a VCR test for semantic search


roaga (approved):

trillville (commented):
@kddubey should we merge this?

PR #2189, opened 2025-03-17
chore(anomaly detection): store datapoint even if we do not have enough history for AD

ram-senth:


src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -205,6 +257,28 @@ def _online_detect(
...
+                    confidence_levels=[],
+                ),
+                algo_data=None,
+                historic=historic,
+                config=config,
+                alert_data_accessor=alert_data_accessor,
+                queue_cleanup_predict_task=(
+                    len(historic.timeseries.timestamps) == min_data - 1
+                ),  # we queue cleanup task if we have just enough data to detect anomalies in the next call
+                force_cleanup=True,
```
aayush-se, line 280: Why are we attempting to cleanup on every stream call?

src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -205,6 +257,28 @@ def _online_detect(
...
+            self._save_and_fire_pruning_task(
+                timepoint=ts_external[0],
+                streamed_anomalies=MPTimeSeriesAnomaliesSingleWindow(
+                    flags=["none"] * len(ts_external),
+                    scores=[0.0] * len(ts_external),
+                    thresholds=[],
+                    matrix_profile=np.array([]),
+                    window_size=3,
+                    original_flags=[],
+                    confidence_levels=[],
```
aayush-se, line 271: It's okay that this is a dummy anomaly right? My understanding is that in the future when there are enough timepoints, then the anomaly data will be overwritten to the proper values.

aayush-se (approved):

PR #2191, opened 2025-03-17
fix(relevant-warnings): Filter warnings by pr diff instead of by files

suejung-sentry:
Filter warnings based on those that have a line number in the modified file line numbers in the PR diff. Previously we allowed the warning to exist anywhere in the same file, which turned out to be too loose.

Closes https://github.com/codecov/overwatch/issues/242

### DIFF
--- a/.github/workflows/relevant-warnings.yml
+++ b/.github/workflows/relevant-warnings.yml
@@ -30,12 +30,6 @@ jobs:
           pip install mypy==1.15.0
           pip install ruff==0.3.3
 
-      - name: Install Build Dependencies
-        run: |
-          sudo apt-get update
-          wget http://archive.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1f-1ubuntu2_amd64.deb
-          sudo dpkg -i libssl1.1_1.1.1f-1ubuntu2_amd64.deb
-
       - name: Install Overwatch CLI
         run: |
           curl -o overwatch-cli https://overwatch.codecov.io/linux/cli
--- a/src/seer/automation/codegen/relevant_warnings_component.py
+++ b/src/seer/automation/codegen/relevant_warnings_component.py
@@ -1,4 +1,5 @@
 import logging
+import re
 import textwrap
 from pathlib import Path
 from typing import Any
@@ -36,7 +37,7 @@ from seer.rpc import RpcClient
 
 class FilterWarningsComponent(BaseComponent[FilterWarningsRequest, FilterWarningsOutput]):
     """
-    Filter out warnings from files that aren't affected by the PR.
+    Filter out warnings that aren't on the PR diff lines.
     """
 
     context: CodegenContext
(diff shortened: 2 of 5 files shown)


src/seer/automation/codegen/relevant_warnings_component.py
```diff
@@ -65,9 +66,64 @@ def _left_truncated_paths(path: Path, max_num_paths: int = 2) -> list[str]:
...
+        - q: number of lines in modified file
+
+        For example, given this diff:
+        @@ -1,3 +1,4 @@
+        def hello():
+            print("hello")
+        +    print("world")  # Line 3 is added
+        print("goodbye")
+
+        This would return [1,2,3,4] since these are all the lines in the updated file
```
kddubey, line 86: big TY for this docstring!

src/seer/automation/codegen/relevant_warnings_component.py
```diff
@@ -65,9 +66,64 @@ def _left_truncated_paths(path: Path, max_num_paths: int = 2) -> list[str]:
...
+
+        Args:
+            pr_file: PrFile object containing the patch/diff
+
+        Returns:
+            List of 1-indexed line numbers in the updated file
+        """
+        patch_lines = pr_file.patch.split("\n")
+        changed_lines: list[int] = []
```
kddubey, line 95: nit: to bound memory a bit more (since there can be many files, and patches can be arbitrarily large), can change this method to output a list of `range`s (one `range` for each hunk) and later check:

```python
any(warning_line in hunk_range for hunk_range in hunk_ranges)
```

src/seer/automation/codegen/relevant_warnings_component.py
```diff
@@ -79,32 +135,64 @@ def _does_warning_match_a_target_file(
...
-            self.logger.info(f"{warning.encoded_location=} {matches=}")
-        return bool(matches)
+
+        # Make possible variations of the pr files' paths
+        pr_file_by_filepath_variation: dict[str, PrFile] = {}
+        for pr_file in pr_files:
+            pr_path = Path(pr_file.filename)
+            pr_file_by_filepath_variation[pr_path.as_posix()] = pr_file
+            for truncated in self._left_truncated_paths(pr_path, max_num_paths=1):
+                pr_file_by_filepath_variation[truncated] = pr_file
```
kddubey, line 150: nit: looks like this block is independent of the warning. can factor it out and re-use the result. might be good b/c there can be 1000s of warnings

kddubey (approved):
looks great! left performance nits and one clarifying question

src/seer/automation/codegen/relevant_warnings_component.py
kddubey: double checking: IIUC, the hunk can include lines not necessarily changed in the PR? regardless, filtering by the hunk instead of the file is a big-enough improvement

src/seer/automation/codegen/relevant_warnings_component.py
```diff
@@ -79,32 +135,64 @@ def _does_warning_match_a_target_file(
...
+                    "matching_files": [pf.filename for pf in matching_pr_files],
+                },
+            )
+
+        return matching_pr_files
+
+    def _is_warning_line_in_pr_files(
+        self, warning: StaticAnalysisWarning, matching_pr_files: list[PrFile]
+    ) -> bool:
+        # Encoded location format: "file:line:col"
```
giovanni-guidini, line 172: hum... is this accurate? If it is the same as [Location](https://github.com/getsentry/seer/blob/14a08a1ece3960632d2a5f50e30deea78bbe1d30/src/seer/automation/codebase/models.py#L76) that doesn't seem to be the proper formatting

It'd be `file:start_line[~end_line]`

giovanni-guidini (commented):
Small question on the warnings encoded location

.github/workflows/relevant-warnings.yml
```diff
@@ -30,12 +30,6 @@ jobs:
           pip install mypy==1.15.0
           pip install ruff==0.3.3
 
-      - name: Install Build Dependencies
```
suejung-sentry, line 33: work by trent removed this dependency

src/seer/automation/codegen/relevant_warnings_component.py
suejung-sentry: Yes from what i could tell the hunk includes at most 3 lines before and after any change. True that could capture slightly irrelevant things, but agree it's a fine approximation for now

src/seer/automation/codegen/relevant_warnings_component.py
suejung-sentry: Yeah that's true. Made the change to save the ranges and since they're sorted, had it do a binary search for slight optimization.

src/seer/automation/codegen/relevant_warnings_component.py
suejung-sentry: Yeah good point - pulled out the map as suggested!

src/seer/automation/codegen/relevant_warnings_component.py
```diff
@@ -79,32 +135,64 @@ def _does_warning_match_a_target_file(
...
+            pr_path = Path(pr_file.filename)
+            pr_file_by_filepath_variation[pr_path.as_posix()] = pr_file
+            for truncated in self._left_truncated_paths(pr_path, max_num_paths=1):
+                pr_file_by_filepath_variation[truncated] = pr_file
+
+        # Find all matching PR files
+        matching_pr_files: list[PrFile] = []
+        for filepath in warning_filepath_variations:
+            if filepath in pr_file_by_filepath_variation:
+                matching_pr_files.append(pr_file_by_filepath_variation[filepath])
```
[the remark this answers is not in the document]
suejung-sentry: Nice yeah that's better, updated!

src/seer/automation/codegen/relevant_warnings_component.py
suejung-sentry: Oh yeah you're right, fixed!

PR #2198, opened 2025-03-18
feat(autofix): Add trace context to prompts and offer tools

roaga:
- Adds trace tree to root cause and solution prompts
- Adds two tools to fetch profiles, spans, and connected errors 

Both are gated just to @roaga (me) for now.

### DIFF
--- a/src/seer/automation/autofix/tools.py
+++ b/src/seer/automation/autofix/tools.py
@@ -694,8 +694,12 @@ class BaseTools:
             and not run_request.options.disable_interactivity
             and (run_request.invoking_user and run_request.invoking_user.id == 3283725)
         ):  # TODO temporary guard for Rohan (@roaga) to test in prod
-            tools.extend(
-                [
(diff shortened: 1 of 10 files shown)


src/seer/automation/autofix/tools.py
```diff
@@ -395,6 +399,91 @@ def google_search(self, question: str, llm_client: LlmClient = injected):
...
+        # if profile available, fetch it via Seer RPC
+        profile_data = rpc_client.call(
+            "get_profile_details",
+            organization_id=trace_tree.org_id,
+            project_id=event.project_id,
+            profile_id=event.profile_id,
+        )  # expecting data compatible with Profile model
+        if not profile_data:
+            return "Could not fetch profile."
+        profile = Profile.model_validate(profile_data)
```
jennmueng, line 436: smells like potential validation errors will happen here, should we gracefully handle this and say return a string saying an error occurred in this tool?

jennmueng (approved):

PR #2207, opened 2025-03-19
fix(autofix) Fix Gemini retryable class, log exception's module

kddubey:


src/seer/utils.py
```diff
@@ -76,7 +76,7 @@ def closing_queue(*queues: Queue):
 
 
 def exception_formatter(exception: Exception) -> str:
-    return f"{type(exception).__name__}: {exception}"
+    return f"{type(exception).__module__}.{type(exception).__qualname__}: {exception}"
```
kddubey, line 79: so we don't have to guess where the exception came from

roaga (approved):

PR #2234, opened 2025-03-20
feat(autofix): Enable trace-connected context for all users

roaga:
Removes guard and enables trace-connected context in prompts and with tools for all users.

Also loads repo preferences for trace-connected projects.


src/seer/automation/autofix/runs.py
```diff
@@ -18,13 +16,36 @@ def create_initial_autofix_run(request: AutofixRequest) -> DbState[AutofixContin
...
+            cur.request.repos = []
+        for trace_connected_preference in trace_connected_preferences:
+            if trace_connected_preference:
+                for repo in trace_connected_preference.repositories:
+                    if not any(
+                        existing_repo.provider == repo.provider
+                        and existing_repo.owner == repo.owner
+                        and existing_repo.name == repo.name
+                        for existing_repo in cur.request.repos
+                    ):
```
jennmueng, line 47: Would an `external_id` comparison not be better here?

roaga: changed it

jennmueng (approved):

PR #2251, opened 2025-03-22
feat(autofix): Disable shell for command tools, make list directory a tree

roaga:
Disabling shell for the `grep` and `find` tools for security purposes.

Also make `list_directory` return a `tree` format of nested directories.
Root cause correctness went from 84 to 88%.
Solution correctness went from 67% to 68%.

### DIFF
--- a/src/seer/automation/autofix/tools.py
+++ b/src/seer/automation/autofix/tools.py
@@ -1,4 +1,5 @@
 import logging
+import shlex
 import subprocess
 import textwrap
 from concurrent.futures import ThreadPoolExecutor, as_completed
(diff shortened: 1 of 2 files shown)


src/seer/automation/autofix/tools.py
```diff
@@ -157,39 +158,35 @@ def explain_file(self, file_path: str, repo_name: str):
...
-        self.context.event_manager.add_log(f"Looking at contents of `{path}` in `{repo_name}`...")
+        self.context.event_manager.add_log(
+            f"Viewing directory tree for `{path}` in `{repo_name}`..."
+        )
 
-        joined = self._format_list_directory_output(dirs, files)
-        return f"<entries>\n{joined}\n</entries>"
+        # Use the _build_file_tree_string method from the repo client
+        tree_representation = repo_client._build_file_tree_string(files_under_path)
+        return f"<directory tree>\n{tree_representation}\n</directory tree>"
```
jennmueng, line 189:
```suggestion
        return f"<directory_tree>\n{tree_representation}\n</directory_tree>"
```
nit but, we might make the llm think it can return invalid xml later

jennmueng (approved):

PR #2272, opened 2025-03-25
feat(automation): EAP Trace Summary

aayush-se:
- Initial setup for individual trace summaries
- Sets up `POST` endpoint for submitting an EAP Trace Waterfall (json) and receiving the respective summary


src/seer/automation/models.py
```diff
@@ -1023,3 +1024,89 @@ def apply(self, file_contents: str | None) -> str | None:
...
 
         return file_contents.replace(self.reference_snippet, "")
+
+
+class EAPTrace(BaseModel):
+    trace_id: str = Field(..., description="ID of the trace")
+    trace: list[dict] = Field(..., description="List of spans in the trace")
+    timestamp: datetime.datetime = Field(..., description="Known timestamp of a span in the trace")
+
+    def _get_transaction_spans(self, trace) -> list[dict]:
```
roaga, line 1034: should we type hint the param?

src/seer/automation/models.py
```diff
@@ -1023,3 +1024,89 @@ def apply(self, file_contents: str | None) -> str | None:
...
+        """
+        Filters the trace to only include the transaction spans.
+        """
+        # Base case: empty trace
+        if not trace:
+            return []
+
+        # Process list of spans
+        if isinstance(trace, list):
```
roaga, line 1043: would it be cleaner if we didn't overload the same param with two different types

src/seer/automation/models.py
```diff
@@ -1023,3 +1024,89 @@ def apply(self, file_contents: str | None) -> str | None:
...
+
+
+class EAPTrace(BaseModel):
+    trace_id: str = Field(..., description="ID of the trace")
+    trace: list[dict] = Field(..., description="List of spans in the trace")
+    timestamp: datetime.datetime = Field(..., description="Known timestamp of a span in the trace")
+
+    def _get_transaction_spans(self, trace) -> list[dict]:
+        """
+        Filters the trace to only include the transaction spans.
```
roaga, line 1036: are transactions considered spans?

aayush-se: yep, in EAP while everything is a span, transaction spans specifically have an `is_transaction` bool to differentiate

src/seer/automation/models.py
aayush-se: So the reason its overloaded to be either a dict or list is because we are recursively going through the children and the root child can be just a dictionary.

src/seer/automation/models.py
```diff
@@ -1119,3 +1119,24 @@ def format_span_as_tag(span, depth=0):
...
+            repositories=[repo.model_dump() for repo in self.repositories],
+        )
+
+    @classmethod
+    def from_db_model(cls, db_model: DbSeerProjectPreference) -> "SeerProjectPreference":
+        return cls(
+            organization_id=db_model.organization_id,
+            project_id=db_model.project_id,
+            repositories=db_model.repositories,
+        )
```
roaga, line 1142: what's this doing here?

roaga (approved):
approving to unblock

src/seer/automation/models.py
roaga: I remember Jenn added this functionality but I don't think it should be part of your diff in this PR, because that means something in main will change?

PR #2273, opened 2025-03-25
Update PR review prompt

rohitvinnakota-codecov:
Attempt to fix `Error generating pr review for ...: Invalid JSON format inside <comments>`:

Checked Langfuse and saw that the output was not comma separating json responses so explicitly prompt to avoid this.


roaga (approved):
this is fine but maybe structured output would be more reliable

PR #2274, opened 2025-03-25
fix(autofix): Fix duplicate solution msg

jennmueng:
don't have to tell the agent twice lol


roaga:
[This PR is more complete ](https://github.com/getsentry/seer/pull/2277)

PR #2280, opened 2025-03-25
fix(issue summary): Retry issue summary with less context if crosses limit

roaga:
Sometimes we're crossing the input token limit. So when we do, this retries one more time by removing the code context, variable context, and breadcrumbs from trace-connected issues.

### DIFF
--- a/src/seer/automation/summarize/issue.py
+++ b/src/seer/automation/summarize/issue.py
@@ -88,82 +88,98 @@ def summarize_issue(
         else []
     )
 
-    connected_issues_input = ""
-    if connected_event_details:
-        connected_issues = "\n----\n".join(
-            [
-                f"Issue from the same session:\n{event.format_event()}"
-                for _, event in enumerate(connected_event_details)
-            ]
+    def _generate_summary(full_context: bool = True) -> IssueSummaryWithScores | None:
+        connected_issues_input = ""
+        if connected_event_details:
+            connected_issues = "\n----\n".join(
+                [
+                    f"Issue from the same session:\n{event.format_event() if full_context else event.format_event_without_breadcrumbs(include_context=False, include_var_values=False)}"
+                    for _, event in enumerate(connected_event_details)
+                ]
+            )
+            connected_issues_input = f"""
+            Also, we know about some other issues that occurred in the same application session, listed below. The issue above occurred somewhere alongside these:
+            {connected_issues}
+            """
+        else:
+            connected_issues_input = "Issues from the same session: none"
+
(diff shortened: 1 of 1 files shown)


src/seer/automation/summarize/issue.py
```diff
@@ -88,82 +88,98 @@ def summarize_issue(
...
+            )
+
+            issue_summary = completion.parsed
+            return IssueSummaryWithScores(
+                **issue_summary.model_dump(),
+                scores=SummarizeIssueScores(
+                    possible_cause_confidence=issue_summary.possible_cause_confidence_score,
+                    possible_cause_novelty=issue_summary.possible_cause_novelty_score,
+                ),
+            )
```
kddubey, line 166: nit: if we wanna keep the check in the except block loose, should move this part under an else block. that way, if there's a pydantic validation error that happens to contain `token` somewhere, it won't get retried

kddubey (approved):
ty for handling

there's been a spike of [these events](https://sentry.sentry.io/issues/6287562143/events/2acf9bae83e046788d2930c1fcc7c927/?project=6178942&referrer=next-event) in the past 5 days (would be cool to have issue-level anomaly detection btw). [this event on 03/05](https://sentry.sentry.io/issues/6287562143/events/7d8ccb8b293140cc8f962dd76774264f/?project=6178942&referrer=previous-event) (a week after we switched from 4o-mini to flash-lite) says the window is 1M tokens, which [matches the docs](https://cloud.google.com/vertex-ai/generative-ai/docs/gemini-v2#2.0-flash-lite). but the [very next event on 03/21](https://sentry.sentry.io/issues/6287562143/events/2acf9bae83e046788d2930c1fcc7c927/?project=6178942&referrer=next-event) says the window is now 32k?

src/seer/automation/summarize/issue.py
roaga: @kddubey what do you mean by moving it under an else block? Not sure how that will make the retry condition more specific

kddubey: rn the `except` clause will catch errors out of the 3 statements in the `try`. moving the last two statements into an `else` block will make the `except` only catch errors out of `llm_client.generate_structured`

roaga: I'm still not sure about the `else` block 😓 but i changed the try except to only wrap the llm call

kddubey: i was referring to try-except-else. the updated code does the same thing since the `except` either raises or returns, so it's good

PR #2285, opened 2025-03-26
build: Bump sentry-sdk to 2.21.x

cleptric:
Via `make upgrade-package-versions`


src/seer/bootup.py
```diff
@@ -58,8 +58,6 @@ def before_send(event: Event, hint: dict) -> Event | None:
         integrations=[*integrations],
         traces_sampler=traces_sampler,
         profiles_sample_rate=1.0,
-        enable_tracing=True,
```
cleptric, line 61: This option is deprecated and did the same as `traces_sample_rate=1.0`.

PR #2302, opened 2025-03-27
fix(autofix): Fix bugs

roaga:
- add more supported languages
- fix custom root cause instruction
- fix repo name autocorrect to only use supported providers


src/migrations/versions/17f604dbf6c5_.py
```diff
@@ -6,9 +6,6 @@
 
 """
 
-import sqlalchemy as sa
-from alembic import op
-
```
jennmueng, line 11: was this complaining? the migration _may_ not work without it I'm not sure

src/seer/automation/codebase/utils.py
```diff
@@ -9,12 +9,22 @@
 logger = logging.getLogger(__name__)
 
 language_to_extensions = {
```
jennmueng, line 11: This used to be a list of languages and files tree sitter supported...we might not need it anymore if we're just reading the files?

src/migrations/versions/17f604dbf6c5_.py
roaga: it was complaining in precommit

src/seer/automation/codebase/utils.py
roaga: now we've been using it to filter non-code files. one example is not looking at PNGs

jennmueng (approved):

PR #2308, opened 2025-03-27
Use formatted output for PR reviews

rohitvinnakota-codecov:
Following up from https://github.com/getsentry/seer/pull/2273. We are still seeing some flakiness when it comes to review generation. Use `formatted_output` to generate a valid response.


src/seer/automation/codegen/pr_review_coding_component.py
```diff
@@ -47,40 +47,17 @@ def invoke(
...
+                messages=agent.memory,
+                prompt=CodingCodeReviewPrompts.pr_review_formatter_msg(),
+                model=GeminiProvider(model_name="gemini-2.0-flash-001"),
+                response_format=list[CodePrReviewOutput.Comment],
+                run_name="Generate PR review structured",
+                max_tokens=8192,
             )
 
-        return CodePrReviewOutput(comments=results)
+            for comment in formatted_response.parsed:
```
kddubey, line 59: a gotcha is that `formatted_response.parsed` can be `None`. [it's retried twice](https://github.com/getsentry/seer/blob/bf008680a92c3cb62f1be7b2c69a4ff87012305a/src/seer/automation/agent/client.py#L878-L893) but can still end up being `None` if gemini internally couldn't create the structured output

src/seer/automation/codegen/pr_review_coding_component.py
```diff
@@ -47,40 +47,17 @@ def invoke(
...
-                CodePrReviewOutput.Comment(
-                    path=comment_data["path"],
-                    line=comment_data["line"],
-                    body=body,
-                    start_line=comment_data["start_line"],
-                )
+            formatted_response = llm_client.generate_structured(
+                messages=agent.memory,
+                prompt=CodingCodeReviewPrompts.pr_review_formatter_msg(),
+                model=GeminiProvider(model_name="gemini-2.0-flash-001"),
```
kddubey, line 53:
```suggestion
                model=GeminiProvider.model(model_name="gemini-2.0-flash-001"),
```

kddubey (approved):
don't think this touches any relevant warnings code so LGTM

src/seer/automation/codegen/pr_review_coding_component.py
rohitvinnakota-codecov: Does this matter? I tested locally and it worked

kddubey: doesn't matter and up to you. the `.model` instantiation is typically what we use. it'll pull in any default config for a model

PR #2309, opened 2025-03-27
feat(autofix): Use claude tools in coding

jennmueng:
Introduces the new claude coding tools for use by the coding agent.

Some tools such as the `view` tool's path viewing and file viewing are redundant but required for the claude tool to work. We're keeping both implementations because ours is better and evals are doing good.

Also adds the edits programattically as insight cards + allows for rethinks 

93.65% correctness and 98.97% conciseness

CleanShot 2025-03-31 at 15 23 48

NOTE: Frontend needs to go out before this


src/seer/automation/agent/agent.py
```diff
@@ -157,19 +158,25 @@ def call_tool(self, tool_call: ToolCall) -> Message:
...
-            )
-        except Exception as e:
-            raise AgentError() from ValueError(f"Invalid tool arguments: {args}\nException: {e}")
+    def parse_tool_arguments(self, tool: FunctionTool | ClaudeTool, args: str) -> dict:
+        if isinstance(tool, FunctionTool):
+            try:
+                return parse_json_with_keys(
+                    args,
+                    [param["name"] for param in tool.parameters if isinstance(param["name"], str)],
+                )
```
jennmueng, line 173: We might not need this anymore, it was from a time when LLMs outputted incorrect json all the time

src/seer/automation/autofix/tools.py
```diff
@@ -87,7 +95,7 @@ class FileLocation(BaseModel):
     @observe(name="Semantic File Search")
     @ai_track(description="Semantic File Search")
     @inject
-    def semantic_file_search(self, query: str, llm_client: LlmClient = injected):
+    def semantic_file_search(self, query: str, llm_client: LlmClient = injected, **kwargs: Any):
```
jennmueng, line 98: need to pass kwargs to every function here so it doesn't error when passed with the tool call id and current message...

src/seer/automation/agent/client.py
```diff
@@ -607,7 +607,13 @@ def to_message_param(message: Message) -> MessageParam:
             )
 
     @staticmethod
-    def to_tool_dict(tool: FunctionTool) -> ToolParam:
+    def to_tool_dict(tool: FunctionTool | ClaudeTool) -> ToolParam:
+        if isinstance(tool, ClaudeTool):
+            return ToolParam(  # type: ignore
```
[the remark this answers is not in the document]
roaga: another reason i hate anthropic

src/seer/automation/agent/client.py
```diff
@@ -663,7 +669,7 @@ def generate_text_stream(
         prompt: str | None = None,
         messages: list[Message] | None = None,
         system_prompt: str | None = None,
-        tools: list[FunctionTool] | None = None,
+        tools: list[FunctionTool | ClaudeTool] | None = None,
```
roaga, line 672: aaaahhhhhhh

src/seer/automation/agent/client.py
```diff
@@ -1241,7 +1247,7 @@ def generate_text(
                     prompt=prompt,
                     system_prompt=system_prompt,
                     temperature=temperature or default_temperature,
-                    tools=tools,
+                    tools=cast(list[FunctionTool], tools),
```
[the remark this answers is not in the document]
jennmueng: we shouldn't be passing `ClaudeTools` into a non anthropic model, let's add guards before the cast actually

src/seer/automation/autofix/steps/coding_step.py
```diff
@@ -89,7 +91,7 @@ def _invoke(self, app_config: AppConfig = injected):
         if not summary:
             summary = self.context.get_issue_summary()
 
-        coding_output = CodingComponent(self.context).invoke(
+        CodingComponent(self.context).invoke(
```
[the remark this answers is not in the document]
jennmueng: thinking about this, good idea

src/seer/automation/autofix/tools.py
roaga: Is it possible to just change the `call_tool` logic to make sure it's a ClaudeTool and not a normal tool, so that we don't need this/

roaga (approved):
left one comment

src/seer/automation/autofix/tools.py
jennmueng: yes, originally chose this solution to allow the other functions to create insights but i guess doing it that way would be cleaner w/ current implementation

PR #2327, opened 2025-03-31
ref(autofix): Download repos async when initializing tools

roaga:
Downloads the repos asynchronously when initializing `BaseTools`, instead of when calling a specific tool, to save waiting time.

This is completely vibe coded, so please review thoroughly. It worked when I tested it though.


src/seer/automation/autofix/tools.py
```diff
@@ -411,53 +467,74 @@ def grep_search(self, command: str, repo_name: str | None = None):
...
         Args:
             repo_name: If provided, only ensures this specific repo is downloaded.
                       If None, ensures all repos are downloaded.
         """
-        if self.tmp_dir is None:
-            self.tmp_dir = {}
-
-        downloaded_something = False
+        # Check if parallel download has completed
+        if self._download_future is not None and self._download_future.done():
```
jennmueng, line 478: my understanding is, if we get here and the download future is not done, we'll just be downloading it again without cancelling the parallel downloads? should we be waiting for it with [.result(timeout=...)](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.result) and using that instead of starting a new download?

roaga: @jennmueng updated it

jennmueng (approved):

PR #2328, opened 2025-03-31
fix(anomaly detection): Fix location detection logic

aayush-se:
- Previously, if there was a prophet override, then it could bypass the location detection
- Simplify conditions needed for location detection 
- Ensures that the direction dictates the **start** of an alert.
  - ie: if an alert is set to "down" and an alert fires when the value is 0, but then the next point jumps to 100 (above the bounds), we will continue to fire because the alert _began_ below the bounds


src/seer/anomaly_detection/detectors/anomaly_detectors.py
```diff
@@ -284,6 +285,7 @@ def detect(
                     history_values=self.history_values,
                     history_timestamps=self.history_timestamps,
                     history_mp_dist=mp_dist_baseline,
+                    history_flags=self.original_combined_flags,  # type: ignore
```
ram-senth, line 288: Think there is a basic type mismatch here - `original_combined_flags` is `list[AnomalyFlags | None]` and `history_flags` is `list[AnomalyFlags] | None`. Shouldn't we fix that instead of using `# type: ignore`?

aayush-se: Thanks for catching this -- forgot to address that prior to pushing. I agree and will fix it

tests/seer/anomaly_detection/detectors/test_anomaly_scorer.py
```diff
@@ -645,184 +649,113 @@ def test_prophet_high_confidence_anomaly(
...
-            "anomaly_higher_confidence",
-            "anomaly_higher_confidence",
-            "anomaly_higher_confidence",
-            "anomaly_higher_confidence",
-            "anomaly_higher_confidence",
-            "anomaly_higher_confidence",
-            "anomaly_higher_confidence",
-            "anomaly_higher_confidence",
-            "anomaly_higher_confidence",
+            "none",  # prophet_flag: none, prev_flag: none, direction: both, y: 15.0, yhat_lower: 12.0, yhat_upper: 16.0
```
kddubey, line 704: maybe can use a [parametrized test](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html) here? comes w/ some testing niceties

kddubey (approved):

PR #2377, opened 2025-04-04
fix(autofix): Make PR creation task pollable

roaga:
Makes the PR creation task set the Autofix state and not wait for a timeout (the task still has a 60 second timeout). This is so the frontend can poll for it. This approach should be more stable and avoid random connection issues.

### DIFF
--- a/src/seer/automation/autofix/tasks.py
+++ b/src/seer/automation/autofix/tasks.py
@@ -5,7 +5,7 @@ import textwrap
 from typing import Type, cast
 
 import sentry_sdk
-from celery.exceptions import SoftTimeLimitExceeded, TimeoutError
+from celery.exceptions import SoftTimeLimitExceeded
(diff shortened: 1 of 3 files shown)


src/seer/automation/autofix/tasks.py
```diff
@@ -342,6 +321,12 @@ def commit_changes_task(run_id, repo_external_id, make_pr):
     except Exception as e:
         logger.error(f"Error committing changes for run {run_id}: {e}")
         raise
+    finally:
+        with state.update() as cur:
+            cur.status = AutofixStatus.COMPLETED
+            changes_step = cur.changes_step
+            if changes_step:
+                changes_step.status = AutofixStatus.COMPLETED
```
jennmueng, line 329: here too

jennmueng (approved):

PR #2378, opened 2025-04-04
chore(autofix): Prevent auto runs from overwriting existing runs

roaga:
It would be really awkward if you had a manual run overwritten by an automatically-triggered run. Or if multiple auto-runs were happening at once.

The sentry-side check is already there, but it could still be bypassed with the network delay and the snuba querying delays.


kddubey (approved):
nice catch

PR #2381, opened 2025-04-07
fix(langfuse): Count all prompt cache tokens as prompt tokens

jennmueng:
We'll need to update our langfuse instances to v3 to correctly account for cached tokens, which needs us to set up clickhouse.

Instead, this as an interim half-solution is to include all cached tokens as prompt tokens.

This won't fully fix it but at least it gets us in the ballpark of what the scale of what it does cost us vs not including any of those tokens at all.


roaga (approved):
since this is already an interm solution, is it worth adding multipliers to the cache token counts so that the price langfuse calculates will be more accurate?

PR #2392, opened 2025-04-07
feat(autofix): Save explanation if no code changes

roaga:
Saves the last memory message as a `termination_reason` if there were no code changes.

### DIFF
--- a/src/seer/automation/autofix/components/root_cause/models.py
+++ b/src/seer/automation/autofix/components/root_cause/models.py
@@ -1,6 +1,6 @@
 from typing import Literal, Optional
 
-from pydantic import BaseModel
+from pydantic import BaseModel, validator
 
 from seer.automation.agent.models import Message
 from seer.automation.component import BaseComponentOutput, BaseComponentRequest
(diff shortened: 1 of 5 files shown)


src/seer/automation/autofix/event_manager.py
```diff
@@ -129,8 +129,8 @@ def send_root_cause_analysis_result(self, root_cause_output: RootCauseAnalysisOu
                 root_cause_step.causes = root_cause_output.causes
                 cur.status = AutofixStatus.COMPLETED
             else:
-                root_cause_step.status = AutofixStatus.ERROR
-                cur.status = AutofixStatus.ERROR
+                root_cause_step.status = AutofixStatus.COMPLETED
+                cur.status = AutofixStatus.COMPLETED
```
roaga, line 133: i put this as a fix here anyway, but i don't think <NO_ROOT_CAUSES/> is ever triggered anymore due to the context+reasoning approach we have now.

Do we want to support this better? One avenue is checking if the context gatherer says <NO_ROOT_CAUSES/>. Another is prompting the reasoning model that it could give up if it wanted too.

src/seer/automation/autofix/components/root_cause/models.py
```diff
@@ -16,12 +16,18 @@ class RelevantCodeFile(BaseModel):
...
     is_most_important_event: bool
 
+    @validator("timeline_item_type", pre=True, allow_reuse=True)
+    @classmethod
+    def validate_timeline_item_type(cls, v):
+        if v not in ("internal_code", "external_system", "human_action"):
+            # Default to internal_code if the value is not one of the allowed literals
+            return "internal_code"
+        return v
+
```
roaga, line 30: tacking on this thing that we've wanted for a long time

roaga, line 30: tacking on this thing that we've wanted for a long time

roaga: it also brings our solution evals back up around 70%!

roaga: it also brings our evals back up to where they used to be!

jennmueng: probably flake lol

src/seer/automation/autofix/event_manager.py
jennmueng: let's just do what you did here but refactor it to be

```py
root_cause_step.status = AutofixStatus.COMPLETED
cur.status = AutofixStatus.COMPLETED

if root_cause_output.causes:
    root_cause_step.causes = root_cause_output.causes
```

jennmueng (approved):

PR #2398, opened 2025-04-09
change relevant warnings prediction

giovanni-guidini:
Significant changes to the Relevant Warnings Step so that we suggestions are better.

* Don't require warning _and_ issue to exist
* Add diff to the analysis
* Change prompt to _predict_ potential issues

This is a significant change to the relevant warnings step. It moves somewhat away from the warnings as the sole main focus, and kinda equally use warnings and issues as extra context for the LLM to evaluate the PR dif.

It changes several smaller LLM calls (that decide if warnings and issues can be paired logically) by 1 bigger one (that analyses the entire diff). Works well for small diffs and a small number of warnings and issues, but we're not very sure the scaling factor here.

The next steps would be to better limit the size of the diff, or break it up into different chunks and analyze the chunks individually.

### DIFF
--- a/src/seer/automation/codegen/models.py
+++ b/src/seer/automation/codegen/models.py
@@ -30,15 +30,31 @@ class RelevantWarningResult(BaseModel):
 
 
 class StaticAnalysisSuggestion(BaseModel):
-    path: str
-    line: int
-    short_description: str
-    justification: str
-    related_warning_id: str | None = None
-    related_issue_id: str | None = None
-    severity_score: float
-    confidence_score: float
-    missing_evidence: list[str]
+    path: str = Field(description="The path to the file that contains the suggestion.")
+    line: int = Field(description="The line number of the suggestion.")
+    short_description: str = Field(
+        description="A short, fluff-free, information-dense description of the problem. Max 30 words."
+    )
+    justification: str = Field(
+        description="A short, fluff-free, information-dense summary of your analysis for why this is a problem. This justification should be at most 15 words."
+    )
+    related_warning_id: str | None = Field(
+        default=None,
+        description="If this suggestion is based on a warning, include the warning id here. Else use null.",
+    )
+    related_issue_id: str | None = Field(
+        default=None,
+        description="If this suggestion is based on an issue, include the issue id here. Else use null.",
+    )
+    severity_score: float = Field(
+        description="From 0 to 1 how serious is this potential bug? 1 being 'guaranteed exception will happen and not be caught by the code'."
+    )
+    confidence_score: float = Field(
+        description="From 0 to 1 how confident are you that this is a bug? 1 being 'I am 100% confident that this is a bug'. This should be based on the amount of evidence you had to reach your conclusion."
+    )
+    missing_evidence: list[str] = Field(
+        description="A short list of evidence that you did NOT have but would increase your confidence score. At most 5 items. Be very specific."
+    )
 
     def to_overwatch_format(self) -> RelevantWarningResult:
         """
--- a/src/seer/automation/codegen/models.py
+++ b/src/seer/automation/codegen/models.py
@@ -29,10 +29,22 @@ class RelevantWarningResult(BaseModel):
     encoded_location: str
 
 
+class StaticAnalysisSuggestion(BaseModel):
+    path: str
+    line: int
+    short_description: str
+    justification: str
+    related_warning: str | None = None
+    related_issue: str | None = None
+    severity_score: float
+    confidence_score: float
+    missing_evidence: list[str]
+
+
 class CodegenState(BaseModel):
     run_id: int = -1
     file_changes: list[FileChange] = Field(default_factory=list)
-    relevant_warning_results: list[RelevantWarningResult] = Field(default_factory=list)
+    static_analysis_suggestions: list[StaticAnalysisSuggestion] = Field(default_factory=list)
     status: CodegenStatus = CodegenStatus.PENDING
     last_triggered_at: datetime.datetime = Field(default_factory=datetime.datetime.now)
     updated_at: datetime.datetime = Field(default_factory=datetime.datetime.now)
@@ -208,6 +220,16 @@ class CodePredictRelevantWarningsRequest(BaseComponentRequest):
 
 class CodeAreIssuesFixableOutput(BaseComponentOutput):
     are_fixable: list[bool | None]  # None means the issue was not analyzed
+
+
+class CodePredictStaticAnalysisSuggestionsRequest(BaseComponentRequest):
+    warnings: list[StaticAnalysisWarning]
+    fixable_issues: list[IssueDetails]
+    pr_files: list[PrFile]
+
+
+class CodePredictStaticAnalysisSuggestionsOutput(BaseComponentOutput):
+    suggestions: list[StaticAnalysisSuggestion]
 
 
 class CodePredictRelevantWarningsOutput(BaseComponentOutput):
--- a/src/seer/automation/codegen/models.py
+++ b/src/seer/automation/codegen/models.py
@@ -5,7 +5,7 @@ from typing import Literal
 from pydantic import BaseModel, Field
 
 from seer.automation.agent.models import Message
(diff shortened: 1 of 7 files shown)


src/seer/automation/codegen/models.py
```diff
@@ -29,10 +29,45 @@ class RelevantWarningResult(BaseModel):
...
+    def to_overwatch_format(self) -> RelevantWarningResult:
+        """
+        Convert a StaticAnalysisSuggestion to a RelevantWarningResult.
+        This is a temporary format to post to Overwatch, because Overwatch expects a list of
+        RelevantWarningResult objects at this time
+        TODO: update Overwatch and then remove this method
+        """
+        return RelevantWarningResult(
+            warning_id=self.related_warning_id,
+            issue_id=self.related_issue_id,
```
kddubey, line 52: `RelevantWarningResult` assumes these are `int`s, but the passed in values are `str | None`? Maybe I missed something since it seemed to run successfully for you

src/seer/automation/codegen/models.py
```diff
@@ -29,10 +29,45 @@ class RelevantWarningResult(BaseModel):
     encoded_location: str
 
 
+class StaticAnalysisSuggestion(BaseModel):
+    path: str
+    line: int
+    short_description: str
+    justification: str
+    related_warning_id: str | None = None
+    related_issue_id: str | None = None
```
kddubey, line 38: Optional: to guard against LLM hallucinations, and since overwatch needs these to be actual warning/issue IDs (if they're not `None`), we could change the type to a dynamically constructed `Literal` ([example](https://github.com/getsentry/seer/blob/a1967d61fe4bfff253ba2a336beed6af4bfc4f9b/src/seer/automation/autofix/tools.py#L125-L130))

Since it might be an annoying change, I'm all for first seeing how often it's a problem, and then maybe doing it later

src/seer/automation/codegen/prompts.py
```diff
@@ -277,6 +277,58 @@ def format_prompt(formatted_error: str):
...
+                - `line`: The line number of the suggestion.
+                - `short_description`: a short, fluff-free, information-dense description of the problem. Max 30 words.
+                - `justification`: a short, fluff-free, information-dense summary of your analysis for why this is a problem. This justification should be at most 15 words.
+                - `related_warning_id`: If this suggestion is based on a warning, include the warning id here. Else use null.
+                - `related_issue_id`: If this suggestion is based on an issue, include the issue id here. Else use null.
+                - `severity_score`: From 0 to 1 how serious is this potential bug? 1 being "guaranteed exception will happen and not be caught by the code".
+                - `confidence_score`: From 0 to 1 how confident are you that this is a bug? 1 being "I am 100%% confident that this is a bug". This should be based on the amount of evidence you had to reach your conclusion.
+                - `missing_evidence`: A short list of evidence that you did NOT have but would increase your confidence score. At most 5 items. Be very specific.
```
kddubey, line 323: Optional: can instead pass these as `Field(description=...)` in the `StaticAnalysisSuggestion` model. That's b/c, internally, Gemini dumps the model to a JSON schema and generates it accordingly

src/seer/automation/codegen/relevant_warnings_component.py
```diff
@@ -237,7 +245,9 @@ def _fetch_issues_for_pr_file(
         return []
     if not pr_filename_to_issues:
         return []
-    assert list(pr_filename_to_issues.keys()) == [pr_file.filename]
+    assert list(pr_filename_to_issues.keys()) == [
+        pr_file.filename
+    ], f"expected {pr_file.filename} but got {list(pr_filename_to_issues.keys())}"
```
kddubey, line 250: Ah thanks, I assume this was annoying locally

kddubey (approved):
I don't see any major problems, so approving to unblock before Friday

I'm also wondering if there were any noisy results for the "negative" cases in the [test repo](https://github.com/codecov/ovc-python)?

giovanni-guidini:
> Good stuff! I don't see any major/breaking problems, so approving to unblock before Friday
> 
> I'm also wondering if there were any noisy results for the "negative" cases in the [test repo](https://github.com/codecov/ovc-python)? Specifically for cases where issues are fetched but they're not implicated by the code change

Yes, I did notice the system getting quite chatty for the negative cases as well, which is why I opened [this ticket](https://github.com/codecov/overwatch/issues/330). 
This is yet another reason to merge with pr-review, because they have tech in place to improve filters over time.

src/seer/automation/codegen/models.py
giovanni-guidini: oh good point. I never actually pushed to Overwatch locally, this could be a problem. I'll add a test for the conversion as well.

src/seer/automation/codegen/relevant_warnings_step.py
```diff
@@ -75,16 +76,29 @@ def get_task():
...
-        relevant_warnings_output: CodePredictRelevantWarningsOutput,
+        llm_suggestions: CodePredictStaticAnalysisSuggestionsOutput | None,
         config: AppConfig = injected,
     ):
+        if llm_suggestions is None:
+            self.logger.info("No relevant warnings output to post to Overwatch.")
+            return
+
+        with open("/app/.artifacts/relevant_warnings_output.json", "w") as f:
+            json.dump(llm_suggestions.model_dump(), f, indent=4)
```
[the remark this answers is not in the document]
giovanni-guidini: yes, my bad.

src/seer/automation/codegen/prompts.py
```diff
@@ -277,6 +277,58 @@ def format_prompt(formatted_error: str):
...
+            # Guidelines:
+            - Return AT MOST 5 suggestions, and AT MOST 1 suggestion per line of code.
+            - Focus on bugs, performance and security issues.
+            - Do NOT propose issues if the are outside the diff.
+            - ALWAYS include the exact file path and line of the suggestion.
+            - Assign a severity score and confidence score to each suggestion, from 0 to 1.
+                - Severity score: 1 being "guaranteed an _unchaught_ exception will happen and not be caught by the code"; 0.5 being "an exception will happen but it's being caught" OR "an exception may happen depending on inputs"; 0 being "no exception will happen";
+                - Confidence score: 1 being "I am 100%% confident that this is a bug";
+            - Only surface issues that _don't_ have a warning if BOTH severity AND confidence are high.
+            - DO NOT surface issues with low severity. DO NOT surface issues that are not bugs.
```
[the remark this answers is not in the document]
giovanni-guidini: hum I like that idea, yes

src/seer/automation/codegen/models.py
```diff
@@ -29,10 +29,45 @@ class RelevantWarningResult(BaseModel):
...
+        TODO: update Overwatch and then remove this method
+        """
+        return RelevantWarningResult(
+            warning_id=self.related_warning_id,
+            issue_id=self.related_issue_id,
+            # Let's pretend our suggestions are important
+            does_fixing_warning_fix_issue=True,
+            # Combining both metrics means we will only surface suggestions with high confidence
+            # and severity.
+            relevance_probability=self.confidence_score * self.severity_score,
```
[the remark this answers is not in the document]
giovanni-guidini: We might have to do that, indeed.

src/seer/automation/codegen/relevant_warnings_step.py
```diff
@@ -153,7 +171,11 @@ def _invoke(self, **kwargs):
         fetch_issues_output: CodeFetchIssuesOutput = fetch_issues_component.invoke(
             fetch_issues_request
         )
-
+        # Clamp issue to max_num_issues_analyzed
+        all_selected_issues = list(
+            itertools.chain.from_iterable(fetch_issues_output.filename_to_issues.values())
+        )
+        all_selected_issues = all_selected_issues[: self.request.max_num_issues_analyzed]
```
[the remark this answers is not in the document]
giovanni-guidini: I did change that behavior... Cause there might be issues that have no associations but would still influence the analysis, that was the rationale for the change

PR #2399, opened 2025-04-09
fix(autofix): Truncate long grep lines & responses

jennmueng:


roaga (approved):

roaga:
#2426 covers this now

PR #2408, opened 2025-04-10
ref: Use SENTRY_REGION instead of USE_EU_REGION

jennmueng:
`SENTRY_REGION` is already set in all the seers in prod via https://github.com/getsentry/ops/pull/14607

(The AppConfig setup will fail the deploy if this var is not actually set)

### DIFF
--- a/src/seer/automation/agent/client.py
+++ b/src/seer/automation/agent/client.py
@@ -431,7 +431,7 @@ class AnthropicProvider:
     def get_client(app_config: AppConfig = injected) -> anthropic.AnthropicVertex:
         return anthropic.AnthropicVertex(
             project_id=app_config.GOOGLE_CLOUD_PROJECT,
-            region="europe-west1" if app_config.USE_EU_REGION else "us-east5",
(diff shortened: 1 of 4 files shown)


src/seer/automation/agent/client.py
```diff
@@ -431,7 +431,7 @@ class AnthropicProvider:
     def get_client(app_config: AppConfig = injected) -> anthropic.AnthropicVertex:
         return anthropic.AnthropicVertex(
             project_id=app_config.GOOGLE_CLOUD_PROJECT,
-            region="europe-west1" if app_config.USE_EU_REGION else "us-east5",
+            region="europe-west1" if app_config.SENTRY_REGION == "eu" else "us-east5",
```
roaga, line 434: do you know if it will be set to "eu" or "de"?

roaga (approved):

src/seer/automation/agent/client.py
jennmueng: de, bruh this was brain fart

PR #2426, opened 2025-04-15
feat(autofix): Use gemini 2.5 pro as context gatherer

roaga:
Swap the root cause and solution step context gatherers with gemini 2.5 pro preview. Also tweak the prompts slightly to do better with this model.

Also include grep search results truncation.

Maintains evals and saves a lot of cost.


src/seer/automation/agent/models.py
```diff
@@ -122,3 +122,9 @@ class LlmStreamInactivityTimeoutError(LlmStreamTimeoutError):
     """Raised when the LLM stream times out due to inactivity."""
 
     pass
+
+
+class LlmNoCompletionTokensError(Exception):
```
kddubey, line 127: should this be added to `GeminProvider.is_completion_exception_retryable`?

kddubey (approved):

src/seer/automation/agent/models.py
roaga: did I not do that?

kddubey: oh was looking at an old diff, my bad

PR #2440, opened 2025-04-16
feat(automation): Enable global endpoints for claude and gemini

roaga:
Enables global endpoints for claude and gemini when not in the EU and the model is supported.
I could test locally and gemini worked, claude global endpoint is only available in `internal-sentry` so we'll see what happens


src/seer/automation/codegen/relevant_warnings_component.py
```diff
@@ -317,7 +317,7 @@ def _is_issue_fixable(issue: IssueDetails, llm_client: LlmClient = injected) ->
     # LRU-cached by the issue id. The same issue could be analyzed many times if, e.g.,
     # a repo has a set of files which are frequently used to handle and raise exceptions.
     completion = llm_client.generate_structured(
-        model=GeminiProvider.model("gemini-2.0-flash-lite"),
+        model=GeminiProvider.model("gemini-2.0-flash-lite-001"),
```
roaga, line 320: flash lite required a pinned version to work on the global endpoint

src/seer/automation/agent/client.py
```diff
@@ -427,14 +427,35 @@ class AnthropicProvider:
         ),
     ]
 
-    @staticmethod
```
kddubey, line 430: I think the retry stream test assumes it's a static method. So may need to update [this line](https://github.com/getsentry/seer/blob/83c43fc80f984686096eeef543b39b3d5a87a6d5/tests/automation/autofix/test_retry_stream.py#L99)

roaga: yep fixed

kddubey (approved):

PR #2445, opened 2025-04-17
feat(autofix): New semantic search tool

roaga:
The old semantic search tool was pretty hit-or-miss, and completely broke on big repos. Replacing it with a small gemini flash agent that can use `tree` and `expand_document` to find the exact answer to the query. In some examples on `sentry` it could find most things within 10 seconds.

Bumps solution score to ~61%, cuts latency and cost further from last eval (though that could be the broken evals).

### DIFF
--- a/src/seer/automation/autofix/components/coding/component.py
+++ b/src/seer/automation/autofix/components/coding/component.py
@@ -14,7 +14,7 @@ from seer.automation.autofix.components.coding.models import CodingOutput, Codin
 from seer.automation.autofix.components.coding.prompts import CodingPrompts
 from seer.automation.autofix.components.root_cause.models import RootCauseAnalysisItem
 from seer.automation.autofix.prompts import format_repo_prompt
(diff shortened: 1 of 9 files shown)


kddubey (approved):
oo, nice

since it just has the `repo_str` to work w/ at the start, are there any concerns around cold starts? looking at [this trace](https://langfuse-zelomd34wa-uc.a.run.app/project/clws5ue5q00006gkjq1d9c7l5/traces/9d3c86ff-1054-4283-ab17-131d0d8b2d15?timestamp=2025-04-17T17%3A04%3A00.281Z&observation=c5bb3c2d-63c8-426d-9fa9-1bc40e423532), the agent understandably starts by calling `tree(".")` (and then it looks like `.` isn't supported by the tree tool)

might save a couple of LLM calls at the cost of prompt tokens if, e.g., it starts w/ a level=3 directory tree for each repo

```bash
tree -d -L 3
```

PR #2446, opened 2025-04-18
fix(severity): Log active thread count

kddubey:
The shallow cause of latency spikes is that PyTorch is running [extremely slowly](https://sentry.sentry.io/explore/profiling/profile/seer/fa6dc97bae6a40d6aed892203b428632/flamegraph/?colorCoding=by%20system%20vs%20application%20frame&eventId=07e32f7afe0d418e897f5be2de6c2145&fov=0%2C0%2C14750144512%2C25&query=&sorting=call%20order&tid=138965653628032&view=top%20down). There are some differences b/t the threads that are running—specifically seems like threads 6-8 are missing compared to [healthy profiles](https://sentry.sentry.io/explore/profiling/profile/seer/a8624b62c331417aa529571054e58228/flamegraph/?colorCoding=by%20system%20vs%20application%20frame&eventId=035ba87a8e69464681f7d708b108cc2c&fov=0%2C7%2C24029970%2C25&query=&sorting=call%20order&tid=139708520080512&view=top%20down). I can locally reproduce a profile like the one we see in prod via [this notebook](https://github.com/getsentry/ml-models/blob/main/issue_severity/threads.ipynb).


aayush-se (approved):

aayush-se:
logs make sense so approving to unblock.


had a few questions about the profiles since i dont really understand them too well. I noticed in the health profile, there are far fewer samples (2 vs 1112). Is this just dropped samples or something else?

Also, in the screenshot below (slow left thread 5 healthy right thread 6), there there are some more frames with the ingestion consumer. Is the Does the lack of the ingestion consumer frames mean anything? 

Screenshot 2025-04-18 at 12 35 06 PM

aayush-se:
Also comparing main threads, looks theres a lot more work being done in the main thread of the unhealthy one? Again could be due to missing samples or more threads in the healthy one?

Screenshot 2025-04-18 at 12 38 20 PM

kddubey:
the difference in frames should be caused by the slowness itself [based on the docs](https://docs.sentry.io/product/explore/profiling/performance-overhead/)

in the healthy profile, pytorch is running fast enough that profiling doesn't have a chance to capture every function getting called

aayush-se:
> in the healthy profile, pytorch is running fast enough that profiling doesn't have a chance to capture every function getting called. in the unhealthy profile, we can see the granular pytorch functions take tens of milliseconds, which means pytorch is running super duper slowly

Oh nice thanks for explaining! Good find

PR #2454, opened 2025-04-21
fix(autofix): Parallelize grep search

roaga:
If we run a bad grep query that times out when multiple repos are available, it seems like the run freezes (and often it'll time out).
e.g. 45 second timeout * 4 selected repos = 3 min of doing nothing

So this PR parallelizes grep calls if there are multiple repos so that at most we wait only 45 seconds.

### DIFF
--- a/src/seer/automation/autofix/tools/grep_search.py
+++ b/src/seer/automation/autofix/tools/grep_search.py
@@ -1,0 +1,140 @@
+import logging
+import subprocess
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+from seer.dependency_injection import copy_modules_initializer
+
+logger = logging.getLogger(__name__)
+
+# Constants moved from tools.py
+GREP_TIMEOUT_SECONDS = 10
+MAX_GREP_LINE_CHARACTER_LENGTH = 1000
+TOTAL_GREP_RESULTS_CHARACTER_LENGTH = 20000
+
+
+def _run_grep_in_repo(
+    repo_name: str,
+    cmd_args: list[str],
+    tmp_dir: dict[str, tuple[str, str]],
+    tmp_repo_dir: str,
+) -> str | None:
+    """Runs the grep command in a specific repository directory."""
+    try:
+        # Run the grep command in the repo directory
+        try:
+            process = subprocess.run(
+                cmd_args,
+                shell=False,
+                cwd=tmp_repo_dir,
+                capture_output=True,
+                text=True,
+                check=False,
+                timeout=GREP_TIMEOUT_SECONDS,
+            )
+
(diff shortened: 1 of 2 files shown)


src/seer/automation/autofix/tools/tools.py
```diff
@@ -28,12 +28,14 @@
 from seer.langfuse import append_langfuse_observation_metadata
 from seer.rpc import RpcClient
 
+from .grep_search import run_grep_search
```
jennmueng, line 31: we don't use relative imports anywhere else in seer right, should do an absolute import here

src/seer/automation/autofix/tools/tools.py
```diff
@@ -31,9 +32,9 @@
 logger = logging.getLogger(__name__)
 
 MAX_FILES_IN_TREE = 100
-GREP_TIMEOUT_SECONDS = 45
-MAX_GREP_LINE_CHARACTER_LENGTH = 1024
-TOTAL_GREP_RESULTS_CHARACTER_LENGTH = 16384
+GREP_TIMEOUT_SECONDS = 10
+MAX_GREP_LINE_CHARACTER_LENGTH = 1000
+TOTAL_GREP_RESULTS_CHARACTER_LENGTH = 20000
```
jennmueng, line 37:
```suggestion
```

src/seer/automation/autofix/tools/grep_search.py
```diff
@@ -0,0 +1,140 @@
...
+# Constants moved from tools.py
+GREP_TIMEOUT_SECONDS = 10
+MAX_GREP_LINE_CHARACTER_LENGTH = 1000
+TOTAL_GREP_RESULTS_CHARACTER_LENGTH = 20000
+
+
+def _run_grep_in_repo(
+    repo_name: str,
+    cmd_args: list[str],
+    tmp_dir: dict[str, tuple[str, str]],
```
jennmueng, line 18: `tmp_dir` isn't actually used here right

src/seer/automation/autofix/tools/grep_search.py
```diff
@@ -0,0 +1,140 @@
+import logging
+import subprocess
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
+from seer.dependency_injection import copy_modules_initializer
+
+logger = logging.getLogger(__name__)
+
+# Constants moved from tools.py
+GREP_TIMEOUT_SECONDS = 10
```
jennmueng, line 10: hmm why 10s now, i've seen a lot of valid calls on the large sentry repo take up to 30s

roaga: bruh cursor changed the constants

jennmueng (approved):

PR #2480, opened 2025-04-23
fix(autofix): don't add blank user messages

roaga:
Whitespace-only user messages can break Anthropic API


src/seer/automation/agent/agent.py
```diff
@@ -117,8 +117,9 @@ def run(self, run_config: RunConfig):
         elif self.name:
             langfuse_context.update_current_observation(name=self.name + " - Agent Run")
 
-        if run_config.prompt:
-            self.add_user_message(run_config.prompt)
+        prompt = run_config.prompt
+        if prompt and prompt.strip():
+            self.add_user_message(prompt)
```
jennmueng, line 122: Should we just be immediately raising an error here if there is a prompt passed but it's blank rather than ignoring it?

jennmueng (commented):
Hmm why were we having blank user messages in the first place?

src/seer/automation/agent/agent.py
roaga: done

jennmueng (approved):

PR #2486, opened 2025-04-24
feat(seer): enable sentry logging

trillville:


roaga (approved):

roaga (changes requested):
can we use the constraints system for the requirements (see readme)

trillville:
> can we use the constraints system for the requirements (see readme)

yeah already noticed, sec

roaga (approved):

PR #2494, opened 2025-04-25
WIP - relevant warnings evaluations

giovanni-guidini:
These changes are basically scaffolding for evaluations (using Langfuse) for the RelevantWarningsStep.

* Creates the endpoint to trigger the evaluation run;
* Creates a `datasets.py` file with helper functions to generate a dataset and its items.
* Code to execute and score evaluations is in `evaluations.py`
* `models.py` are typedefs of useful things
* `tasks.py` is the glue that powers the evals.

Not sure if we save a dump of the datasets too, so it's easy to recreate? Probably in some other repo, if we want to do that.


src/seer/automation/codegen/evals/datasets.py
```diff
@@ -0,0 +1,169 @@
...
+            sha=file.sha,
+        )
+        for file in pr_files
+        if file.patch
+    ]
+
+    if issues is None:
+        issues = []
+        sentry_token = os.getenv("SENTRY_TOKEN")
+        assert sentry_token, "SENTRY_TOKEN is not set"
```
suejung-sentry, line 107: I've lost track here how we expect this `SENTRY_TOKEN` to get injected into the environment - also wouldn't each org have a different expected token?

src/seer/app.py
```diff
@@ -331,6 +333,13 @@ def codegen_relevant_warnings_endpoint(
     return codegen_relevant_warnings(data)
 
 
+@json_api(blueprint, "/v1/automation/codegen/relevant-warnings/evaluation/start")
```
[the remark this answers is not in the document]
giovanni-guidini: we expect to call this manually in dev mode. It's similar to the endpoint for autofix evals.

We don't expect to use this in prod at all (at least that's my understanding of how evals are being setup right now)

src/seer/automation/codegen/evals/datasets.py
giovanni-guidini: Sentry token is optional envvar from the README instruction. Obviously this means that the issues you can add to the eval are limited by the issues the sentry token you use can access.

The same sentry token can typically access multiple orgs. Your personal token will be able to access issues from all teams where you can see issues in the sentry dashboard. This makes the "addressable issue list" roughly issues in codecov and sentry repos.

src/seer/app.py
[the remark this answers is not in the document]
suejung-sentry: yeah totally makes sense - sorry I missed that!

src/seer/automation/codegen/evals/datasets.py
suejung-sentry: Great, got it - thanks! I had missed that we are only running this from local as well, so against our test orgs (sentry, codecov). That makes sense. Thanks!

src/seer/automation/codegen/evals/datasets.py
```diff
@@ -0,0 +1,255 @@
...
+    dataset_name: str
+    metadata: dict
+    dataset_items: list[DatasetItem]
+    org_definitions: dict[str, OrgDetails]
+
+
+class ItemRawData(TypedDict):
+    github_info: GitHubInfo
+    warnings: list[StaticAnalysisWarning]
+    expected_suggestions: list[StaticAnalysisSuggestion]
```
JerrySentry, line 75: I think we should have the expected output be a paragraph of natural language instead of coming up with what we think is the correct value for each field in `StaticAnalysisSuggestion`. Then have the LLM in the scoring function to compare if the actual `StaticAnalysisSuggestion` is similar to the description we provided in the expected output.

giovanni-guidini: I like that by doing this we have less coupling from the evaluation to the final suggestion format, which is nice. Especially given that we will probably change it again when merging pr review

At the same time I feel that we should have a deeper understanding of what everything needs and the specific values we'd like the system to create.

So, risking overthinking this, I will make the change but in the long term I think the expected output should exactly match the output of the system

src/seer/automation/codegen/evals/tasks.py
```diff
@@ -0,0 +1,139 @@
...
+                trace_id=dataset_item_trace_id,
+                name="error_running_evaluation",
+                value=1,
+            )
+            suggestions = None
+
+    # If suggestions is None we assume no suggestions were generated.
+    # rather than an error happening.
+    # TODO: Is this the best way to handle this?
+    suggestions = suggestions or []
```
kddubey, line 123: in prod this kinda makes sense haha

though for evals i think it makes more sense to either not upload a score for this item, or upload a score of 0. i think langfuse doesn't like missing scores so for now it's easier to upload a 0

src/seer/automation/codegen/evals/tasks.py
```diff
@@ -0,0 +1,139 @@
...
+    dataset_item_trace_id = None
+    with dataset_item.observe(run_name=run_name, run_description=run_description) as trace_id:
+        dataset_item_trace_id = trace_id
+        try:
+            suggestions = sync_run_evaluation_on_item(dataset_item, langfuse_session_id=trace_id)  # type: ignore
+            langfuse.score(
+                trace_id=dataset_item_trace_id,
+                name="error_running_evaluation",
+                value=0,
+            )
```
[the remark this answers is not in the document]
giovanni-guidini: The idea is basically what you commented below:
> not upload a score for this item, or upload a score of 0. i think langfuse doesn't like missing scores so for now it's easier to upload a 0

If we failed to generate suggestions for some reason, there's a score for that (1, error). If we succeeded, the score is 0 (no error).

Runs where `error_running_evaluation` is 1 should be mostly ignored?

There's probably a better way to handle run errors, I'm sure... 😅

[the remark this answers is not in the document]
kddubey: ah i missed that. that's good

src/seer/automation/codegen/evals/evaluations.py
```diff
@@ -0,0 +1,144 @@
...
+    You are given a list of actual bugs in the codebase, and a list of suggestions made by an AI model.
+
+    <goal>
+    Score how well an AI model predicts bugs in the codebase, given we know the actual bugs.
+    Follow the reasoning rules below to score the model.
+    </goal>
+
+    <reasoning_rules>
+    When evaluating the actual bugs in the suggestions, consider:
+    1. WHAT the bug is, and what the suggestion says it is should be similar and clear for a developer reading the suggestion.
```
kddubey, line 103: nit: for now should be less about accurate suggestions, and more about accurately identifying the problem itself

src/seer/automation/codegen/evals/tasks.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    langfuse = Langfuse()
+
+    dataset_item = langfuse.get_dataset_item(item_id)
+
+    logger.info(
+        f"Starting relevant warnings evaluation for item {item_id}, number {item_index}/{item_count}, with run name '{run_name}'."
+    )
+
+    scoring_n_panel = 5  # do we even need this?
```
kddubey, line 99: doesn't look like we use it, so no. we can see later if this is useful, but i doubt it. we'll likely realize that there are more important places to iterate

kddubey (approved):
ty for updating

for some time, we should check predictions and scores for individual items to make sure they're sane, and feel free to iterate

giovanni-guidini:
> for some time, we should check predictions and scores for individual items to make sure they're sane, and feel free to iterate

Here are you considering "individual item" to be a PR or each individual bug/suggestion?

kddubey:
> Here are you considering "individual item" to be a PR or each individual bug/suggestion?

at first was thinking PR but w/in these yes, for some time, we should be scrutinizing how well individual bug predictions line up with the expected output and whether the scores seem to capture that

PR #2495, opened 2025-04-25
fix(autofix): Attempt fix for repo metadata read

jennmueng:
o3 suggested that the undocumented `repositories/:id` endpoint is not whitelisted for github app installations? This will try to use the full name before using the id.

- fixes a trace for repo downloading
- adds spans for the repo metadata read call


kddubey (approved):
🙏

PR #2504, opened 2025-04-29
fix(autofix): Truncate long tool results in autofix agent if erroring from context limit

roaga:
If our LLM calls in the Autofix agent error from a context limit error (custom defined for each LLM provider), we now omit the content from the 3 largest tool results and retry once. This should resolve our intermittent "prompt is too long" errors from Anthropic.


src/seer/automation/autofix/autofix_agent.py
```diff
@@ -209,15 +232,37 @@ def get_completion(
...
+                    f"Input too long for model {run_config.model.model_name} after initial attempt(s). "
+                    "Truncating all tool messages and retrying with backoff."
+                )
+                truncated_memory = self._truncate_biggest_n_tool_messages(self.memory, 3)
+                get_completion_retryable_truncated = retrier(
+                    lambda: attempt_completion(messages_to_send=truncated_memory)
+                )
+                return get_completion_retryable_truncated()
+            else:
+                raise e
```
kddubey, line 265: implementation-wise, can this be cleaner by making it a decorator?

```python
def maybe_truncate_and_retry(get_completion):
    @wraps(get_completion)
    def wrapped(self, run_config):
        # try-except stuff here
```

so that `get_completion` looks simpler

```python
        get_completion_retryable = retrier(maybe_truncate_and_retry(self._get_completion))
        return get_completion_retryable(run_config)
```

this pattern seems more extensible if there are other things we want to retry once

src/seer/automation/autofix/autofix_agent.py
```diff
@@ -209,15 +232,37 @@ def get_completion(
...
+        try:
+            return get_completion_retryable()
+        except Exception as e:
+            is_input_too_long_func = getattr(run_config.model, "is_input_too_long", lambda _: False)
+            if is_input_too_long_func(e):
+                logger.warning(
+                    f"Input too long for model {run_config.model.model_name} after initial attempt(s). "
+                    "Truncating all tool messages and retrying with backoff."
+                )
+                truncated_memory = self._truncate_biggest_n_tool_messages(self.memory, 3)
```
kddubey, line 259: curious how you landed on 3️⃣

src/seer/automation/autofix/autofix_agent.py
```diff
@@ -98,7 +98,30 @@ def _submit_insight(self, text: str):
             )
         )
 
-    def _get_completion(self, run_config: RunConfig):
+    def _truncate_biggest_n_tool_messages(self, messages: list[Message], n: int) -> list[Message]:
```
kddubey, line 101: nit: `truncate` -> `omit`

kddubey (approved):

src/seer/automation/autofix/autofix_agent.py
roaga: it just seems like a good number based on the cases we typically see where it's just slightly over the context limit. Good balance between cutting down length and keeping all the context.

src/seer/automation/autofix/autofix_agent.py
```diff
@@ -98,7 +98,30 @@ def _submit_insight(self, text: str):
...
+        ]
+        tool_messages_with_size.sort(key=lambda x: x[1], reverse=True)
+        indices_to_truncate = {idx for idx, size in tool_messages_with_size[:n]}
+
+        if indices_to_truncate:
+            new_messages = []
+            for i, msg in enumerate(messages):
+                if i in indices_to_truncate:
+                    modified_msg = msg.model_copy(deep=True)
+                    modified_msg.content = "[omitted for brevity]"
```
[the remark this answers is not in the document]
roaga: I fear that anything too explicit will confuse the agent, because it treats memory as the actual conversation that happened. Usually the thoughts after the tool result are enough to convince it that it got what it needed anyway.

PR #2505, opened 2025-04-29
fix(autofix): Set temp to 0 for gemini flash

roaga:
For the context gathering step in the root cause and solution, switches the temperature from 1.0 (remnant from claude thinking) to 0.0. Seems to improve evals slightly by a few points.


kddubey (approved):
is this in fahrenheit or celsius??

roaga:
kelvin @kddubey

PR #2506, opened 2025-04-29
feat(autofix): Infer commit title format from previous commits

roaga:
Gets 5 recent commit messages from the repo and asks the change describer to follow the format it sees. Replaces the hardcoded format when available/

Example prompt excerpt for an issue within `getsentry/sentry`:
Screenshot 2025-04-29 at 4 32 20 PM

Example output:
Screenshot 2025-04-29 at 4 32 22 PM

(ignore the spaces in the prompt's example messages, screenshot is outdated)


src/seer/automation/codebase/repo_client.py
```diff
@@ -454,6 +454,15 @@ def get_valid_file_paths(self, commit_sha: str | None = None) -> set[str]:
 
         return valid_file_paths
 
+    @functools.lru_cache(maxsize=16)
```
jennmueng, line 457: is this something we even call multiple times per repo to need a cache?

jennmueng (approved):

PR #2508, opened 2025-04-30
feat(autofix): Use much faster ripgrep in place of grep

jennmueng:
ripgrep should be much faster to run vs grep. We also rework the tool to be more defined with specific parameters rather than letting the LLM loose with grep.

Hard to get a real apples to apples comparison of speed from our evals given that I'd have to spin up a notebook to pull langfuse spans, so next best thing is a comparison:

Example of similar grep vs a much faster ripgrep:

CleanShot 2025-05-01 at 02 32 26@2x

CleanShot 2025-05-01 at 02 24 31@2x

After using it a couple of times, in a couple eval runs, seems to be no real memory issues caused by ripgrep itself


jennmueng:
> ### 🚨 Sentry detected 1 potential issue in your recent changes 🚨
> * [Line 15, src/seer/automation/autofix/tools/ripgrep_search.py](https://github.com/getsentry/seer/blob/97fc6377cbebaccbb6a709173fdecc8cdb104473/src/seer/automation/autofix/tools/ripgrep_search.py#L15): `'subprocess' call with 'shell=True' identified, security issue`
>   
>   * Avoid `shell=True` in `subprocess.run` to prevent command injection.
> 
> Did you find this useful? React with a 👍 or 👎

this is fine as we make the command ourselves rather than having the llm make the command.

roaga (approved):
this looks good, but
1. what's the impact on eval scores?
2. would this pattern of giving multiple params instead of a raw command improve things for the `find` tool as well?

jennmueng:
> this looks good, but
> 
> 1. what's the impact on eval scores?
> 2. would this pattern of giving multiple params instead of a raw command improve things for the `find` tool as well?


Fundamentally the same if not slightly better solution is fixed (if not for flakiness)
CleanShot 2025-05-01 at 21 59 11@2x
CleanShot 2025-05-01 at 22 03 13@2x

2. Yes I think so

PR #2514, opened 2025-05-02
feat(autofix): Better ripgrep option, response, and logging

jennmueng:
1. Add `fixed_strings` as an option to ripgrep as another search option
2. Better responses to show the command to the LLM for debugging purposes
3. Add logging in failure cases to `ls` the repo and the types
4. Bumps ripgrep threads to 2

### DIFF
--- a/src/seer/automation/autofix/tools/ripgrep_search.py
+++ b/src/seer/automation/autofix/tools/ripgrep_search.py
@@ -3,6 +3,7 @@ import subprocess
 from pathlib import Path
 
 from langfuse.decorators import observe
+import sentry_sdk
 
 MAX_RIPGREP_TIMEOUT_SECONDS = 20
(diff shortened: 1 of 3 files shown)


src/seer/automation/autofix/tools/ripgrep_search.py
```diff
@@ -1,19 +1,24 @@
...
+logger = logging.getLogger(__name__)
+
 
 @observe(name="Run ripgrep in repo")
 def run_ripgrep_in_repo(
     repo_dir: str, cmd: list[str], timeout: float = MAX_RIPGREP_TIMEOUT_SECONDS
 ) -> str:
     try:
+        prepared_cmd = " ".join(cmd)
         result = subprocess.run(
```
trillville, line 20: add span?

trillville (approved):

PR #2519, opened 2025-05-05
fix(autofix): Fast fail expand_document when file doesn't exist

jennmueng:
**Main fix is:** We shouldn't be falling back to waiting & checking on the downloaded repo for `expand_document` if the file path doesn't even exist.

**Included refactors:**
This PR also changes the behavior so we move the file path fixing to the tool call and outside of the Autofix context. This is because any stored file path that's used elsewhere such as the memory prefill should be made correct anyways, and isolates LLM-related feedback behavior to the tool. We should be isolating any LLM-specific error handling and usage to the tool set because we use the Autofix context elsewhere

`get_index_file_set` and `get_valid_file_paths` from `RepoClient` were redundant, and the former was originally for when we were making codebase indexes. Refactors all uses to use the latter.


src/seer/automation/autofix/tools/tools.py
```diff
@@ -131,6 +138,21 @@ def semantic_file_search(self, query: str, llm_client: LlmClient = injected):
     @observe(name="Expand Document")
     @sentry_sdk.trace
     def expand_document(self, file_path: str, repo_name: str):
+        fixed_repo_name = (
+            self.context.autocorrect_repo_name(repo_name)
+            if isinstance(self.context, AutofixContext)
```
roaga, line 143: not for this pr but i don't like how we are special-casing logic between codecov and autofix contexts

src/seer/automation/autofix/tools/tools.py
```diff
@@ -719,9 +739,21 @@ def handle_claude_tools(self, **kwargs: Any) -> str:
...
 
+        if not repo_name:
+            return "Error: Repo could not be resolved"
+
         tool_call_id = kwargs.get("tool_call_id", None)
         current_memory_index = kwargs.get("current_memory_index", -1)
 
+        fixed_repo_name = (
+            self.context.autocorrect_repo_name(repo_name)
+            if isinstance(self.context, AutofixContext)
```
roaga, line 750: why error out when no repo name? what if we're in a situation where there is only one repo? then the autocorrect below should handle it, right?

roaga (approved):

src/seer/automation/autofix/tools/tools.py
jennmueng: ah, `self._get_repo_name_and_path` above should always ensure that `repo_name` is defined when there's one repo, the erroring out is to appease mypy because in our code path we have to set it to `str | None` lol

PR #2520, opened 2025-05-05
log(autofix): Add logging for when repos are large

jennmueng:
Add logging to get more debugging evidence.


roaga (approved):
does this add significant latency?

jennmueng:
@roaga  only if we get rate limited and it does the backoff again, but thats every github app call...

PR #2521, opened 2025-05-05
feat(autofix): Structure insight card output

roaga:
Insight card justifications now output a structured model with:
- text
- markdown of snippets
- links to traces, profiles, source code, and more

Should be backwards compatible as we're only adding fields and give default values for all of them. 
This must go in before the frontend PR (https://github.com/getsentry/sentry/pull/90963)


src/seer/automation/autofix/components/insight_sharing/component.py
```diff
@@ -69,22 +83,111 @@ def format_step_one(
...
+                start_line = result[1]
+                end_line = result[2]
+                if repo_client.provider == "github":
+                    code_url = f"https://github.com/{repo_name}/blob/{repo_client.base_branch}/{file_name}#L{start_line}-L{end_line}"
+                    if code_url not in final_sources.code_used_urls:
+                        final_sources.code_used_urls.append(code_url)
+            else:
+                if repo_client.provider == "github":
+                    code_url = (
+                        f"https://github.com/{repo_name}/blob/{repo_client.base_branch}/{file_name}"
```
jennmueng, line 176:
```suggestion
                        f"https://github.com/{repo_name}/blob/{repo_client.base_commit_sha}/{file_name}"
```

PR #2524, opened 2025-05-06
chore(autofix): Bump timeline formatters to 2.5 flash

roaga:
Bumps evals to about 65% solution fixed


jennmueng (commented):
in line with my comment in an earlier thread on slack, should we be using `preview` models this extensively, are they introducing instability into our pipeline?

roaga:
> in line with my comment in an earlier thread on slack, should we be using `preview` models this extensively, are they introducing instability into our pipeline?

@jennmueng i don't think it's preview models specifically, it's just all the SOTA models. And anyway I would trade slight unreliability for 5% more accuracy

jennmueng (approved):

PR #2533, opened 2025-05-06
fix(anomaly-detection): way more prophet predictions than required

ram-senth:
Due to a bug in the code we were predicting way more (100x more) data points than what was actually needed. This change reduces that. It also includes some refactoring and reduction of the history for which batch inference in done  in combo mode.

### DIFF
--- a/src/seer/anomaly_detection/anomaly_detection.py
+++ b/src/seer/anomaly_detection/anomaly_detection.py
@@ -84,7 +84,7 @@ class AnomalyDetection(BaseModel):
         batch_detector = MPBatchAnomalyDetector()
         prophet_detector = ProphetAnomalyDetector()
         forecast_len = (
-            prophet_forecast_len_days * 24 * 60 // config.time_period
+            prophet_forecast_len_days * 24 * (60 // config.time_period)
             if prophet_forecast_len_days
             else algo_config.prophet_forecast_len * (60 // config.time_period)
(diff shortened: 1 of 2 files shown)


src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -480,7 +446,9 @@ def _combo_detect(
             window_size=None,
             time_budget_ms=time_budget_ms,
             algo_config=algo_config,
-            prophet_forecast_len_days=max_stream_detection_len,  # This ensures that the there is only prophet detection run always
+            prophet_forecast_len_days=algo_config.max_stream_days_for_combo_detection[
```
ram-senth, line 449: This is the actual bug fix - the total number of predictions was being passed as number of days.

src/seer/anomaly_detection/detectors/anomaly_scorer.py
```diff
@@ -277,7 +277,7 @@ def merge(timestamps, mp_flags_and_scores, prophet_map):
                     flags.append(mp_flag)
                 previous_flag = flags[-1]
                 algo_types.append(algo_type)
-            if missing > 0:
+            if found == 0:
```
ram-senth, line 280: This is reduce the noise in logging. When we limit the history size it is expected to not have predictions for some of the data.

PR #2540, opened 2025-05-07
chore(anomaly-detection): separate transaction names for the different operations

ram-senth:
With our current API, the same detection endpoint is serving three different use cases that need three different SLAs. It is hard to monitor and alert on breaches for each one. This PR ensures that transaction of each use case has unique identification.

### DIFF
--- a/src/seer/anomaly_detection/anomaly_detection.py
+++ b/src/seer/anomaly_detection/anomaly_detection.py
@@ -7,7 +7,6 @@ import numpy as np
 import pandas as pd
 import sentry_sdk
 import stumpy  # type: ignore # mypy throws "missing library stubs"
-from datadog.dogstatsd.base import statsd
 from pydantic import BaseModel
 from stumpy import cache
 
@@ -41,7 +40,7 @@ from seer.anomaly_detection.models.timeseries import ProphetPrediction
 from seer.db import TaskStatus
 from seer.dependency_injection import inject, injected
 from seer.exceptions import ClientError, ServerError
-from seer.tags import AnomalyDetectionModes, AnomalyDetectionTags
+from seer.tags import AnomalyDetectionTags
 
 anomaly_detection_module.enable()
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


src/seer/anomaly_detection/anomaly_detection.py
```diff
@@ -598,31 +597,16 @@ def detect_anomalies(
...
         if isinstance(request.context, AlertInSeer):
-            mode = AnomalyDetectionModes.STREAMING_ALERT
+            sentry_sdk.set_tag(AnomalyDetectionTags.ALERT_ID, request.context.id)
+            ts, anomalies = self._online_detect(request.context, request.config)
         elif isinstance(request.context, TimeSeriesWithHistory):
-            mode = AnomalyDetectionModes.STREAMING_TS_WITH_HISTORY
-        else:
-            mode = AnomalyDetectionModes.BATCH_TS_FULL
-
-        sentry_sdk.set_tag(AnomalyDetectionTags.MODE, mode)
```
ram-senth, line 607: My original intention with mode was to use that for filtering. However, mode is not surfaced as a tag at the transaction level and hence its not of much use. So changing the transaction names instead and removing the mode tag.

src/seer/app.py
```diff
@@ -450,15 +452,27 @@ def detect_anomalies_endpoint(data: DetectAnomaliesRequest) -> DetectAnomaliesRe
...
+    elif isinstance(data.context, TimeSeriesWithHistory):
+        transaction_name = f"{BASE_TRANSACTION_NAME}.combo"
+    else:
+        transaction_name = f"{BASE_TRANSACTION_NAME}.batch"
+
+    scope = sentry_sdk.get_current_scope()
+    scope.set_transaction_name(transaction_name)
+
     try:
-        with statsd.timed("seer.anomaly_detection.detect.duration"):
```
ram-senth, line 454: There is no real value in gathering combined Datadog stats at this level as each underlying use case - online, batch and combo detection - has very different perf characteristics. So removing the outer one and just keeping individual metrics.

src/seer/anomaly_detection/anomaly_detection.py
ram-senth: Actually, I'm reinstating the mode tag as it can still be handy when looking at spans.

PR #2543, opened 2025-05-07
feat(relevant-warnings): Add mypy to gha

suejung-sentry:
Adjust the relevant-warnings dogfooding to use outputs from both ruff and mypy (instead of ruff only)

### DIFF
--- a/src/seer/automation/codegen/relevant_warnings_step.py
+++ b/src/seer/automation/codegen/relevant_warnings_step.py
@@ -58,7 +58,7 @@ class RelevantWarningsStepRequest(PipelineStepTaskRequest, CodegenRelevantWarnin
 
 class RelevantWarningsStep(CodegenStep):
     """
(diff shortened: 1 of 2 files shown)


kddubey (approved):

src/seer/automation/codegen/relevant_warnings_step.py
```diff
@@ -58,7 +58,7 @@ class RelevantWarningsStepRequest(PipelineStepTaskRequest, CodegenRelevantWarnin
 
 class RelevantWarningsStep(CodegenStep):
     """
-    Predicts which static analysis warnings in a pull request are relevant to a past Sentry issue.
+    Predicts which static analysis warnings in a pull request are relevant to a past Sentry issue
```
suejung-sentry, line 61: CI was getting stuck waiting for results of the run so forced a change within the main app to get those gha steps to run..

PR #2544, opened 2025-05-07
feat(autofix): Use a git client to clone and checkout repos

jennmueng:
This PR changes how we download repos from downloading repo archives to cloning the repo with `depth=1` and then checking out a specific commit sha.

- This sets us up to be able to cache and load repos from GCS using this `RepoManager`
- As the repo is downloading, we also update the Autofix state to ensure that the state does not get timed out

## Implementation
Introduce the `RepoManager`, conceptually it's supposed to be the class that manages a repo's download, caching etc. It should encapsulate the entire lifecycle of a downloaded repo. It's also entirely source control provider agnostic, and instead relies on getting passed a `RepoClient` that it interacts with.

## How it works
1. The tools are initialized, the `RepoManager`s are created and then called to initialize each on its own thread in the background.
2. The repo is cloned with `depth=1`
3. The repo is fetched at a specific commit sha from the `RepoClient` and checked out at that specific sha.
4. Once a tool is called that needs to use the downloaded repo, via `BaseTools._ensure_repos_downloaded`, we wait for the futures with a `120s` timeout.
5. If a repo download times out, we tell the LLM that the repo is not usable and to not use that tool with the repo.

## Next Steps
1. Setup GCS buckets in the required regions
2. Implement the GCS upload/download in the `RepoManager`

### DIFF
--- a/src/seer/automation/autofix/tools/tools.py
+++ b/src/seer/automation/autofix/tools/tools.py
@@ -39,7 +39,7 @@ REPO_WAIT_TIMEOUT_SECS = 120.0
 class BaseTools:
     context: AutofixContext | CodegenContext
     retrieval_top_k: int
-    repo_managers: Dict[str, RepoManager] = {}
+    repo_managers: dict[str, RepoManager] = {}
     repo_client_type: RepoClientType = RepoClientType.READ
 
     def __init__(
@@ -75,6 +75,7 @@ class BaseTools:
             # Do nothing, the state should self update updated_at
             pass
 
+    @sentry_sdk.trace
     def _ensure_repos_downloaded(self, repo_name: str | None = None):
         """
         Helper method to wait for repos to be downloaded.
--- a/src/seer/automation/autofix/tools/tools.py
+++ b/src/seer/automation/autofix/tools/tools.py
@@ -5,7 +5,7 @@ import subprocess
 import textwrap
 from concurrent.futures import Future, ThreadPoolExecutor, TimeoutError, as_completed
 from threading import Lock
-from typing import Any, cast
+from typing import Any, cast, Dict
 
 import sentry_sdk
 from langfuse.decorators import observe
@@ -21,6 +21,7 @@ from seer.automation.autofix.models import AutofixRequest
 from seer.automation.autofix.tools.read_file_contents import read_file_contents
 from seer.automation.autofix.tools.ripgrep_search import run_ripgrep_in_repo
 from seer.automation.codebase.file_patches import make_file_patches
+from seer.automation.codebase.repo_manager import RepoManager
 from seer.automation.codebase.models import BaseDocument
 from seer.automation.codebase.repo_client import RepoClientType
 from seer.automation.codebase.utils import cleanup_dir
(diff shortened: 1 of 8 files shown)


src/seer/automation/autofix/tools/tools.py
```diff
@@ -120,6 +158,9 @@ def _make_repo_not_found_error_message(self, repo_name: str) -> str:
             else ""
         )
 
+    def _make_repo_unavailable_error_message(self, repo_name: str) -> str:
+        return f"Error: We had an issue loading the repository `{repo_name}`. This tool is unavailable for this repository, you must stop using it for this repository `{repo_name}`."
+
```
jennmueng, line 163: Originally wanted to just mark the repo as not readable, but realized we don't update the system prompt retroactively to the state once it's created, so the next best thing is to just tell the LLM...

src/seer/automation/autofix/tools/tools.py
```diff
@@ -304,26 +348,14 @@ def _normalize_path(self, path: str) -> str:
...
-        # Ensure lock exists before using it
-        if not hasattr(self, "_tmp_dir_lock"):
-            self._tmp_dir_lock = Lock()
-
-        # Clean up any in-progress downloads
-        if self._download_future and not self._download_future.done():
+        """Clean up all repository clients."""
+        for repo_name, local_client in list(self.repo_managers.items()):
             try:
-                self._download_future.cancel()
```
jennmueng, line 314: can't actually cancel a thread

src/seer/automation/autofix/tools/tools.py
```diff
@@ -33,15 +33,14 @@
...
 MAX_FILES_IN_TREE = 100
+REPO_WAIT_TIMEOUT_SECS = 120.0
 
 
 class BaseTools:
     context: AutofixContext | CodegenContext
     retrieval_top_k: int
-    tmp_dir: dict[str, tuple[str, str]] = {}  # Maps repo_name to (tmp_dir, tmp_repo_dir)
-    tmp_repo_dir: str | None = None
+    repo_managers: Dict[str, RepoManager] = {}
```
roaga, line 42: why `Dict` over `dict`?

src/seer/automation/autofix/tools/tools.py
```diff
@@ -52,47 +51,88 @@ def __init__(
...
+        self._download_repos()
 
-    def _start_parallel_repo_download(self):
-        """Start downloading all repositories in parallel in the background."""
+    def _download_repos(self):
         repo_names = self._get_repo_names()
         if not repo_names:
             return
 
-        @sentry_sdk.trace
```
roaga, line 67: can we put back trace instrumentation?

roaga (commented):
LGTM 🚀 
Will using clone affect speed compared to the tarball from before?

src/seer/automation/autofix/tools/tools.py
jennmueng: cursor vibe

jennmueng:
@roaga yeah we would be increasing the latency a bit with the clone + checkout strat but hopefully the fast follow of caching evens it out.

roaga (approved):
i'm just gonna trust that you know what you're doing

PR #2549, opened 2025-05-08
feature(anomale-detection): use approximate MP algorithm to improve performance

ram-senth:
This change switches to using approximate matrix profile that is faster than the exact MP and not too far off.


ram-senth:
Approximate MP improves the time taken for the worst case but makes it worse for the median case. So not going forward with this change.
newplot

PR #2550, opened 2025-05-08
feat(autofix): Run repos with GCS caching

jennmueng:
This PR introduces caching of git repos in GCS.

When initializing repos, it works in the steps below (bolded steps are new with this PR):
1. Downloads the repo
 1.1. **If GCS caching is enabled and the repo archive exists, downloads it from GCS**
 1.2. Otherwise Clones a repo using `git clone`
2. Syncs it to the expected commit sha 
**3. If GCS caching is enabled**
 **3.1. If the repo has not been cached before**
   **- Copies the repo into a separate directory**
   **- "Prunes" the git directory to be a minimal representation of the repository**
   **- Creates a tar archive of the repo**
   **- Uploads the tar archive of the repo to GCS**
 **3.2. If the repo already has been cached**
  **- Uses the db entry's sha and the new sha to compare the commit timestamps**
  **- If the new synced version has a newer commit time, then we upload it according to the steps in 3.1**

## Limitation
With this setup enabled, we'll be still doing a slow download on the first use of a repo, and this is a current limitation until we introduce background codebase caching.

## Artifacts
A cached repo has 2 artifacts, first the tar archive itself on GCS, and a db entry for it that contains the current sha. 

## Note
Note: GCS Caching with this PR is only going to be enabled for the Seer project in the Sentry org for testing and validation purposes.


src/seer/automation/autofix/tools/tools.py
```diff
@@ -86,16 +92,20 @@ def _ensure_repos_downloaded(self, repo_name: str | None = None):
                       If None, waits for all repos to be downloaded.
         """
         if repo_name:
+            if repo_name not in self.repo_managers:
+                raise ValueError(f"Repository {repo_name} not found")
             repo_managers_to_wait_for = (
                 [self.repo_managers[repo_name]]
-                if not self.repo_managers[repo_name].is_available
+                if (not self.repo_managers[repo_name].is_available)
+                and (not self.repo_managers[repo_name].is_cancelled)
```
jennmueng, line 100: fixes bug where we wait for cancelled repos

src/seer/automation/codebase/repo_client.py
```diff
@@ -171,6 +171,7 @@ class RepoClient:
     repo_external_id: str
     base_commit_sha: str
     base_branch: str
+    repo_definition: RepoDefinition
```
jennmueng, line 174: because we initialize with one of these anyways, should just store it

src/seer/automation/codebase/repo_manager.py
```diff
@@ -27,51 +46,128 @@ class RepoManager:
...
+    is_cancelled: bool
 
+    organization_id: int
+    project_id: int
+
+    _use_gcs: bool
     _last_liveness_update: float
     _trigger_liveness_probe: Callable[[], None] | None
-    _has_timed_out: bool
+    _app_config: AppConfig  # This is here because we can't inject on a property...
```
roaga, line 57: does it really need to be a property? can't we just make it `get_bucket_name()` for example?

src/seer/automation/codebase/repo_manager.py
```diff
@@ -27,51 +46,128 @@ class RepoManager:
...
+        self.organization_id = organization_id
+        self.project_id = project_id
+
         self._trigger_liveness_probe = trigger_liveness_probe
-        self._has_timed_out = False
+        self.is_cancelled = False
         self._last_liveness_update = 0.0
+        self._app_config = config
+
+        self._use_gcs = organization_id == 1 and project_id == 6178942  # Hardcoded ONLY for Seer
```
roaga, line 88:
```suggestion
        self._use_gcs = organization_id == 1 and project_id == 6178942  # TODO: Hardcoded ONLY for Seer
```

src/seer/automation/codebase/repo_manager.py
```diff
@@ -125,23 +221,77 @@ def _sync_repo(self):
...
         try:
             start_time = time.time()
-
             commit_sha = self.repo_client.base_commit_sha
 
+            # Reset any perceived local changes first
+            self.git_repo.git.reset("--hard")
+            # Clean any untracked files
+            self.git_repo.git.clean("-fdx")
+
```
roaga, line 230: this won't actually do anything in practice right? do we actually modify downloaded files anywhere?

src/seer/automation/codebase/repo_manager.py
```diff
@@ -125,23 +221,77 @@ def _sync_repo(self):
...
+            # Force checkout to avoid the "local changes" error
+            self.git_repo.git.checkout(commit_sha, force=True)
 
             end_time = time.time()
             logger.info(
                 f"Checked out repo {self.repo_client.repo_full_name} to commit {commit_sha} in {end_time - start_time} seconds"
             )
+
+            if self._use_gcs:  # We'll only hardcode this to be used with the seer project
```
roaga, line 242:
```suggestion
            if self._use_gcs:
```

src/seer/db.py
```diff
@@ -288,6 +288,36 @@ class DbSeerProjectPreference(Base):
     __table_args__ = (UniqueConstraint("organization_id", "project_id"),)
 
 
+class DbSeerRepoArchive(Base):
+    __tablename__ = "seer_repo_archive"
+    id: Mapped[int] = mapped_column(BigInteger, nullable=False, primary_key=True)
+    organization_id: Mapped[int] = mapped_column(BigInteger, nullable=False)
+    project_id: Mapped[int] = mapped_column(BigInteger, nullable=False)
```
roaga, line 295: should repos be associated with a specific project? it's possible users will have them assigned to multiple projects (e.g. we use `getsentry/sentry` in both `seer` and `sentry`). So we wouldn't need to maintain two separate archives for it.

src/seer/automation/codebase/repo_manager.py
jennmueng: It shouldn't, it was added because we were getting unclean checkout states from the archive for some reason

src/seer/db.py
jennmueng: fair, we can just box it by org and that should be enough

tests/automation/codebase/test_repo_manager.py
```diff
@@ -501,4 +505,263 @@ def test_upload_lock_already_locked(repo_manager, caplog):
             with repo_manager.upload_lock():
                 pass
 
-    assert "Repository is already locked for upload" in caplog.text
+        mock_logger.info.assert_called_once()
+        # Optionally, verify message content
+        assert "Repository is already locked for upload" in mock_logger.info.call_args[0][0]
```
roaga, line 510: what in the vibes

src/seer/automation/codebase/repo_manager.py
jennmueng: yes it is, but i couldn't get to the bottom of it and this makes sure the state is clean + the verification passes

PR #2557, opened 2025-05-09
fix(seer): Fix the logging

jennmueng:
Our logging was a mess with duplicated logs, hard to read json format, and complicated implementation.

This PR simplifies that by using the standard logger, with a substring filter that filters out noisy logs. Especially needed now that we show these logs in Sentry.

May the screenshots below be all the rest of what I have to say.

## Before:
CleanShot 2025-05-09 at 15 25 19@2x

## After:
CleanShot 2025-05-09 at 15 23 33@2x

### DIFF
--- a/src/seer/bootup.py
+++ b/src/seer/bootup.py
@@ -1,20 +1,17 @@
 import logging
 
-
 import sentry_sdk
 from psycopg import Connection
 from sentry_sdk.integrations import Integration
 from sentry_sdk.types import Event
 
+# Make sure this import is here
+import seer.logging  # noqa: F401
 from seer.automation.utils import AgentError
 from seer.configuration import AppConfig
 from seer.db import initialize_database
 from seer.dependency_injection import Module, inject, injected
 from seer.inference_models import initialize_models
-
-# Make sure this import is here
-import seer.logging  # noqa: F401
-
 
 logger = logging.getLogger(__name__)
 
--- a/src/seer/bootup.py
+++ b/src/seer/bootup.py
@@ -1,5 +1,6 @@
 import logging
 
+
 import sentry_sdk
 from psycopg import Connection
 from sentry_sdk.integrations import Integration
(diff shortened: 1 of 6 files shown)


src/seer/bootup.py
```diff
@@ -5,12 +5,13 @@
 from sentry_sdk.integrations import Integration
 from sentry_sdk.types import Event
 
+# Make sure this import is here
```
roaga, line 8: Bad vibes

src/seer/bootup.py
```diff
@@ -37,7 +38,6 @@ def bootup(
 ):
     initialize_sentry_sdk(integrations)
     with sentry_sdk.metrics.timing(key="seer_bootup_time"):
```
roaga, line 40: We can just remove this whole metrics wrapper

src/seer/logging.py
```diff
@@ -84,26 +18,35 @@ def access(self, resp, req, environ, request_time):
...
+                "AFC is enabled",  # Google genai library logs this
+                "AFC remote call",  # Google genai library logs this
+                "Item exceeds size limit",  # Langfuse
+            ]
+        )
     )
-)
-
-logging_module.enable()
+    logger.setLevel(logging.INFO)
```
roaga, line 36: Does setting to info still allow warnings and errors to show up properly?

kddubey:
does this also address that all of our logs, including INFO level ones, show up as severity="ERROR" in GCP?

src/seer/logging.py
jennmueng: yep, cuz INFO is a more verbose log level than warnings and errors

jennmueng:
@roaga I'm pretty sure the Sentry SDK breadcrumb logging works independently of the structlog formatting we had
@kddubey I'm not sure, and I'm also not sure what's causing that issue in GCP

PR #2558, opened 2025-05-09
chore: don't mock PrFiles when running evaluations

giovanni-guidini:
The idea of mocking PrFiles in the evals was to allow us to run them without reaching out to GitHub. However with the move to the agentic approach this is useless. The agents have code searching capabilities, and are expected to communicate quite often with GitHub anyway.

So when running evaluations the GitHub token you have in `.env` needs to be able to access all the repos for which there are eval items.


kddubey (approved):
🧹

PR #2561, opened 2025-05-09
Aiml 286 commitsha with autofix run

Mihir-Mavalankar:
# PR Details
+ Ticket: [AIML-286](https://linear.app/getsentry/issue/AIML-286/fix-draft-pr-failing-due-to-changed-codebase)
+ I implemented this change by changing the _clone_repo function to clone the repo and checkout to a specific commit hash 
+ The commit hash it uses is the `base_commit_hash` from the `RepoClient` class. This commit hash is passed to the class at initialization or if it's not set, the commit of the branch HEAD is fetched.
+  I removed the _sync_repo method since that method is now rolled up into _clone_repo
  + I'm okay will adding _sync_repo back if anyone thinks we might need it in other functions. 
+ pre-commit file is updated too since "push" isn't a valid stage for the git hook.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,7 +36,7 @@ repos:
         types: [python]
       - id: prevent-push
         name: prevent pushing to main
-        stages: [push]
+        stages: [pre-push]
         entry: bash -c 'test "$PRE_COMMIT_REMOTE_BRANCH" != "refs/heads/main"'
         always_run: true
         pass_filenames: false
--- a/src/seer/automation/codebase/repo_manager.py
+++ b/src/seer/automation/codebase/repo_manager.py
@@ -104,7 +104,6 @@ class RepoManager:
                 self.repo_path,
                 progress=lambda *args, **kwargs: self._throttled_liveness_probe(),
                 depth=1,
-                no_checkout=True,
             )
             # Fetch the specific commit
             try:
@@ -116,7 +115,7 @@ class RepoManager:
             self.git_repo.git.checkout(commit_sha)
             end_time = time.time()
             logger.info(
-                f"Cloned and checked out repository at commit {commit_sha} in {end_time - start_time} seconds"
+                f"Cloned and checked out repository at commit {commit_sha} in {end_time - start_time:.4f} seconds"
             )
 
             return self.repo_path
--- a/src/seer/automation/codebase/repo_manager.py
+++ b/src/seer/automation/codebase/repo_manager.py
@@ -102,8 +102,8 @@ class RepoManager:
             self.git_repo = git.Repo.clone_from(
                 repo_clone_url,
                 self.repo_path,
-                no_checkout=True,
                 depth=1,
+                progress=lambda *args, **kwargs: self._throttled_liveness_probe(),
             )
             # Fetch the specific commit if it's not present
             try:
--- a/src/seer/automation/codebase/repo_manager.py
+++ b/src/seer/automation/codebase/repo_manager.py
@@ -85,9 +85,13 @@ class RepoManager:
     @sentry_sdk.trace
     def _clone_repo(self) -> str:
         """
-        Clone a repository to a local temporary directory.
+        Clone a repository to a local temporary directory and checkout a specific commit.
+        We fetch the specific commit to avoid a scenario where if you start a root cause -> solution
+        -> code changes then someone changes a file, then you try to draft a pr, that file could
+        potentially not exist anymore for you to make changes to. - AIML-286
         """
         repo_clone_url = self.repo_client.get_clone_url_with_auth()
+        commit_sha = self.repo_client.base_commit_sha
 
         try:
             logger.info(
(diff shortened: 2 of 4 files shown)


src/seer/automation/codebase/repo_manager.py
```diff
@@ -116,32 +127,6 @@ def _clone_repo(self) -> str:
             self.git_repo = None  # clear the repo to fail the available check
             raise
 
-    @sentry_sdk.trace
-    def _sync_repo(self):
-        """
-        Ensure the repository is up to date with only the target commit.
-        """
-        logger.info(f"Syncing repository {self.repo_client.repo_full_name}")
-
```
roaga, line 125: i believe we want to preserve `_sync_repo` as a separate method and part of the core logic, instead of rolling all the logic into `clone_repo`. The reason is that we plan to cache downloaded repos in the future, and sync just to get the latest updates.

src/seer/automation/codebase/repo_manager.py
```diff
@@ -99,14 +99,25 @@ def _clone_repo(self) -> str:
...
             self.git_repo = git.Repo.clone_from(
                 repo_clone_url,
                 self.repo_path,
                 progress=lambda *args, **kwargs: self._throttled_liveness_probe(),
                 depth=1,
             )
+            # Fetch the specific commit
+            try:
+                self.git_repo.git.fetch("origin", commit_sha, depth=1)
+            except git.GitCommandError as e:
```
roaga, line 112: just because i don't understand, is `.fetch` different than the `.execute` from before in any way? If not, this is a good improvement for cleanliness

src/seer/automation/codebase/repo_manager.py
```diff
@@ -71,11 +71,9 @@ def initialize(self):
...
             if self._has_timed_out:
                 return
-            self._sync_repo()
-        except Exception:
-            logger.exception(
-                "Failed to initialize repo", extra={"repo": self.repo_client.repo_full_name}
-            )
+        except Exception as e:
+            logger.error(f"Failed to initialize repo {self.repo_client.repo_full_name}: {e}")
+
```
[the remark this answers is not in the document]
Mihir-Mavalankar: Done. Added it back.

src/seer/automation/codebase/repo_manager.py
Mihir-Mavalankar: Done. Added it back.

src/seer/automation/codebase/repo_manager.py
Mihir-Mavalankar: fetch corresponds to the git fetch command and execute is a general function function which has fetch as one of the options. Either command does the same thing. I can change it to fetch in _sync_repo if that looks cleaner.

.pre-commit-config.yaml
```diff
@@ -36,7 +36,7 @@ repos:
         types: [python]
       - id: prevent-push
         name: prevent pushing to main
-        stages: [push]
```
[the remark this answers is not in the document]
Mihir-Mavalankar: I got an error when running this before as push but it seems to have gone now. Reverting this too.

src/seer/automation/codebase/repo_manager.py
roaga: yeah that would be nice

src/seer/automation/codebase/repo_manager.py
```diff
@@ -87,9 +88,11 @@ def initialize(self):
     @sentry_sdk.trace
     def _clone_repo(self) -> str:
         """
-        Clone a repository to a local temporary directory.
+        Clone a repository to a local temporary directory and checkout a specific commit.
+        Reason to checkout specific commit - AIML-286
         """
```
roaga, line 93: i don't think you need the "Reason to checkout specific commit - AIML-286" in the comment

roaga (commented):
- This seems to only affect repo downloads. Does this fix the issue described in the ticket of us drafting PRs against the wrong commit on the remote repo?
- Can we also change the PR's title to use the format we typically use?
- I'd ask @jennmueng for a review, he knows this stuff better

src/seer/automation/codebase/repo_manager.py
Mihir-Mavalankar: Done.

src/seer/automation/codebase/repo_manager.py
```diff
@@ -99,14 +101,19 @@ def _clone_repo(self) -> str:
...
             start_time = time.time()
+            # Clone with no checkout to allow checking out a specific commit
             self.git_repo = git.Repo.clone_from(
                 repo_clone_url,
                 self.repo_path,
                 progress=lambda *args, **kwargs: self._throttled_liveness_probe(),
                 depth=1,
             )
+            # Fetch the commit saved in the repo_client base_commit_sha
+            self._sync_repo()
```
jennmueng, line 112: btw we already call this in the `initialize` method, so we already check out to a specific commit sha after cloning

PR #2569, opened 2025-05-12
feat(codegen): Support multiple repos

kddubey:
This PR de-couples the tools from the type of `context` a little bit. As a result, multi-repo search is supported for `CodegenContext`, which is useful for bug prediction.

This doesn't change functionality. Would need to supply `more_readable_repos` to a `CodegenBaseRequest` to cause multi-repo search.

This PR can be merged in independently of #2568.


src/seer/automation/codegen/codegen_event_manager.py
```diff
@@ -41,6 +41,8 @@ def on_error(
         with self.state.update() as cur:
             cur.status = CodegenStatus.ERRORED
 
-    def send_insight(self, insight: InsightSharingOutput):
-        # Do nothing for now, this is only used for autofix
+    def send_insight(self, insight_card: InsightSharingOutput):
```
kddubey, line 44: fyi just changed the arg name to match `AutofixEventManager` so that there's one param spec

src/seer/automation/codegen/models.py
```diff
@@ -223,19 +241,6 @@ class CodegenRelevantWarningsResponse(CodegenBaseResponse):
...
-    run_id: int
-
-
-class CodegenRelevantWarningsStateResponse(BaseModel):
-    run_id: int
-    status: CodegenStatus
-    relevant_warning_results: list[RelevantWarningResult]
-    triggered_at: datetime.datetime
-    updated_at: datetime.datetime
-    completed_at: datetime.datetime | None = None
```
kddubey, line 236: never used

src/seer/automation/autofix/tools/tools.py
```diff
@@ -155,19 +155,15 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-        return f"Error: Repo '{repo_name}' not found." + (
-            f" Available repos: {', '.join([repo.full_name for repo in self.context.repos if isinstance(self.context, AutofixContext)])}"
-            if isinstance(self.context, AutofixContext)
-            else ""
-        )
+        if isinstance(self.context, AutofixContext):
+            available_repos = self.context.repos
+        else:
+            available_repos = self.context.state.get().readable_repos
+        available_repos_str = ", ".join([repo.full_name for repo in available_repos])
```
roaga, line 165: is there a way to decouple this from the autofix context too?

roaga (approved):
avast m8y, ye the hero of thar lands

src/seer/automation/autofix/tools/tools.py
kddubey: not sure. actually wanted to ask why autofix uses `self.context.repos` here instead of `self.context.state.get().readable_repos`

roaga: yk, idk. it seems like context.repos is just defined as state.readable_repos. And we use both in different places. Only advantage to the context is avoiding an extra db read I guess but I wonder if we could consolidate.

cc @jennmueng since you wrote all this

rohitvinnakota-codecov (approved):

src/seer/automation/autofix/tools/tools.py
jennmueng: this is a bug actually, we should be using `readable_repos` for autofix too, we should also probably be setting `repos` in `AutofixContext` to readable repos there too or remove it - seems to not be used outside of here anymore?

kddubey: there are some places outside of tools where using `request.repos` seems to make sense

https://github.com/getsentry/seer/pull/2569/commits/5edc6ea19e7f952c4c19a56b2fb1b10f1b589219 updates tools to use `readable_repos`. pretty sure these changes make sense but lmk if not

PR #2570, opened 2025-05-12
chore(trace queries): Default cache

aayush-se:
- Default to no values if cache not immediately available upon request
- Fetching values is the longest step of the caching process so skipping this should result in a very minimal cold start

### DIFF
--- a/src/seer/automation/assisted_query/assisted_query.py
+++ b/src/seer/automation/assisted_query/assisted_query.py
@@ -35,8 +35,10 @@ def translate_query(request: TranslateRequest) -> TranslateResponse:
 
     if not cache_name:
         # Will result in cold start
-        logger.info(f"Cache miss for {cache_display_name}, creating new cache")
-        res = create_cache(CreateCacheRequest(org_id=org_id, project_ids=project_ids))
+        logger.info(f"Cache miss for {cache_display_name}, using default cache")
+        res = create_cache(
+            CreateCacheRequest(org_id=org_id, project_ids=project_ids, no_values=True)
+        )
         cache_name = res.cache_name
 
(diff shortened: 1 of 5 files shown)


src/seer/automation/assisted_query/create_cache.py
```diff
@@ -47,39 +48,45 @@ def create_cache(data: CreateCacheRequest, client: RpcClient = injected) -> Crea
...
+                    filtered_fields.append(field)
+
+        filtered_field_values_response = client.call(
+            "get_attribute_values",
+            fields=filtered_fields,
+            org_id=org_id,
+            project_ids=project_ids,
+            stats_period="48h",
+            limit=5,
+        )
```
aayush-se, line 79: I'm not sure why github doesn't show it here, but this code is all old, just indented to be under the if statement

src/seer/automation/assisted_query/assisted_query.py
```diff
@@ -33,8 +33,10 @@ def translate_query(request: TranslateRequest) -> TranslateResponse:
 
     if not cache_name:
         # Will result in cold start
-        logger.info(f"Cache miss for {cache_display_name}, creating new cache")
-        res = create_cache(CreateCacheRequest(org_id=org_id, project_ids=project_ids))
+        logger.info(f"Cache miss for {cache_display_name}, using default cache")
```
ram-senth, line 36: Would adding a tag be more useful than this log message? Tag will allow you to group traces based on cache hit/miss.

aayush-se: updated

ram-senth (approved):

PR #2574, opened 2025-05-12
feat(bug-predictor): Add bug predictor formatter

suejung-sentry:
Add a formatter for the output of the bug predictor.

Relates to https://github.com/getsentry/seer/pull/2568

Closes https://linear.app/getsentry/issue/PREVENT-74/seer-add-a-bug-prediction-post-processor-in-seer


src/seer/automation/codegen/bug_prediction_step.py
```diff
@@ -111,11 +105,11 @@ def _complete_run(
                 else []
             )
 
-    @observe(name="Codegen - Bug Prediction Step")
```
kddubey, line 114: does this not work for you locally? lmk if i can help debug. it can be pretty useful to see traces for the agent calls in the bug predictor so wanna make sure everyone has access

src/seer/automation/codegen/models.py
```diff
@@ -367,6 +367,33 @@ def validate_lists_same_length(
...
+        return RelevantWarningResult(
+            reasoning=self.description,
+            short_description=self.description,
+            short_justification=self.description,
+            encoded_location=self.code_locations[0] if self.code_locations else None,
+        )
+
+
+class BugPredictorFormatterInput(BaseComponentRequest):
+    followups: list[str] | None = None
```
kddubey, line 390:
```suggestion
    followups: list[str | None]
```

`None` indicates the verify agent unexpectedly failed for its hypothesis

src/seer/automation/codegen/bug_prediction_component.py
```diff
@@ -202,3 +205,28 @@ def invoke(
...
+class BugPredictionFormatterComponent(
+    BaseComponent[BugPredictorFormatterInput, BugPredictorFormatterOutput]
+):
+    @inject
+    def invoke(
+        self, request: BugPredictorFormatterInput, llm_client: LlmClient = injected
+    ) -> BugPredictorFormatterOutput:
+        if not request.followups:
+            return BugPredictorFormatterOutput(formatted_predictions=[])
+
```
kddubey, line 219: another early exit is that all of the `followups` are `None` b/c all verify agent calls failed

src/seer/automation/codegen/prompts.py
```diff
@@ -613,3 +613,34 @@ def format_prompt_followup(
             hypothesis=hypothesis,
             focus=cls._focus_on_crashes(),
         )
+
+    @classmethod
+    def format_prompt_bug_prediction_formatter(cls, followups: list[str]) -> str:
```
kddubey, line 618: reminder to handle `list[str | None]` instead

tests/automation/codegen/test_bug_prediction_formatter.py
```diff
@@ -0,0 +1,113 @@
...
+
+from seer.automation.codegen.bug_prediction_component import BugPredictionFormatterComponent
+from seer.automation.codegen.models import BugPredictorFormatterInput, BugPredictorFormatterOutput
+
+
+class TestBugPredictionFormatterComponent:
+    def setup_method(self):
+        self.component = BugPredictionFormatterComponent(context=MagicMock())
+        self.mock_llm_client = MagicMock()
+        self.use_real_llm = os.environ.get("USE_REAL_LLM", "0") == "1"
```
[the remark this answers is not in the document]
kddubey: we've been using VCR tests when we want to locally test LLM API calls and have it be reproducible in CI

some [small examples in here](https://github.com/getsentry/seer/blob/main/tests/automation/agent/test_client.py)

[important docs here](https://github.com/getsentry/seer?tab=readme-ov-file#vcrs)

lmk if it gets annoying—we've been skipping some especially intense VCRd tests. but i think for testing one component / 1 LLM API call it'll be manageable

kddubey (commented):
structure looks good. left some initial comments while you iterate

kind of strong opinion i have is that we should be more choosy—only surface the "i'm pretty sure prod will crash" bugs. a simple reason for this is that the rest of the bug predictions are often incorrect 🥴

tests/automation/codegen/fixtures/bug_prediction/1870.txt
```diff
@@ -0,0 +1,261 @@
+Based on my analysis of the codebase, I can now evaluate the hypothesis about whether the `set_selected_root_cause` method in the `event_manager.py` file could cause server crashes if the payload is invalid.
```
kddubey, line 1: if it's not too much of a hassle, this should prolly be a json list of strings to more generically mimic `followups`

kddubey:
i'll adjust #2568 to output info needed for the location. rn it's not prompted to predict it

src/seer/automation/codegen/bug_prediction_step.py
suejung-sentry: Oh this is really cool - thanks for pointing it out! Removing that decorator was unintentional when just doing local testing here but glad I did to pick up this tidbit!
I put it back and see it populating in dev langfuse - super cool - thanks!

src/seer/automation/codegen/prompts.py
```diff
@@ -613,3 +613,34 @@ def format_prompt_followup(
...
+            </followups>
+
+            # Your goal:
+            Review the bug prediction analyses and extract the information listed below.
+
+            # Guidelines:
+            - Ensure all fields are properly populated based on each analysis.
+            - Use the following JSON format for each finding:
+            ```json
+            {{
```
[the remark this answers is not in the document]
suejung-sentry: Totally missed this great trick I see used elsewhere in this codebase - thanks for the tip! Really nice client library design by seer team to build that in 🧠

src/seer/automation/codegen/prompts.py
```diff
@@ -613,3 +613,34 @@ def format_prompt_followup(
...
+              "affected_files": string[], // List of filenames affected by this bug. Include the full path
+              "suggested_fix": string,    // Recommended approach to address the issue. Phrase it politely as a suggestion.
+              "severity": float,         // Severity level, from 0 (no issue) to 1 (critical issue)
+              "confidence": float,       // Confidence in prediction, from 0 (no confidence) to 1 (high confidence)
+              "is_valid": boolean,        // Whether this prediction should be shown to users
+              "code_locations": string[]  // Specific locations in the format of "/path/to/file.py:start_line~end_line". If the line numbers are unknown, just include the full path to the file.
+            }}
+            ```
+            """
+        ).format(followups="\n".join(followups))
```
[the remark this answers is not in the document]
suejung-sentry: good call - added!

src/seer/automation/codegen/models.py
```diff
@@ -367,6 +367,33 @@ def validate_lists_same_length(
...
+    affected_files: list[str]
+    suggested_fix: str
+    severity: float
+    confidence: float
+    is_valid: bool
+    code_locations: list[str]
+
+    def to_overwatch_result(self) -> RelevantWarningResult:
+        return RelevantWarningResult(
+            reasoning=self.description,
```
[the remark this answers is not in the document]
suejung-sentry: To document the convos in slack on these topics - 
We are scrapping the old model `RelevantWarningResult`.
In the new result model to overwatch we'll have `short_description` and `description`, where `description` is similar to the old "reasoning" (a summary of the `followup`). The actual raw much longer `followup` we can stash in the `diagnostics`

src/seer/automation/codegen/prompts.py
```diff
@@ -613,3 +613,34 @@ def format_prompt_followup(
...
+        return textwrap.dedent(
+            """
+            You are a helpful assistant that extracts structured information from bug prediction analyses.
+            You are given the following bug prediction analyses for a pull request.
+            <followups>
+            {followups}
+            </followups>
+
+            # Your goal:
+            Review the bug prediction analyses and extract the information listed below.
```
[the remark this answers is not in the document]
suejung-sentry: Good call - I added the `focus` to the `formatter`'s prompt. Also updated the prompt to only return the filtered outputs (instead of with `is_valid`).

I was on the fence about iterating over each followup vs. sending the whole list in one go. From my limited testing, I got the same results basically every time if I did one by one vs. sending it all at once.. so I think putting them together is fine and yeah has the added benefit of free deduping too. 

One interesting thing is when I was trying to get it to filter, if I gave it more than ~2 filtering related guidelines, it performed worse at the task (included more items than I wanted). So if we start seeing too many false positives, can add some more focus here. So far I don't think false positives is that high so should be at an okay balance of excessive calls to the LLM vs. getting good enough results for now...

src/seer/automation/codegen/prompts.py
```diff
@@ -613,3 +613,34 @@ def format_prompt_followup(
...
+
+            # Guidelines:
+            - Ensure all fields are properly populated based on each analysis.
+            - Use the following JSON format for each finding:
+            ```json
+            {{
+              "title": string,            // A concise summary of the bug prediction
+              "description": string,      // A detailed explanation of the potential issue. Phrase it politely as a suspicion.
```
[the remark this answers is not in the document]
suejung-sentry: Lol yeah I was pleasantly surprised that it worked. I tried a little both as command and question and it didn't look that different. I read around a bit on if there's an industry standard of what works better and it seems like what matters more is if you are clear - so yeah agreed can tweak it later as we go..

Per other slack convos, here's a [gist](https://gist.github.com/suejung-sentry/be7a2c3db1c2cedcc54c71261d264810) with the example on the `short_description` and `description` (summary of the `followup`) + we can add the entire raw followup in `diagnostics` or something (worst case will be there in Langfuse)

src/seer/automation/codegen/models.py
suejung-sentry: nice, thanks fixed!

src/seer/automation/codegen/prompts.py
suejung-sentry: nice thanks! good call. filtered out the None upstream so should be good by the point we reach here

tests/automation/codegen/fixtures/bug_prediction/1870.txt
suejung-sentry: nice fixed these to split it up. Currently thinking through if/how I want to represent this stuff in a test (vs. my exploratory scripting..)...

tests/automation/codegen/test_bug_prediction_formatter.py
```diff
@@ -0,0 +1,113 @@
...
+
+        for fixture_file in fixture_files:
+            fixture_id = os.path.splitext(fixture_file)[0]
+            fixture_path = os.path.join(fixture_dir, fixture_file)
+            with open(fixture_path, "r") as f:
+                raw_prediction = f.read()
+
+            expected_results = {
+                "1870": [
+                    {"is_valid": True, "expected_severity": 0.7},
```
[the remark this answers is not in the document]
suejung-sentry: Yeah that's a good point. I think erring on the side of reporting less / more conservatively makes sense. 
I added some filtering to the component to remove the non "crashy" ones (including this 1st one in 1870)

src/seer/automation/codegen/prompts.py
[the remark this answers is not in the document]
kddubey: > if I gave it more than ~2 filtering related guidelines, it performed worse at the task (included more items than I wanted)

that's a really good nugget of info. have had the same experience where instructing/constraining as little as possible can be better. appreciate you digging into this!

tests/automation/codegen/test_bug_prediction_formatter.py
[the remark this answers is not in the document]
suejung-sentry: Thanks for this info. I scrapped the attempt as it is kinda jarring to how everything else is done in this repo anyway. I can stick the fixtures I was using to play around with in a different place more suited to the exploration it was doing (maybe data-analysis repo)..

src/seer/automation/codegen/prompts.py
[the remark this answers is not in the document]
suejung-sentry: oh nice, interesting! that's a nice tidbit I don't see in generic prompt engineering guides, so good to keep in mind 💭

suejung-sentry (commented):
Here's a local end to end test run of this new endpoint, which looked not bad

```
curl --location 'http://localhost:9091/v1/automation/codegen/bug-prediction' \
--header 'Content-Type: application/json' \
--data '{
    "repo": {
        "provider": "github",
        "owner": "getsentry",
        "name": "seer",
        "external_id": "651733415"
    },
    "pr_id": 2128,
    "callback_url": "http://localhost:9091",
    "organization_id": 1,
    "warnings": [],
    "commit_sha": "e2e074687e6c7736cd70f4aa77b3e76ddaa472e9"
}
'
```

result - langfuse [url](https://langfuse-zelomd34wa-uc.a.run.app/project/clws5ue5q00006gkjq1d9c7l5/traces/36233cc3-0b46-476e-994c-d9d44c718a2c?timestamp=2025-05-19T17%3A33%3A52.734Z&observation=111ac57e-67b6-4e49-9901-a9c66ff37631) 

```
parsed: [
0: {
title: "Assertion in `_fetch_issues_for_pr_file` can cause server crash"
severity: 1
confidence: 0.9
description: "The `_fetch_issues_for_pr_file` function asserts that the keys in the dictionary returned by the RPC call exactly match the `pr_file.filename`. If the RPC service returns a dictionary with different keys (e.g., multiple keys, a similar but not identical key), the assertion will fail and raise an `AssertionError`, causing the server to crash. This occurs because there is no error handling for cases where the keys don't match the expected filename. For example, if `pr_file.filename` is 'file.py' and the RPC returns `{'different_file.py': [...]}` or `{'file.py': [...], 'other.py': [...]}`, the assertion will fail. The code `assert list(pr_filename_to_issues.keys()) == [pr_file.filename]` is the source of the crash."
suggested_fix: "Modify the assertion to handle cases where the RPC call returns a dictionary with different keys. Instead of asserting, add proper error handling or logging and return an appropriate value (e.g., an empty list) to prevent the server from crashing. Consider validating the contents of the dictionary instead of strictly enforcing the key."
encoded_location: "src/seer/automation/codegen/relevant_warnings_component.py"
short_description: "The assertion in `_fetch_issues_for_pr_file` may cause a server crash if the RPC call returns a dictionary with unexpected keys. Specifically, it expects a single key matching the PR file's filename."
}
]
```

src/seer/automation/codegen/models.py
```diff
@@ -95,10 +95,33 @@ def to_text_format(self) -> str:
...
 
 
+class BugPrediction(BaseModel):
+    title: str = Field(description="A concise summary title of the bug prediction. Max 10 words.")
+    short_description: str = Field(
+        description="A concise, technical explanation of the potential bug that identifies the specific risk and affected code. Phrase it politely as a suspicion. Max 75 words."
+    )
+    description: str = Field(
+        description="A detailed technical analysis of the bug that explains: 1) The exact failure mechanism, 2) The conditions that trigger it, 3) Why the current code is problematic, 4) Which specific code paths are affected, and 5) What could go wrong if this bug isn't fixed. Include concrete examples of variable states that would cause failure and ideally a short code snippet example. Max 350 words."
+    )
```
kddubey, line 105: did you see cases where the `followup` is missing any of 1-5 and the formatter hallucinates a bit to make up for it? rn the verifier agent isn't explicitly prompted for 1-5. tho one day we should look into making it do the more detailed deep dive written out here

src/seer/automation/codegen/prompts.py
```diff
@@ -576,3 +576,35 @@ def format_prompt_followup(
             hypothesis=hypothesis,
             focus=BugPredictionPrompts._focus_on_crashes(),
         )
+
+    @classmethod
```
kddubey, line 580: this is my bad but i changed these to `@staticmethod`

src/seer/automation/codegen/bug_prediction_component.py
```diff
@@ -209,3 +212,42 @@ def invoke(
...
+        self, request: FormatterRequest, llm_client: LlmClient = injected
+    ) -> FormatterOutput:
+        followups: list[str] = [
+            followup for followup in request.followups if followup is not None and followup != ""
+        ]
+        if not followups:
+            self.logger.info("No valid followups found to format into bug predictions")
+            return FormatterOutput(bug_predictions=[])
+
+        try:
```
kddubey, line 236: nit: any specific errors in this try-except that you anticipate? for context, the verify agent call is try-excepted b/c it can run for too many iterations, and it'll just fail for one of the bug hypotheses. here maybe the run should just fail. up to you since it depends on how we want to surface the PR comment in overwatch if seer fails

src/seer/automation/codegen/bug_prediction_step.py
```diff
@@ -96,19 +90,17 @@ def _post_results_to_overwatch(
...
+            self._post_results_to_overwatch(bug_predictions_output, diagnostics)
         except Exception:
             self.logger.exception("Error posting relevant warnings results to Overwatch")
             raise
         finally:
-            self.context.event_manager.mark_completed_and_extend_static_analysis_suggestions(
-                static_analysis_suggestions_output.suggestions
-                if static_analysis_suggestions_output
-                else []
+            self.context.event_manager.mark_completed_and_extend_bug_predictions(
```
kddubey, line 102: ty for going ahead w/ these changes 🙏

src/seer/automation/codegen/bug_prediction_component.py
```diff
@@ -209,3 +212,42 @@ def invoke(
...
+            )
+
+            if response.parsed is None:
+                self.logger.warning("Failed to extract structured information from bug prediction")
+                return FormatterOutput(bug_predictions=[])
+
+            return FormatterOutput(bug_predictions=response.parsed)
+
+        except Exception as e:
+            self.logger.error(f"Error formatting bug predictions: {e}")
```
kddubey, line 252:
```suggestion
            self.logger.exception("Error formatting bug predictions")
```

nit in case you wanna keep try-except

src/seer/automation/codegen/models.py
```diff
@@ -95,10 +95,33 @@ def to_text_format(self) -> str:
         return f"{self.short_description}\n{self.justification}\n<location>{self.path}:{self.line}</location>"
 
 
+class BugPrediction(BaseModel):
+    title: str = Field(description="A concise summary title of the bug prediction. Max 10 words.")
```
kddubey, line 99: nitty nit: it might be better to order this field last. gemini generates fields in the order you give it, so putting title last allows the model to have the description in context when it generates the title. prolly doesn't matter but worth knowing i guess haha

src/seer/automation/codegen/models.py
```diff
@@ -95,10 +95,33 @@ def to_text_format(self) -> str:
...
+        description="A concise, technical explanation of the potential bug that identifies the specific risk and affected code. Phrase it politely as a suspicion. Max 75 words."
+    )
+    description: str = Field(
+        description="A detailed technical analysis of the bug that explains: 1) The exact failure mechanism, 2) The conditions that trigger it, 3) Why the current code is problematic, 4) Which specific code paths are affected, and 5) What could go wrong if this bug isn't fixed. Include concrete examples of variable states that would cause failure and ideally a short code snippet example. Max 350 words."
+    )
+    suggested_fix: str = Field(
+        description="A short, fluff-free, information-dense explanation of the suggested fix. Max 100 words."
+    )
+    encoded_location: str = Field(
+        description="Specific locations in the format of '/path/to/file.py:start_line~end_line'. If the line numbers are unknown, just include the full path to the file."
```
kddubey, line 110: nit: is the leading slash fine? not sure what exactly overwatch expects

src/seer/automation/codegen/models.py
```diff
@@ -95,10 +95,33 @@ def to_text_format(self) -> str:
...
+    )
+    severity: float = Field(
+        description="From 0 to 1 how serious is this potential bug? Score this as if you are a developer who would have to respond to an incident caused by this bug. 1 being 'guaranteed production crash'"
```
kddubey, line 113: nice phrasing. it's implied but one thing to add might be scoring low if the bug requires very specific conditions to be met

another thing is to tell it that the score should be very granular, e.g., 0.432

kddubey (approved):
looks good. can take a quick look after the location stuff is passed through. otherwise just some small questions and comments

src/seer/automation/codegen/models.py
suejung-sentry: I didn't see any hallucinations in the ~dozen examples, and it generally did a best attempt at answering as many of those as possible (sometimes left some out). I found asking it for these specific things yielded a more tailored pr comment than asking it for a more general summary which many of them I read it and thought "huh??" haha.

The question of hallucinations is a good one though... might be something good to add to our evals measurements

src/seer/automation/codegen/prompts.py
suejung-sentry: oops sorry I missed that, thanks!

src/seer/automation/codegen/models.py
suejung-sentry: Oh I didn't know that about the ordering, thanks for the tip! Swapped short_description to go after description also

src/seer/automation/codegen/bug_prediction_component.py
suejung-sentry: My intention there was just a generic fallback (say llm client connection fails if pod goes down, or other edge case?) so we return empty results instead of leaving overwatch hanging.
I think we don't have a clear pattern right now for seer to communicate failures to overwatch... I'm good having the run just fail in this case - updated.

PR #2580, opened 2025-05-13
fix(autofix state): Save the branch and base_commit_sha to the autofix state

Mihir-Mavalankar:
# PR Details
+ This PR makes sure that the base_commit_sha and branch are saved to the database whenever an autofix run is created. 
+ Currently none of the autofix runs in prod (see run_state table) have a branch name or base_commit_sha since the autofix state is not updated with that information.
+ Locally run eval runs show a base commit (but null branch) since the dev database has those base commits in it already.
+ This PR addresses the first half of work in this ticket: [AIML-286](https://linear.app/getsentry/issue/AIML-286/fix-draft-pr-failing-due-to-changed-codebase) Sub task: [AIML-393](https://linear.app/getsentry/issue/AIML-393/make-sure-that-branch-and-base-commit-sha-are-being-saved-to-postgres)
+ Lastly, I have added comments to some classes and functions that helped me understand the codebase better. Open to changing the exact wording of some of these comments.


roaga (commented):
When drafting a PR, does it draft against the commit in the request?
If so, could you show that with a test or otherwise explain how?

src/seer/automation/autofix/state.py
```diff
@@ -15,5 +15,9 @@ def before_update(self, state: AutofixContinuation):
         state.mark_updated()
 
     def apply_to_run_state(self, state: AutofixContinuation, run_state: DbRunState):
+        """
+        This function updates AutofixContinuation state whihc is turn updates the updated_at and
```
jennmueng, line 19:
```suggestion
        This function updates AutofixContinuation state which in turn updates the updated_at and
```

tests/automation/autofix/test_runs.py
```diff
@@ -68,3 +68,36 @@ def test_create_initial_autofix_run_error_handling(self, mock_request, exception
...
+        mock_state.readable_repos = MagicMock(return_value=[mock_repo])
+
+        # Patch state.update() as a context manager
+        mock_update_cm = MagicMock()
+        mock_update_cm.__enter__.return_value = mock_state
+        mock_update_cm.__exit__.return_value = None
+
+        mock_continuation_state = MagicMock()
+        mock_continuation_state.get.return_value = mock_state
+        mock_continuation_state.update.return_value = mock_update_cm
```
jennmueng, line 97: We actually have a test db that's set up to be cleared for every test case, so for tests like these you can actually write and read to db to do a more comprehensive test

jennmueng (approved):
LGTM 🚢

src/seer/automation/autofix/models.py
```diff
@@ -319,6 +324,10 @@ class AutofixRequestOptions(BaseModel):
 
 
 class AutofixRequest(BaseModel):
+    """
+    This class maps to the JSON key "request" inside the "value" column of the run_state table.
+    """
+
```
roaga, line 330: i think these comments would be better explaining what these classes represent rather than what other parts of the code use them

src/seer/automation/autofix/runs.py
```diff
@@ -124,3 +134,23 @@ def set_accessible_repos(state: ContinuationState) -> None:
...
+    cur_state = state.get()
+    for repo in cur_state.readable_repos:
+        if not repo.branch_name:
+            repo_client = RepoClient.from_repo_definition(repo, "read")
+            repo.branch_name = repo_client.base_branch
+            if not repo.base_commit_sha:
+                repo.base_commit_sha = repo_client.base_commit_sha
+
+    with state.update() as cur:
+        cur.request.repos = cur_state.request.repos
```
roaga, line 156: could we do this inside `set_accessible_repos` if it's clean enough so we don't introduce extra database writes?

src/seer/automation/autofix/tools/tools.py
```diff
@@ -38,6 +38,11 @@
 
 
 class BaseTools:
+    """
+    Base class for tools that that can be used by any component/agent.
+    See usage examples in autofix/components/solution/component.py and autofix/autofix_agent.py
+    """
```
roaga, line 44: probably better to not cite other files in this class

src/seer/db.py
```diff
@@ -211,6 +211,11 @@ class DbSeerEvent(Base):
 
 
 class DbRunState(Base):
+    """
+    This is the schema of the run_state table that stores 1 row for every Autofix run.
+    The value field maps to a JSON column that has all the detailed information of the autofix state.
```
roaga, line 216: it's no longer just autofix i believe, codecov stores runs in this table too

src/seer/automation/autofix/models.py
Mihir-Mavalankar: Sounds good. I'll add more info about the actual class here.

src/seer/automation/autofix/runs.py
Mihir-Mavalankar: I'm okay with that. @jennmueng suggested a new function. @jennmueng what do you think?

src/seer/automation/autofix/tools/tools.py
Mihir-Mavalankar: It's to point to some examples on how to use this class. Is the concern that if these files are deleted then this comment will be stale?

src/seer/db.py
Mihir-Mavalankar: Will change.

src/seer/automation/autofix/tools/tools.py
roaga: Yeah that was my concern, but no strong opinion

src/seer/db.py
Mihir-Mavalankar: Done.

src/seer/automation/autofix/tools/tools.py
Mihir-Mavalankar: I think I'll stick with these 2 examples. We can revisit if this comment does become stale.

src/seer/automation/autofix/models.py
Mihir-Mavalankar: Done.

src/seer/automation/autofix/runs.py
jennmueng: we can combine and rename the function

Mihir-Mavalankar: Actually I think I have an argument to keep them separate. There is another function called `update_repo_access` that called `set_accessible_repos` and then that would have to be changed to along with it's declaration.

roaga: Is there any way to batch the db writing? I want to minimize the amount of time we have interacting with the db because we were having some db connection issues, plus high latency for starting autofix, recently due to over-using it

Mihir-Mavalankar: + That's a good call. Currently 3 different functions separately writing to the DB in create_initial_autofix_run. 
+ One option is that, I can remove the DB write from those functions and add them to the create_initial_autofix_run. This will involve changing the calls to all these 3 functions too.
+ That actually raises another issue too where the part so fht edb state are updated by different functions. This is generally not a good pattern. Ideally the whole state object should be written to the DB by 1 function.

Mihir-Mavalankar: Happy to sync on this too with the whole team.

Mihir-Mavalankar: Letting this extra DB call slide for now. Let's scope the work required to reduce DB calls throughout the app.

PR #2582, opened 2025-05-13
fix(autofix): Still make repo manager work without org id

jennmueng:
The codegen request doesn't have an org id and for some reason linter didn't throw a fit??


roaga (commented):
won't this still mean codecov can't download repos at all?

jennmueng:
> won't this still mean codecov can't download repos at all?

no, it just means GCS caching is disabled for them (and is disabled for everyone except autofix running on seer project anyways)

roaga (approved):

PR #2605, opened 2025-05-15
chore(assisted query): Update Langfuse tracing 

aayush-se:
- Add langfuse tracing to cache creation

**Question**: Should I be more granular here or should this generally be enough?

### DIFF
--- a/src/seer/automation/assisted_query/create_cache.py
+++ b/src/seer/automation/assisted_query/create_cache.py
@@ -32,7 +32,9 @@ REQUIRED_FIELD_PREFIXES = [
 @inject
 @observe(name="Create assisted query cache")
 @sentry_sdk.trace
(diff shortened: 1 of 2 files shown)


kddubey (approved):
looks good. can also consider logging or `@observe`ing the `client.call`s in case they're pretty sensitive

unrelated to the PR but am curious why the `LlmClient` is manually instantiated instead of using `@inject`

aayush-se:
> looks good. can also consider logging or `@observe`ing the `client.call`s in case they're pretty sensitive
> 
> unrelated to the PR but am curious why the `LlmClient` is manually instantiated instead of using `@inject`

ack to client calls 👍 

For the LlmClient...no idea should have caught that in the original PR let me fix and just include it as part of this PR

PR #2613, opened 2025-05-15
fix(autofix db writes): Merge functions to reduce DB calls

Mihir-Mavalankar:
# PR Details
+ This is a followup MR based on some comments in [2580](https://github.com/getsentry/seer/pull/2580).
+ This MR merges three functions in `create_initial_autofix_run` into one. This is done to reduce the numbers of DB writes we make from 3 to 1. 
+ Other usages of these functions have also been handled.


roaga (approved):
LGTM

PR #2619, opened 2025-05-16
feat(repo_manager): Pt.1 of Repo backfills, only for specific projects

jennmueng:
1. Adds a periodic celerybeat job to run repo backfill every 30 minutes\
 1.1. This will only run the backfill ONCE because we check if the project has gcs archives before we run
3. The collect repos for backfill is right now hardcoded to run on specific projects to see the efficacy
4. Implements a backfill job and backfill state table. These will be used later when we fully implement backfill.

**Additional Changes**
We now also limit both files Autofix sees and files Autofix downloads to 1MB, this should filter out most data dumps, minified code, etc that we shouldn't need.


src/seer/automation/codebase/repo_client.py
```diff
@@ -470,8 +470,10 @@ def _get_valid_file_paths(self, commit_sha: str | None = None) -> set[str]:
         valid_file_extensions = get_all_supported_extensions()
 
         for file in tree.tree:
-            if file.type == "blob" and any(
-                file.path.endswith(ext) for ext in valid_file_extensions
+            if (
+                file.type == "blob"
+                and any(file.path.endswith(ext) for ext in valid_file_extensions)
+                and file.size < 1024 * 1024  # 1MB
```
jennmueng, line 476: this is so the file list here aligns with what we download

src/seer/automation/codebase/repo_manager.py
```diff
@@ -83,27 +84,38 @@ def __init__(
...
+
+    @property
+    def blob_name(self):
+        return self.make_blob_name(
+            self.organization_id,
+            self.repo_client.provider,
+            self.repo_client.repo_owner,
+            self.repo_client.repo_name,
+            self.repo_client.repo_external_id,
+        )
```
roaga, line 117: does base commit sha matter?

src/seer/automation/codebase/tasks.py
```diff
@@ -0,0 +1,286 @@
...
+        )
+
+        # TODO: Implement backfill cursor
+        # ...
+
+        # start of hack #
+        # This logic below with the task_taken_key is basically because of how our celerybeat setup is.
+        # Every deployment of celerybeat will each schedule a task, so this is a hack to ensure that only one
+        # deployment runs the task at a time.
+        task_taken_key = backfill_state.task_taken_key if backfill_state else None
```
roaga, line 57: is this hack why we need a new DB model? Because otherwise if this was just for testing timing, i'm not understanding why backfills need a state

src/seer/automation/codebase/tasks.py
```diff
@@ -0,0 +1,286 @@
...
+
+@celery_app.task(
+    soft_time_limit=timedelta(minutes=1).total_seconds(),
+    time_limit=timedelta(minutes=1, seconds=10).total_seconds(),
+)
+def collect_all_repos_for_backfill():
+    logger.info("Collecting repos for backfill")
+
+    # TODO: Remove this once we have a real backfill cursor
+    backfill_project_ids = [1, 300688, 1320254, 5613870, 4507936016302080]
```
roaga, line 43: which projects are these? ('m assuming some belong to the big customers, but did you check that it's the project associated with their big repos and not their smaller repos?

roaga (approved):
going to approve to unblock your experimentation but i'm unclear if the DB state has any purpose beyond the hack

(also let's remember to clean this all up once we learn what we need to)

src/seer/automation/codebase/repo_manager.py
jennmueng: nope, should just be a blob per repo

src/seer/automation/codebase/tasks.py
jennmueng: we'll need to be storing the cursor, and also the hack

src/seer/automation/codebase/tasks.py
jennmueng: I mentioned in slack, pulled from bad langfuse traces

marcklingen: langfuse maintainer here, anything we can help with?

jennmueng: @marcklingen ah no, I was referring to failure cases seen in langfuse traces not langfuse itself failing haha

marcklingen: @jennmueng great, glad to hear this!

PR #2623, opened 2025-05-16
fix(autofix): Fix freezing LLM stream freezing the timeout check

jennmueng:
Handles the stream output in a separate "producer" thread for each llm text stream generator. This allows langfuse to still work as normal because the `@observe` decorator is outside the thread.

Tested locally with both pre-first token and post-first-token timeouts

CleanShot 2025-05-17 at 02 14 42@2x
CleanShot 2025-05-17 at 02 15 08@2x

### DIFF
--- a/src/seer/automation/agent/client.py
+++ b/src/seer/automation/agent/client.py
@@ -1,9 +1,11 @@
 import json
 import logging
+import queue
 import re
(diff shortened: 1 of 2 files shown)


roaga (approved):
🐐

PR #2625, opened 2025-05-16
fix(autofix): Default to regex in ripgrep

roaga:
Going back to defaulting to regex = true in ripgrep. My reasoning is that I see a lot of runs hitting max iterations, where they try ripgrep with regex with fixed strings = true, get no results, and get confused. We do have a retry for failed regex so that other direciton should be more reliable.

Also removing the prompt that tells the agent to primarily use ripgrep. So it hopefully falls back on other tools too.

These changes are all speculation, so feel free to shoot down this PR

### DIFF
--- a/src/seer/automation/autofix/tools/tools.py
+++ b/src/seer/automation/autofix/tools/tools.py
@@ -503,7 +503,7 @@ class BaseTools:
         exclude_pattern: str | None = None,
         case_sensitive: bool = False,
         repo_name: str | None = None,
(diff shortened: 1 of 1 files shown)


jennmueng:
When we originally had regex by default didn't we see a lot of keyword searches fail the regex because it wasn't escaped correctly? My anecdotal feel is that although it errors often it should work for the majority of cases the current way rather than what this PR is right?

PR #2628, opened 2025-05-17
fix(repo client): Fix get git tree

roaga:
The method to get the full git tree was broken on large repos. This was because when building the tree recursively, turns out tree items from that point only return paths relative to the subdirectory you made the request for. We needed to reconstruct the full path from the root of the repo. Previous testing of this method only compared the count of files, so this slipped through.

This should fix the bug where expand document (and other things like autocorrect, stack trace validation, and more) were breaking on large repos 🙏 

Tested with a script that compared git tree results and cloned repo results. Tried on `getsentry/seer` and `torvalds/linux`.


jennmueng (approved):
🔥

PR #2637, opened 2025-05-19
chore(anomaly-detection): separate out the AD null value error to track it separately with lower priority

ram-senth:
The value error is too generic and overloaded as any of our end points can throw this error without any context. Also, the error from anomaly detection endpoint is high volume and low priority. So separating it out.

### DIFF
--- a/src/seer/anomaly_detection/models/external.py
+++ b/src/seer/anomaly_detection/models/external.py
@@ -20,7 +20,7 @@ class Anomaly(BaseModel):
 
 class TimeSeriesPoint(BaseModel):
     timestamp: float
-    value: float
+    value: Optional[float]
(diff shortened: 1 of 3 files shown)


src/seer/anomaly_detection/models/external.py
```diff
@@ -20,7 +20,7 @@ class Anomaly(BaseModel):
 
 class TimeSeriesPoint(BaseModel):
     timestamp: float
-    value: float
+    value: Optional[float]
```
aayush-se, line 23: qq: does value being `None` here have any consequences in other parts of the codebase? ie: are there any checks on Value that could be impacted by it now being None?

aayush-se (approved):

PR #2640, opened 2025-05-19
chore(autofix): Don't log error when creating file

roaga:
When creating a file via a tool, we called get_file_content to validate that the file didn't already exist. This had the disadvantages of:
1. Making an API call
2. Throwing an error to Sentry

This PR switches to get_valid_file_paths, which should already be cached in the coding step.

### DIFF
--- a/src/seer/automation/autofix/tools/tools.py
+++ b/src/seer/automation/autofix/tools/tools.py
@@ -368,12 +368,6 @@ class BaseTools:
 
         return None
 
-    def _does_file_exist(self, path: str, repo_name: str) -> bool:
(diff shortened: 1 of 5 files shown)


src/seer/automation/autofix/tools/tools.py
```diff
@@ -368,6 +368,12 @@ def _attempt_fix_path(self, path: str, repo_name: str, files_only: bool = False)
 
         return None
 
+    def _does_file_exist(self, path: str, repo_name: str) -> bool:
```
kddubey, line 371: does it make more sense for this to be a `RepoClient` method?

kddubey (approved):

PR #2642, opened 2025-05-19
chore(trace-queries): Support multiple options

aayush-se:
- Workflow returns multiple options for the user to select
  - Returns in a list
  - Updated prompts
- Please merge after: [#91916](https://github.com/getsentry/sentry/pull/91916)

### DIFF
--- a/src/seer/automation/assisted_query/assisted_query.py
+++ b/src/seer/automation/assisted_query/assisted_query.py
@@ -13,6 +13,7 @@ from seer.automation.assisted_query.models import (
     RelevantFieldsResponse,
     TranslateRequest,
     TranslateResponse,
+    TranslateResponses,
 )
 from seer.automation.assisted_query.utils import get_cache_display_name, get_model_provider
 from seer.dependency_injection import inject, injected
@@ -23,8 +24,9 @@ logger = logging.getLogger(__name__)
 REQUIRED_FIELDS = ["span.op", "span.description", "transaction"]
 
 
-def translate_query(request: TranslateRequest) -> TranslateResponse:
-
+def translate_query(
+    request: TranslateRequest,
+) -> TranslateResponses:
     natural_language_query = request.natural_language_query
     org_id = request.org_id
     project_ids = request.project_ids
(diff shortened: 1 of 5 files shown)


kddubey (approved):
> le bug predictor thinks it might be a problem if (i'd think in the super rare case) the structured output is an empty list, since the sentry-side change [has this line](https://github.com/getsentry/sentry/pull/91916/files#diff-476ad3bb07e4eb69eb871d680bf8c49d10afafd8e9072a375dd9e20890e27b06R106)

Bug predictor was right!...but kind of indirectly lol

> should the prompt ask for generating 1-3 instead of just max 3? and if the empty list case is unhandled, maybe should handle on sentry side

This is a good prompting call and good call to handle it on the sentry side. Im thinking just to set some sort of default?

Screenshot 2025-05-20 at 4 58 30 PM

kddubey:
lol i just ran it again and in the same output it said it isn't a problem 😞

fyi am playing w/ [this notebook](https://github.com/getsentry/data-analysis/blob/main/error_prevention/demo.ipynb)

kddubey:
> Im thinking just to set some sort of default?

up to you ofc but it doesn't seem unreasonable to error out and say Seer couldn't interpret your query or something like that. and log error on seer/sentry side too for further study

src/seer/automation/assisted_query/assisted_query.py
```diff
@@ -112,19 +115,17 @@ def create_query_from_natural_language(
...
+    final_query_prompt = prompts.get_final_query_prompt(
         natural_language_query, relevant_fields, field_values
     )
     generated_query = llm_client.generate_structured(
-        prompt=fields_and_values_prompt,
+        prompt=final_query_prompt,
         model=model,
         cache_name=cache_name,
-        response_format=ModelResponse,
+        response_format=list[ModelResponse],
```
kddubey, line 126: had another thought. (was thinking about this for bug pred and realized it might be worth mentioning here)

if the goal is to generate a set of candidate queries that are somewhat unique (maybe b/c the NL query is too underspecified), it makes sense to prompt it for a list and surface these. though if the goal is more around accuracy, sampling multiple candidates independently, i.e., w/o plugging in each as context for the next, and then picking a consensus (somehow) is a good self-consistency approach

aayush-se: Yeah I think its an interesting case because both are valid goals for this with the latter probably being more important by like 20%.

For the latter, if I understand correctly, I did try this (though not exhaustively) for generating multiple options in one step, then having a final judge step select the best and/or combine the best features of each. 

Unless its more on the lines of running n translation requests independently then having a judge pick the consensus? That would be really cool but I assume would need to run these steps in parallel to save on latency and figure out costs somehow

kddubey: > That would be really cool but I assume would need to run these steps in parallel to save on latency and figure out costs somehow

supplying `candidate_count > 1` should do this. but given your initial experiment, agree this one is prolly low priority

PR #2652, opened 2025-05-21
feat(bug-prediction-evals): Allow custom commit_sha

kddubey:
a custom `commit_sha`—representing a point before the PR branch head—is useful for evaluating PRs which had a bug in the middle of the review process. this lets us add more buggy PRs to our eval dataset, w/ the only caveat that we technically weren't certain that the PR caused a prod problem. we're just trusting our knowledge about the PR, which should be fine

### DIFF
--- a/src/seer/automation/codegen/bug_prediction_step.py
+++ b/src/seer/automation/codegen/bug_prediction_step.py
@@ -104,17 +104,28 @@ class BugPredictionStep(CodegenStep):
             )
 
     @observe(name="Codegen - Bug Prediction Step")
-    def _invoke(self, **kwargs) -> None:
+    def _invoke(self, app_config: AppConfig = injected, **kwargs) -> None:
         self.logger.info("Executing Codegen - Bug Prediction Step")
         self.context.event_manager.mark_running()
 
         # 1. Read the PR.
         if self.request.repo.base_commit_sha is None:
+            # Overwatch currently passes the commit_sha in the request.
+            # Need to set this on the repo definition so that code search happens at this commit.
             self.request.repo.base_commit_sha = self.request.commit_sha
         repo_client = self.context.get_repo_client(
             repo_name=self.request.repo.full_name, type=RepoClientType.READ
         )
+
         pr = repo_client.repo.get_pull(self.request.pr_id)
+        pr_head_sha = repo_client.get_pr_head_sha(pr.url)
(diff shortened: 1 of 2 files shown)


src/seer/automation/codegen/bug_prediction_step.py
```diff
@@ -104,17 +104,28 @@ def _complete_run(
...
+            # Need to set this on the repo definition so that code search happens at this commit.
             self.request.repo.base_commit_sha = self.request.commit_sha
         repo_client = self.context.get_repo_client(
             repo_name=self.request.repo.full_name, type=RepoClientType.READ
         )
+
         pr = repo_client.repo.get_pull(self.request.pr_id)
+        pr_head_sha = repo_client.get_pr_head_sha(pr.url)
+        if (pr_head_sha != self.request.commit_sha) and not app_config.is_production:
+            # This should only be used when we're evaluating, which happens locally.
```
kddubey, line 123: it's possible for this to happen in prod if a change is pushed while overwatch sets up and sends the request to seer. in that case, i think it probably makes sense to still look at the latest changes / trigger the `else`? minor but wanted to point this out

src/seer/automation/codegen/bug_prediction_step.py
```diff
@@ -105,16 +105,27 @@ def _complete_run(
...
+            # Overwatch currently passes the commit_sha in the request.
+            # Need to set this on the repo definition so that code search happens at this commit.
             self.request.repo.base_commit_sha = self.request.commit_sha
         repo_client = self.context.get_repo_client(
             repo_name=self.request.repo.full_name, type=RepoClientType.READ
         )
+
         pr = repo_client.repo.get_pull(self.request.pr_id)
+        pr_head_sha = repo_client.get_pr_head_sha(pr.url)
+        if pr_head_sha == self.request.commit_sha:
```
suejung-sentry, line 124: would pr_head_sha ever be None if some problem fetching/determining it? can add a condition to protect against that here if so

suejung-sentry (approved):

src/seer/automation/codegen/bug_prediction_step.py
kddubey: i think no? it's typed as `str` and looks like existing uses in our codebase assume it's always a string

kddubey: good to keep an eye on this tho

PR #2655, opened 2025-05-21
feat(repo-manager): Scaling backfill time limit to repo size

jennmueng:
Scale the time limit for a repo according to it's size on metadata. So for example, a 5GB repo would have a `15 + (5 - 1) * 7 = 43 min` timeout. Any repo less than 1GB would have the default 15 minute timeout.


roaga (approved):
😭

PR #2656, opened 2025-05-21
fix(autofix): Handle view directory in claude tools

roaga:
If the claude view command was called with a directory path, our check was not working correctly. This should fix it and return the tree for the path. Should fix [SEER-145](https://sentry.sentry.io/issues/6593399718/).

Tacking on: modify `get_file_contents` in `AutofixContext` to not unnecessarily call GitHub if the file is a locally created file. Should fix [SEER-EB](https://sentry.sentry.io/issues/6045489887/).

### DIFF
--- a/src/seer/automation/autofix/tools/tools.py
+++ b/src/seer/automation/autofix/tools/tools.py
@@ -342,7 +342,12 @@ class BaseTools:
         return f"<did you mean>\n{joined}\n</did you mean>"
 
     def _attempt_fix_path(
-        self, path: str, repo_name: str, files_only: bool = False, ignore_local_changes: bool = True
+        self,
+        path: str,
+        repo_name: str,
+        files_only: bool = False,
+        directories_only: bool = False,
+        ignore_local_changes: bool = True,
     ) -> str | None:
         """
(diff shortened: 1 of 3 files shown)


src/seer/automation/autofix/tools/tools.py
```diff
@@ -836,12 +842,20 @@ def _handle_view_command(
...
+            if not does_file_exist:
+                directory_path = self._attempt_fix_path(
+                    path, repo_name, directories_only=True, ignore_local_changes=False
+                )
+                if directory_path:
+                    if view_range:
+                        return "Error: Cannot view a directory with a line range."
+                    return self.tree(directory_path, repo_name)
+                else:
+                    return "Error: File not found."
```
jennmueng, line 858: Would it not be cheaper to just do the `os.path.isdir(path)` first before we even try the file existence and path fixing/

roaga: I don't think `os.path.isdir` works if the repo is not downloaded, right?

jennmueng: oh, that makes sense

jennmueng (approved):

PR #2657, opened 2025-05-21
[ft] Add overwatch request endpoint

rohitvinnakota-codecov:
This will enable AI PR review (and eventually unit test gen) for Sentry Prevent. 

The goal here is to support 2 apps, Codecov AI, and Autofix(Seer by Sentry). `Overwatch`, which is already setup to communicate with seer, will send requests to initiate reviews and unit tests by hitting this endpoint. 

Post migration on Codecov's side, we can remove the `codecov-request` endpoint and use `overwatch-request` for all our needs. A refactoring effort will also be required to change all our variable and class names at that time. 

Screenshot 2025-05-21 at 10 29 06 AM

### DIFF
--- a/src/seer/automation/codegen/pr_review_step.py
+++ b/src/seer/automation/codegen/pr_review_step.py
@@ -21,6 +21,7 @@ from seer.automation.state import DbStateRunTypes
 class PrReviewStepRequest(PipelineStepTaskRequest):
     pr_id: int
     repo_definition: RepoDefinition
+    is_codecov_request: bool
 
 
 @celery_app.task(
@@ -55,8 +56,14 @@ class PrReviewStep(CodegenStep):
         self.logger.info("Executing Codegen - PR Review Step")
         self.context.event_manager.mark_running()
 
+        client_type = (
+            RepoClientType.CODECOV_PR_REVIEW
+            if self.request.is_codecov_request
+            else RepoClientType.WRITE  # WRITE is the autofix app
+        )
+
         repo_client = self.context.get_repo_client(
-            repo_name=self.request.repo_definition.full_name, type=RepoClientType.CODECOV_PR_REVIEW
+            repo_name=self.request.repo_definition.full_name, type=client_type
         )
         pr = repo_client.repo.get_pull(self.request.pr_id)
         diff_content = repo_client.get_pr_diff_content(pr.url)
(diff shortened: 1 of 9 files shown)


src/seer/automation/codegen/pr_review_step.py
```diff
@@ -21,6 +21,7 @@
 class PrReviewStepRequest(PipelineStepTaskRequest):
```
kddubey, line 21: seems like [`CodegenPrReviewRequest`](https://github.com/getsentry/seer/blob/1ddb32a2513e2e6b77ec030c5cbe52836b12286d/src/seer/automation/codegen/models.py#L153) is used in other places but not here, maybe leading to awkward problems

what do you think of adding `is_codecov_request` to `CodegenPrReviewRequest` and then changing this to—

```python
class PrReviewStepRequest(PipelineStepTaskRequest, CodegenPrReviewRequest):
    pass
```

—so that `CodegenPrReviewRequest` and `CodegenBaseRequest` are re-used?

src/seer/configuration.py
```diff
@@ -72,7 +72,7 @@ class AppConfig(BaseModel):
     LANGFUSE_HOST: str = ""
 
     API_PUBLIC_KEY_SECRET_ID: str = ""
-    JSON_API_SHARED_SECRETS: ParseList = Field(default_factory=list)
+    JSON_API_SHARED_SECRETS: ParseList = ["some-secret-to-seer", "some-secret-from-seer"]
```
kddubey, line 75: if this is for local env to work, it's prolly a safer change to leave this as is and instead add instructions to add this to `.env`. can also add this in `.env.example`

src/seer/automation/codegen/tasks.py
```diff
@@ -141,7 +141,9 @@ def get_unittest_state(request: CodegenUnitTestsStateRequest):
 
 
 @inject
-def codegen_pr_review(request: CodegenBaseRequest, app_config: AppConfig = injected):
+def codegen_pr_review(
+    request: CodegenBaseRequest, app_config: AppConfig = injected, is_codecov_request: bool = True
```
kddubey, line 145: nit: why's it a `CodegenBaseRequest` instead of `CodegenPrReviewRequest`? then can re-use `request.is_codecov_request`, which defaults to `True` in the model

kddubey (approved):

src/seer/automation/codegen/pr_review_step.py
rohitvinnakota-codecov: `is_codecov_request` is something we should check on all `Codecov` requests(pr reviews, unit tests, pr closed) when they are added. 

I think I will add `is_codecov_request` directly to the base request as a value that is passed in from overwatch so we can reuse the classes at that point, but I prefer to refactor once we have all workflows in overwatch. I agree it's not exactly clean at the moment.

PR #2661, opened 2025-05-22
feat(autofix): Prompt engineering and new solution architecture

roaga:
Much more robust prompts and a solution architecture that is just 1 step using gemini 2.5 pro instead of flash + claude 3.7. Also bumping the 2.5 flash preview checkpoint. 

Root cause ~87%
Solution ~62%
(so stable, or at least a return to form)

& Faster run time

### DIFF
--- a/src/seer/automation/autofix/components/root_cause/prompts.py
+++ b/src/seer/automation/autofix/components/root_cause/prompts.py
@@ -1,12 +1,7 @@
 import textwrap
 from typing import Literal
 
-from seer.automation.autofix.prompts import (
-    format_code_map,
-    format_instruction,
-    format_summary,
-    format_trace_tree,
-)
+from seer.automation.autofix.prompts import format_code_map, format_summary, format_trace_tree
 from seer.automation.models import Profile, TraceTree
 from seer.automation.summarize.issue import IssueSummary
 
@@ -65,7 +60,7 @@ class RootCauseAnalysisPrompts:
     ):
         return textwrap.dedent(
             """\
-            Please begin by gathering all relevant context to understand what happened, from the entry point of the code to the error, and WHY it happened. {instruction_str}I have included everything I know about the Sentry issue so far below:
+            Please begin by gathering all relevant context to understand what happened, from the entry point of the code to the error, and WHY it happened. {instruction_str} I have included everything I know about the Sentry issue so far below:
 
             <issue_details>
             {summary_str}
@@ -76,7 +71,7 @@ class RootCauseAnalysisPrompts:
             </issue_details>"""
         ).format(
             error_str=event,
-            instruction_str=format_instruction(instruction),
+            instruction_str=instruction,
             summary_str=format_summary(summary),
             code_map_str=format_code_map(code_map),
             trace_tree_str=format_trace_tree(trace_tree),
--- a/src/seer/automation/autofix/components/root_cause/prompts.py
+++ b/src/seer/automation/autofix/components/root_cause/prompts.py
@@ -1,4 +1,5 @@
 import textwrap
+from typing import Literal
 
 from seer.automation.autofix.prompts import (
     format_code_map,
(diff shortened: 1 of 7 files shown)


src/seer/automation/autofix/components/root_cause/prompts.py
```diff
@@ -12,49 +13,69 @@
...
+            Now, you must lead the effort to find the root cause of the ISSUE.
 
-            # Guidelines:
+            <tool_calling>
+            {tool_calling_str}
+            </tool_calling>
+
+            <available_repos>
+            {repos_str}
+            </available_repos>
```
aayush-se, line 30: Might be good to pull these to the bottom of the prompt to make use of implicit prompt caching.

src/seer/automation/autofix/components/root_cause/prompts.py
```diff
@@ -12,49 +13,69 @@
...
         code_map: Profile | None = None,
         trace_tree: TraceTree | None = None,
     ):
         return textwrap.dedent(
             """\
-            <goal>{explore_msg}</goal>
-            <available_repos>
-            {repos_str}
-            </available_repos>
+            Please begin by gathering all relevant context to understand what happened, from the entry point of the code to the error, and WHY it happened. {instruction_str}I have included everything I know about the Sentry issue so far below:
```
aayush-se, line 68: nit: space or new line after `{instruction_str}`

src/seer/automation/autofix/tools/tools.py
```diff
@@ -1136,7 +1136,7 @@ def get_tools(
                     FunctionTool(
                         name="grep_search",
                         fn=self.run_ripgrep,
-                        description="Runs a ripgrep command over the codebase to find what you're looking for. Use this as your main tool for searching codebases. Use the include and exclude patterns to narrow down the search to specific paths or file types.",
```
aayush-se, line 1139: Curious for this removal. Was it overusing it resulting in longer times?

roaga: i saw a lot of traces where it was just spamming ripgrep and forgetting all other tools. Hoping this reduces that but no good way to evaluate yet.

src/seer/automation/autofix/components/solution/prompts.py
```diff
@@ -43,24 +63,16 @@ def format_root_cause(root_cause: RootCauseAnalysisItem | str):
...
         trace_tree: TraceTree | None,
     ):
         return textwrap.dedent(
             """\
-            <goal>Gather all information that may be needed to understand how to fix this issue at its root in the way the developer would most likely want to fix it. (don't worry about proposing the fix itself yet)</goal>
-
-            <available_repos>
-            {repos_str}
-            </available_repos>
+            Please begin by gathering all relevant context to understand how to fix the issue. {original_instruction}I have included everything I know about the Sentry issue so far below:
```
aayush-se, line 73: same space/newline after `{original_instruction}`

src/seer/automation/autofix/components/solution/prompts.py
```diff
@@ -2,35 +2,55 @@
...
+                if has_tools
+                else "You do not have to ability to gather more context at this point. You must come up with the best solution you can based on what you know so far."
+            ),
+            repos_str=repos_str,
         )
 
     @staticmethod
     def format_original_instruction(instruction: str):
-        return f"Earlier, the user provided context: {format_instruction(instruction)}"
+        return instruction
```
aayush-se, line 53: What is the point of this function if it's just returning the parameter untouched?

src/seer/automation/autofix/components/root_cause/prompts.py
roaga: Eh I don't think the impact will be too high, plus these are static over the course of a run

src/seer/automation/autofix/tools/tools.py
aayush-se: Might be nice to include how many times different tools were called over the course of a run? Languse traces might already be able to do that though

roaga:
@aayush-se previous baseline is unclear because we saw a drop recently without any changes. It was 65% solution at my last eval, but then recently seemed to have dropped to 50%. So I'm just gonna call it "stable"

PR #2662, opened 2025-05-22
Set seed for OAI, Gemini

kddubey:
this [isn't guaranteed](https://platform.openai.com/docs/advanced-usage#reproducible-outputs) to result in reproducible outputs. but it should help. can make evals of prompt-completion tasks like issue summary more reproducible

side benefit that editing VCR tests might be less painful


aayush-se (approved):

aayush-se:
Is there any benefit of setting a seed in prod or is it mostly useful for evals/vcrs?

kddubey:
> Is there any benefit of setting a seed in prod or is it mostly useful for evals/vcrs?

hard to say concretely. should be used whenever it's a good UX for the output to be reproducible

PR #2664, opened 2025-05-22
feat(autofix): Receive logs from sentry in initial request

roaga:
Receives logs from `sentry` in the initial Autofix request. Unused for now as I want to make sure it works. Then I'll follow up and include them in the prompt

### DIFF
--- a/src/seer/automation/models.py
+++ b/src/seer/automation/models.py
@@ -980,6 +980,62 @@ class TraceTree(BaseModel):
         return list(set(event.project_id for event in self._get_all_events() if event.project_id))
 
 
+class Log(BaseModel):
+    message: str | None = None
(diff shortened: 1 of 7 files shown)


src/seer/automation/models.py
```diff
@@ -980,6 +980,62 @@ def get_all_project_ids(self) -> list[int]:
...
+
+        # sort logs by timestamp in descending order (assumes timestamp is in ISO format)
+        sorted_logs = sorted(
+            self.logs,
+            key=lambda log: log.timestamp or "",
+            reverse=True,
+        )
+
+        lines = []
+        for log in sorted_logs[:50]:
```
aayush-se, line 1008: Should this limit of 50 be a configurable param

PR #2668, opened 2025-05-22
chore(autofix): AIML 476 - Added a created_at column to the run_state table

Mihir-Mavalankar:
## PR Details
+ This PR creates a new created at column in the run state table.
+ This value is set at the start of each autofix run.
+ I also used `datetime.datetime.now(datetime.UTC)` since `datetime.datetime.utcnow` is deprecated.


src/seer/automation/autofix/models.py
```diff
@@ -637,6 +640,9 @@ def get_selected_solution(
                     return solution_step.solution, solution_step.selected_mode
         return None, None
 
+    def mark_created(self):
+        self.created_at = datetime.datetime.now()
```
roaga, line 644: isn't this unnecessary if the field auto-populates?

src/migrations/versions/4a87188db505_migration.py
```diff
@@ -0,0 +1,32 @@
...
+down_revision = "9ac60c7534ce"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("run_state", schema=None) as batch_op:
+        batch_op.add_column(sa.Column("created_at", sa.DateTime(), nullable=False))
+
```
roaga, line 23: if the column is not nullable, what will the value of the column be when migrating all our existing runs?

Mihir-Mavalankar: Changed that.

src/seer/automation/autofix/models.py
Mihir-Mavalankar: True. Changing it.

Mihir-Mavalankar: Removed it.

Mihir-Mavalankar: I am setting created_at just once in create_initial_autofix_run now. So that's the only place it can be set.

roaga: What I mean is, can't you just set a default value in the db model? so when a new row is created, it auto-populates the field. You never need to manually mark it

src/migrations/versions/4a87188db505_migration.py
roaga: It's still marked as not nullable in the db model, right?

Mihir-Mavalankar: Ya I can do that. i have set it to current date time right now. But last_triggered seems more accurate.

Mihir-Mavalankar: Made that change.

src/seer/automation/autofix/models.py
Mihir-Mavalankar: So I am a little fuzzy on initialization of DbRunState objects:
+ If I keep a default value for created_at every time DbRunState object is created it will set the created_at as the current timestamp.
+ But this is not the behaviour we want. We want it set once at the start of each run and then everytime it should be read from the postgres Db.

Mihir-Mavalankar: For example, every time this line `db_state = DbRunState(id=self.id, value=value.model_dump(mode="json"))` is called in the `update` function won't it update the created_at time stamp?

roaga: My understanding is that when you create a new row, it will populate the column with the default value. In future updates, the column will not change because it already has a value.

roaga: Feel free to test it out though, I might be wrong

src/seer/automation/autofix/models.py
```diff
@@ -272,6 +272,7 @@ class AutofixGroupState(BaseModel):
     updated_at: Annotated[datetime.datetime, Examples(datetime.datetime.now() for _ in gen)] = (
         Field(default_factory=datetime.datetime.now)
     )
+    created_at: datetime.datetime | None = None
```
jennmueng, line 275: is this correct? because `created_at` here would map to a key in the json inside the `value` column of the table not the `created_at` column we are creating right?

Mihir-Mavalankar: Okay I'll remove created_at from AutofixGroupState since it is only set once and doesn't need to synced like updated_at or last_triggered_at.

Mihir-Mavalankar: If this looks good I'll merge @jennmueng

jennmueng (approved):

PR #2669, opened 2025-05-23
fix(bug-prediction): Adjust evals for bug-prediction feature

suejung-sentry:
Update our existing Evals used for relevant-warnings to instead point to the new bug predictor. This is done with breaking changes since we won't be using the old one anymore.

Closes https://linear.app/getsentry/issue/PREVENT-76/seer-update-evals-to-match-format-of-the-new-bug-predictor


src/seer/automation/codegen/evals/tasks.py
```diff
@@ -96,6 +97,10 @@ def run_relevant_warnings_evaluation_on_item(
         logger.error(f"Item {item_id} not found or not active. Skipping scoring.")
         return
 
+    if dataset_item.status == DatasetStatus.ARCHIVED:
+        logger.error(f"Item {item_id} is archived. Skipping scoring.")
```
kddubey, line 101: ty for spotting this

kddubey (approved):

PR #2674, opened 2025-05-24
feat(autofix): Receive span evidence from issues

roaga:
Adds span evidence model on autofix and issue summary. Should allow us to operate on most performance issues. 

Follow ups:
- Test in prod with `yarn dev-ui`
- Enable UI on frontend
- Enable automation

### DIFF
--- a/src/seer/automation/models.py
+++ b/src/seer/automation/models.py
@@ -407,7 +407,9 @@ class EventDetails(BaseModel):
         )
 
         title = event.get("title")
-        if issue_title and title and title not in issue_title:
+        if (
+            issue_title and title and title not in issue_title and len(spans) > 0
+        ):  # if it's a performance issue with spans, use the issue title as well if not a duplicate
             title = f"{issue_title} - {title}"
 
         return cls(
--- a/src/seer/automation/models.py
+++ b/src/seer/automation/models.py
@@ -407,7 +407,7 @@ class EventDetails(BaseModel):
         )
 
         title = event.get("title")
-        if issue_title and title not in issue_title:
+        if issue_title and title and title not in issue_title:
             title = f"{issue_title} - {title}"
 
         return cls(
--- a/src/seer/automation/models.py
+++ b/src/seer/automation/models.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import re
 import textwrap
 from typing import Annotated, Any, Literal, NotRequired, Optional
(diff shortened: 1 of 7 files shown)


src/seer/automation/models.py
```diff
@@ -386,18 +402,24 @@ def from_event(cls, error_event: SentryEventData):
                 "has_transaction_name": transaction_name is not None,
                 "has_http_request": request is not None,
                 "has_error_message": message is not None,
+                "has_spans": spans and len(spans) > 0,
             }
         )
 
+        title = event.get("title")
+        if issue_title and title and title not in issue_title:
+            title = f"{issue_title} - {title}"
```
roaga, line 411: for performance issues, the issue title is very important and different than the event title. For error issues, they are usually identical

jennmueng: I wonder if for error issues where the titles differ a little will this cause a weird looking title where we're mostly repeating the same thing?

Something like
"Error: foo 4 in bar - Error: foo 5 in bar"

src/seer/automation/models.py
```diff
@@ -586,6 +613,62 @@ def format_request(self):
...
+                # Recurse into children
+                children = parent_to_children.get(span.span_id, [])
+                if children:
+                    format_span_tree(children, data_prefix, lines, is_last)
+            return lines
+
+        lines = ["Relevant Spans"]
+        lines.extend(format_span_tree(roots))
+        return "\n".join(lines)
+
```
jennmueng, line 671: should we do some try-catching here to make sure we're not erroring in prod?

src/seer/automation/models.py
```diff
@@ -465,6 +491,7 @@ def format_event_without_breadcrumbs(
                 f"<stacktraces>\n{stacktraces}\n</stacktraces>" if stacktraces.strip() else ""
             ),
             message=f"<message>\n{message}\n</message>" if message.strip() else "",
+            spans=f"<relevant_spans>\n{spans}\n</relevant_spans>" if spans.strip() else "",
```
jennmueng, line 494: Because we're doing

```
        if not self.spans:
            return "No spans available"
```
We'd always be saying `<relevant_spans>\nNo spans available\n</relevant_spans>` right if there aren't any spans rather than leaving it blank, is this intended?

src/seer/automation/models.py
roaga: Added a try except

PR #2682, opened 2025-05-28
feat(repo_manager): 30 day TTL for cached repos

jennmueng:


src/seer/app.py
```diff
@@ -339,6 +343,17 @@ def autofix_sync_start_endpoint(data: AutofixNoopRequest) -> AutofixEndpointResp
     return AutofixEndpointResponse(started=True, run_id=-1)
 
 
+@json_api(blueprint, "/v1/automation/autofix/cache-ttl/start")
+def autofix_cache_ttl_endpoint(data: AutofixNoopRequest) -> AutofixEndpointResponse:
```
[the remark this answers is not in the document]
jennmueng: This endpoiint is only for local development, it should error if called outside dev. only the celerybeat scheduler calls the underlying tsk.

src/migrations/versions/b27f28d90e1a_archive_last_downloaded_at.py
```diff
@@ -0,0 +1,32 @@
...
+revision = "b27f28d90e1a"
+down_revision = "ac95ebfa16cb"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("seer_repo_archive", schema=None) as batch_op:
+        batch_op.add_column(sa.Column("last_downloaded_at", sa.DateTime(), nullable=True))
```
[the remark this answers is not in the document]
jennmueng: For repos archives that have never been downloaded. Such as a backfill task timed out before it downloads

src/seer/automation/codebase/tasks.py
```diff
@@ -634,3 +641,99 @@ def run_repo_sync_for_repo_archive(repo_sync_job_dict: dict):
...
+            # 3. For each archive, use RepoManager to delete both GCS blob and database record
+            for repo_archive in repo_archives:
+                try:
+                    # Create a RepoClient and RepoManager for the repository
+                    repo_client = RepoClient.from_repo_definition(
+                        RepoDefinition.model_validate(repo_archive.repo_definition),
+                        RepoClientType.READ,
+                    )
+                    repo_manager = RepoManager(
+                        repo_client, organization_id=repo_archive.organization_id
```
roaga, line 716: probably a rare scenario, but what if for example, an org revokes GitHub access for Seer. They don't use Seer for 30 days. Will we still be able to delete their repo (i.e. wouldn't creating a repo client here fail)?

jennmueng: Hmm you're right, can make it where if the repo can't be accessed we delete it. But let's see if we keep caching after chatting today.

jennmueng: Because we're leaving this code and continuing off it later, I left a TODO comment for this beacause it'll need a bit of a refactor of the delete code and the repo manager initialization

src/celery_app/tasks.py
```diff
@@ -33,20 +32,26 @@ def setup_periodic_tasks(sender, config: AppConfig = injected, **kwargs):
             name="Delete old Automation runs for 30 day time-to-live",
         )
 
-    if config.is_autofix_backfill_enabled:
-        sender.add_periodic_task(
-            crontab(minute="*/30", hour="*"),  # run once every 30 minutes
-            collect_all_repos_for_backfill.signature(kwargs={}, queue=config.CELERY_WORKER_QUEUE),
-            name="Collect all repos for backfill every 30 minutes",
-        )
+    # if config.is_autofix_backfill_enabled:
```
roaga, line 35: maybe leave a note here?

roaga (approved):
lgtm, just noting follow ups:
- delete any archives in GCS
- switch back to tarball download instead of git clone
- disable GCS uploads/downloads for everyone (i think that's already done)

PR #2683, opened 2025-05-28
fix(autofix): Attempt to fix the AIML 319 unicode char issue

Mihir-Mavalankar:
## PR details
+ This PR tries to fix the bug found documented in this [ticket](https://linear.app/getsentry/issue/AIML-319/unicode-character-issue-during-applying)
+ This issues stems from chardet (encoding detector library) detecting some other encoding when it's actually just utf-8. Most files in Github are utf-8.
+ With this change we try to decode the file as utf-8 and if that fails we try chardet's suggestion. 
+ I also changed the detect function name as now it gives out the decode content too.


src/seer/automation/utils.py
```diff
@@ -251,19 +251,26 @@ def extract_parsed_model(completion: ParsedChatCompletion[T]) -> T:
...
     try:
-        result = chardet.detect(raw_data)
-        encoding = result["encoding"] if result["confidence"] > 0.9 else fallback_encoding
-    # if something went wrong, fallback
-    except Exception as e:
-        logger.exception(f"Error detecting encoding of data: {e}")
-        encoding = fallback_encoding
-
-    return encoding
+        # Try UTF-8 first, since it's the standard for most Github repos
```
roaga, line 261:
```suggestion
        # Try UTF-8 first, since it's the most common file encoding
```

src/seer/automation/utils.py
```diff
@@ -251,19 +251,26 @@ def extract_parsed_model(completion: ParsedChatCompletion[T]) -> T:
...
+    except UnicodeDecodeError:
+        # If it's not utf-8, use chardet to guess the encoding
+        detected = chardet.detect(raw_data)
+        encoding = detected.get("encoding")
+        if not encoding:
+            raise Exception("Could not detect encoding with chardet.")
+        try:
+            return raw_data.decode(encoding), encoding
+        except Exception as e:
+            raise Exception(f"Could not decode raw data with detected encoding '{encoding}': {e}")
```
roaga, line 273: should we raise a more specific exception type?

tests/data/autofix_full_finished_run.json
```diff
@@ -232,7 +232,7 @@
             ],
```
roaga, line 232: why are you touching these json files?

src/seer/automation/utils.py
Mihir-Mavalankar: Okay I'll raise a UnicodeDecodeError

src/seer/automation/utils.py
```diff
@@ -251,19 +251,28 @@ def extract_parsed_model(completion: ParsedChatCompletion[T]) -> T:
...
+        detected = chardet.detect(raw_data)
+        encoding = detected.get("encoding")
+        if not encoding:
+            raise Exception("Could not detect encoding with chardet.")
+        try:
+            return raw_data.decode(encoding), encoding
+        except UnicodeDecodeError as e:
+            raise UnicodeDecodeError(
+                f"Could not decode raw data with detected encoding '{encoding}': {e}"
+            )
```
aayush-se, line 275: Is there a way to split up this nested try-catch? Its a bit hard to follow on an immediate pass especially since the nested except is catching a `UnicodeDecodeError` as well. Perhaps splitting it up into 2 separate try-catches sequentially? 

Mostly just a readability and debuggability thing imo but feel free to ignore if this is indeed the best way

src/seer/automation/utils.py
```diff
@@ -251,19 +251,28 @@ def extract_parsed_model(completion: ParsedChatCompletion[T]) -> T:
     return structured_message.parsed
 
 
-def detect_encoding(raw_data: bytes, fallback_encoding: str = "utf-8"):
+def decode_raw_data(raw_data: bytes):
```
aayush-se, line 254: nit: include return type

roaga (approved):

PR #2685, opened 2025-05-29
chore(anomaly-detection): log the request  payload for combo detection calls that take too long

ram-senth:


tests/seer/anomaly_detection/test_anomaly_detection.py
```diff
@@ -536,6 +536,62 @@ def slow_function(
...
+        self, mock_add_attachment, mock_batch_detect
+    ):
+        def slow_function(
+            ts_internal,
+            config,
+            window_size,
+            algo_config,
+            time_budget_ms=None,
+            prophet_forecast_len_days=None,
+        ):  # -> Tuple[MPTimeSeriesAnomaliesSingleWindow, pd.DataFrame]:
```
aayush-se, line 551: can remove the return type if not used

PR #2688, opened 2025-05-29
Mock CodecovClient API calls

kddubey:
[my deploy failed](https://github.com/getsentry/seer/actions/runs/15333105668/job/43144410334) for a PR that's independent of this code b/c it looks like these API calls finally failed after some API change(?). pretty sure these need to be mocked

### DIFF
--- a/tests/automation/codegen/test_retry_unit_test_coding_component.py
+++ b/tests/automation/codegen/test_retry_unit_test_coding_component.py
@@ -8,6 +8,14 @@ from seer.automation.models import FileChange
 from seer.db import DbPrContextToUnitTestGenerationRunIdMapping
 
 
+@patch(
+    "seer.automation.codegen.retry_unittest_coding_component.CodecovClient.fetch_test_results_for_commit",
+    return_value=[],
+)
+@patch(
+    "seer.automation.codegen.retry_unittest_coding_component.CodecovClient.fetch_coverage",
+    return_value="mocked coverage",
+)
 class TestRetryUnitTestCodingComponent:
     @pytest.fixture
     def component(self):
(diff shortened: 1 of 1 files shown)


rohitvinnakota-codecov (approved):

kddubey:
> ### Sentry detected 1 potential issue in your recent changes
> The component may crash due to a TypeError when formatting a string with None values returned from CodecovClient methods.
> * Description:  The `RetryUnitTestCodingComponent.invoke()` method calls `RetryUnitTestPrompts.format_continue_unit_tests_prompt()` without null checks on the `code_coverage_data` and `test_result_data` variables. These variables can be `None` if the `CodecovClient.fetch_coverage()` or `CodecovClient.fetch_test_results_for_commit()` methods fail. Passing a `None` value to the `.format()` method in `RetryUnitTestPrompts.format_continue_unit_tests_prompt()` will raise a `TypeError` in Python 3.7+, crashing the server. The tests in the pull request mock the `CodecovClient` methods to always return non-None values, masking the bug.
> * Code location:  [src/seer/automation/codegen/retry_unittest_coding_component.py:66~84](https://github.com/getsentry/seer/blob/3834622b0f1d8c7b14bc5d390dc4412be905e71b/src/seer/automation/codegen/retry_unittest_coding_component.py#L66-L84)
> * Suggested fix:  Add null checks for `code_coverage_data` and `test_result_data` in `RetryUnitTestCodingComponent.invoke()` before passing them to `RetryUnitTestPrompts.format_continue_unit_tests_prompt()`. Handle the case where these values are None, possibly by providing default values or skipping the formatting step.
> 
> Did you find this useful? React with a 👍 or 👎

maybe worth pointing out since the prompt could arguably handle the `None` output a little better by, e.g., explicitly saying there's no data. but it's definitely wrong and a weird hallucination to say it'll cause a TypeError in 3.7+ lol

i kind of like that it pointed out that the change to the test can cause a prod problem down the line (assuming its logic is right). not sure how others feel about it

PR #2690, opened 2025-05-29
fix(bug-prediction): Bump formatter and hypotheses structure models

kddubey:
these steps require some intelligence. should update to almost-latest models


suejung-sentry (approved):

src/seer/automation/codegen/bug_prediction_component.py
```diff
@@ -166,7 +166,7 @@ def invoke(
                 prompt=BugPredictionPrompts.format_prompt_structured_hypothesis(
                     hypotheses_unstructured
                 ),
-                model=GeminiProvider.model("gemini-2.0-flash-001"),
+                model=GeminiProvider.model("gemini-2.5-flash-preview-04-17"),
```
kddubey, line 169: note to self: no PT for the May preview model so using April

PR #2695, opened 2025-05-30
chore(anomaly-detection): support for source alert id and type

ram-senth:
Add support for source_id and source_type (in addition to the existing alert_id). With this change, an alert can be referred to using the original `alert_id` or the new `source_id` + `source_type`. The store, delete and detect APIs can now be called with either of these ids or both. If the APIs are called with both then the system will prioritize alert_id over the other id.


tests/seer/anomaly_detection/test_accessors.py
```diff
@@ -24,39 +25,269 @@
...
+            with self.assertRaises(ClientError) as e:
+                alert_data_accessor.query(
+                    external_alert_id=external_alert_id,
+                    external_alert_source_id=external_alert_source_id,
+                    external_alert_source_type=external_alert_source_type,
+                )
+                self.assertEqual(
+                    str(e.exception),
+                    error_message,
+                )
```
aayush-se, line 150: Should use pytest assert here and across the file:

``` python
assert str(e.exception) == error_message
```

ram-senth: The full file needs to migrated to pytest, not just the assert statement. I played with it a bit. Too big to tackle with this change.

tests/seer/anomaly_detection/test_accessors.py
```diff
@@ -325,71 +587,618 @@ def test_save_alert(self):
...
+        )
+        self.assertEqual(
+            alert_from_db.external_alert_source_type,
+            external_alert_source_type,
+            "external_alert_source_type should match",
+        )
+
+    def test_timepoint_save_invalid_alert_id(self):
+        alert_data_accessor = DbAlertDataAccessor()
+        alert_id = random.randint(1, 1000000)
```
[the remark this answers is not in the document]
ram-senth: Just a good practice - some parallel test frameworks can cause interference.

PR #2704, opened 2025-06-02
chore(trace queries): Remove Time Prompt

aayush-se:
- Remove the time based queries in the prompt
  - This functionality will be revamped to support absolute time searches as part of the `statsPeriod` rather than the query itself
- Also fix a few definitions causing erroneous outcomes

### DIFF
--- a/src/seer/automation/assisted_query/prompts.py
+++ b/src/seer/automation/assisted_query/prompts.py
@@ -45,7 +45,7 @@ def get_cache_prompt(
 
         # Search Query and Syntax Guidelines:
         You will use the user's query below to create a valid Sentry search query. The search query will be used to filter the traces and spans that are displayed in the Trace Explorer page and ultimately help you answer the user's question.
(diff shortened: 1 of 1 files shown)


src/seer/automation/assisted_query/prompts.py
```diff
@@ -22,7 +22,9 @@ def get_cache_prompt(
         ---
         # Key Concepts:
         - Trace:
-          - A trace represents a single transaction or request through your system. This includes things like user browser sessions, HTTP requests, DB queries, middleware, caches and more.
+          - A trace represents a collection of one or more transactions through your system.
+        - Transaction:
+          - A transaction represents a single instance of a service being called. This includes things like user browser sessions, HTTP requests, DB queries, middleware, caches and more.
```
roaga, line 27: most of those examples sound more like spans that transactions, no?

roaga (approved):

PR #2706, opened 2025-06-02
chore(logging): Move run_id to extra

kddubey:


kddubey:
> what's the benefit of this?

more queryable in GCP and Sentry, and looks less annoying

PR #2707, opened 2025-06-02
chore(bug-prediction): Don't run on draft until it's opened

kddubey:
based on my understanding of https://github.com/orgs/community/discussions/25722, this should not run the bug predictor on draft PRs, run it when it's open for review (no commit needed), and then run it on future commits 

putting this PR in draft mode, then will open for review to test this change


kddubey:
ok looks like it's right

aayush-se (approved):
🙏 thank you!

PR #2709, opened 2025-06-02
feat(llm-client): LLM configuration fallback mechanism

jennmueng:


src/seer/automation/agent/client.py
```diff
@@ -1632,59 +1941,112 @@ def generate_structured(
...
+                    reasoning_effort=resolved.reasoning_effort,
                 )
-            elif model.provider_name == LlmProviderType.ANTHROPIC:
-                raise NotImplementedError("Anthropic structured outputs are not yet supported")
-            elif model.provider_name == LlmProviderType.GEMINI:
-                model = cast(GeminiProvider, model)
+            elif model_to_use.provider_name == LlmProviderType.ANTHROPIC:
+                model_cast = cast(AnthropicProvider, model_to_use)
+                # Note: AnthropicProvider's generate_structured method doesn't exist, using generate_text instead
+                # This appears to be an issue in the original code - AnthropicProvider doesn't have generate_structured method
```
roaga, line 2009: what's with this comment?

src/seer/automation/agent/client.py
```diff
@@ -1790,31 +2236,55 @@ def generate_text_from_web_search(
...
+            )
+
+        if models is None and model is None:
+            raise ValueError("Must specify either 'model' or 'models' parameter.")
+
+        # Convert single model to list for unified processing
+        models_to_try = models if models is not None else ([model] if model is not None else [])
+
+        if not models_to_try:
+            raise ValueError("At least one model must be provided.")
```
roaga, line 2258: again why not just accept only the list parameter?

src/seer/automation/autofix/components/change_describer.py
```diff
@@ -59,16 +60,32 @@ class ChangeDescriptionComponent(BaseComponent[ChangeDescriptionRequest, ChangeD
...
+                GeminiProvider.model("gemini-2.0-flash-001", region="europe-west4"),
+            ],
+        }
+        us_config = {
+            "models": [
+                GeminiProvider.model("gemini-2.0-flash-001", region="us-central1"),
+                GeminiProvider.model("gemini-2.0-flash-001", region="us-east1"),
+            ],
+        }
+
```
roaga, line 80: I feel like users of `client.py` should not have to worry about regions. Especially if we're going to have other teams using Seer in the future. And wouldn't it also make the code simpler? Like what if each provider just had a hardcoded mapping of available models and what region they should go to depending on Seer region?

The users would only have to select the models to use from the available list

jennmueng: The original intention was to use the existing setup we have for regions and passing `region` here will overwrite it, so any users could just use the llm client without needing to worry about regions.

But now that I think about it, would be nice to have a hardcoded region preference per model while keeping this functionality.

src/seer/automation/agent/client.py
jennmueng: leftover from cursor it seems

src/seer/automation/agent/agent.py
```diff
@@ -34,7 +34,8 @@ class RunConfig(BaseModel):
         default=16, description="Maximum number of iterations the agent can perform"
     )
     max_tokens: int | None = None
-    model: LlmProvider
+    model: LlmProvider | None = None
+    models: list[LlmProvider] | None = None
```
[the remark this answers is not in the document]
jennmueng: Intention was to keep the API for the LlmClient simple for external users, where they don't have to worry about the `models` functionality unless they explicitly want fallbacking

I'll refactor the behavior slightly

src/seer/automation/autofix/autofix_agent.py
```diff
@@ -126,6 +125,11 @@ def _omit_biggest_n_tool_messages(self, messages: list[Message], n: int) -> list
 
         return messages
 
+    def _fallback_callback(self):
+        self.context.event_manager.add_log(
+            "Our LLM provider is overloaded. Now retrying desperately..."
+        )
+
```
jennmueng, line 132: This will now be logged for region _and_ model fallbacks

PR #2710, opened 2025-06-02
Draft: Merged context classes to make a new base class

Mihir-Mavalankar:
## PR Details
+ This PR addresses the work in this [ticket](https://linear.app/getsentry/issue/AIML-443/refactor-autofix-and-codegen-context-to-inherit-from-base-context).
+ It creates anew Base context class and AutofixContext and CodegenContext inherit from this class. 
+ It implements all the common methods to reduce code duplication.

### DIFF
--- a/src/seer/automation/codegen/codegen_context.py
+++ b/src/seer/automation/codegen/codegen_context.py
@@ -1,16 +1,11 @@
 import logging
 
 from seer.automation.agent.models import Message
-from seer.automation.codebase.repo_client import (
-    RepoClientType,
-    autocorrect_repo_name,
-    get_repo_client,
-)
 from seer.automation.codegen.codegen_event_manager import CodegenEventManager
 from seer.automation.codegen.models import CodegenContinuation, UnitTestRunMemory
 from seer.automation.codegen.state import CodegenContinuationState
+from seer.automation.context import BasePipelineContext
 from seer.automation.models import RepoDefinition
-from seer.automation.pipeline import PipelineContext
 from seer.automation.state import DbStateRunTypes
 from seer.db import DbPrContextToUnitTestGenerationRunIdMapping, DbRunMemory, Session
 
@@ -22,7 +17,7 @@ RepoKey = RepoExternalId | RepoInternalId
 RepoIdentifiers = tuple[RepoExternalId, RepoInternalId]
 
 
-class CodegenContext(PipelineContext):
+class CodegenContext(BasePipelineContext):
     state: CodegenContinuationState
     event_manager: CodegenEventManager
     repo: RepoDefinition
@@ -47,35 +42,6 @@ class CodegenContext(PipelineContext):
 
         return cls(state)
 
-    @property
-    def run_id(self) -> int:
-        return self.state.get().run_id
-
-    @property
-    def signals(self) -> list[str]:
-        return self.state.get().signals
-
-    @signals.setter
-    def signals(self, value: list[str]):
-        with self.state.update() as state:
-            state.signals = value
-
-    def get_repo_client(
-        self,
-        repo_name: str | None = None,
-        repo_external_id: str | None = None,
-        type: RepoClientType = RepoClientType.READ,
-    ):
-        repos = self.state.get().readable_repos
-        return get_repo_client(
-            repos=repos, repo_name=repo_name, repo_external_id=repo_external_id, type=type
-        )
-
-    def autocorrect_repo_name(self, repo_name: str) -> str | None:
-        return autocorrect_repo_name(
-            readable_repos=self.state.get().readable_repos, repo_name=repo_name
-        )
-
     def get_file_contents(
         self, path: str, repo_name: str | None = None, ignore_local_changes: bool = False
     ) -> str | None:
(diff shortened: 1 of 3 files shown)


roaga (commented):
Seems like `get_file_contents` could also be shared in the base class? I also wonder if some of the autocorrect methods from `autofix_context` should be moved up to the base class as well so others can reuse them if they want.

Otherwise LGTM. Should get a review from @kddubey for the codecov side

kddubey:
> Suspicion of an `AttributeError` in `BasePipelineContext.does_file_exist` due to accessing a **non-existent `codebases` attribute on `CodegenContinuation`** state objects, potentially causing unexpected behavior.

pretty sure the monke is right?

a larger problem might be that there aren't parent classes for `state` and `event_manager` to (1) specify what the `state`'s continuation needs to have, e.g., `readable_repos` and (2) to make typing happier

Mihir-Mavalankar:
> Seems like `get_file_contents` could also be shared in the base class? I also wonder if some of the autocorrect methods from `autofix_context` should be moved up to the base class as well so others can reuse them if they want.
> 
> Otherwise LGTM. Should get a review from @kddubey for the codecov side

I can add `get_file_contents` as an abstract method since the implementations in the child classes are different. I can make a follow up PR for moving more `autofix_context` classes to the base class

Mihir-Mavalankar:
> > Suspicion of an `AttributeError` in `BasePipelineContext.does_file_exist` due to accessing a **non-existent `codebases` attribute on `CodegenContinuation`** state objects, potentially causing unexpected behavior.
> 
> pretty sure the monke is right?
> 
> a larger problem might be that there aren't parent classes for `state` and `event_manager` to (1) specify what the `state`'s continuation needs to have, e.g., `readable_repos` and (2) to make typing happier. i'm not certain these need to be there, just flagging

The state and event manager will be different based on the base class hence the `Any` there in the base class. Can we have a single state class and event manager class for code gen and autofix? @kddubey

roaga (commented):
I think the AI/Kush's comment needs to be addressed still, no?

kddubey:
@Mihir-Mavalankar is the [updated bug pred](https://github.com/getsentry/seer/pull/2710#issuecomment-2932502744) correct? if so you can also 👍. if it's wrong then definitely 👎. thanks 🙏

one difference between codegen and autofix is that codegen originates from PRs, so it prolly makes sense for there to be a `self.repo`. i'll double check that `readable_repos` is being used in the right places as well

Mihir-Mavalankar:
> @Mihir-Mavalankar is the [updated bug pred](https://github.com/getsentry/seer/pull/2710#issuecomment-2932502744) correct? if so you can also 👍. if it's wrong then definitely 👎. thanks 🙏
> 
> one difference between codegen and autofix is that codegen originates from PRs, so it prolly makes sense for there to be a `self.repo`. i'll double check that `readable_repos` is being used in the right places as well

Thta's good catch. 👍 's it. Updated the PR too. @kddubey

src/seer/automation/context.py
```diff
@@ -0,0 +1,61 @@
...
+    autocorrect_repo_name,
+    get_repo_client,
+)
+from seer.automation.pipeline import PipelineContext
+
+
+class BasePipelineContext(PipelineContext, ABC):
+    state: Any  # Child class specific state
+    event_manager: Any
+    repos: list
```
kddubey, line 17: nit: type as `list[RepoDefinition]`

tests/automation/codegen/test_relevant_warnings.py
```diff
@@ -543,32 +543,23 @@ def test_invoke(self, test_case: _TestInvokeTestCase, component: FilterWarningsC
...
         mock_context = MagicMock(spec=CodegenContext)
-        mock_context.repo = MagicMock()
-        mock_context.repo.provider = "github"
-        mock_context.repo.provider_raw = "integrations:github"
-        mock_context.repo.external_id = "123123"
+        mock_context.repos = [mock_repo]
         mock_context.run_id = 1
         return FetchIssuesComponent(mock_context)
 
-    def test_bad_provider_raw(self, component: FetchIssuesComponent):
```
kddubey, line 558: why rm this method?

src/seer/automation/context.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def get_repo_client(
+        self,
+        repo_name: str | None = None,
+        repo_external_id: str | None = None,
+        type: RepoClientType = RepoClientType.READ,
+    ) -> RepoClient:
+        return get_repo_client(
+            repos=self.repos, repo_name=repo_name, repo_external_id=repo_external_id, type=type
+        )
```
kddubey, line 40: seems to differ from the current behavior of `CodegenContext`, which currently supplies `self.state.get().readable_repos` here. may need to override in the child

for context, `CodegenContext` had `self.repo` for the origin PR's repo, and `readable_repos` for all repos that tools can search over

src/seer/automation/codegen/codegen_context.py
```diff
@@ -31,9 +26,10 @@ def __init__(
         self,
         state: CodegenContinuationState,
     ):
+
         request = state.get().request
 
-        self.repo = request.repo
+        self.repos = [request.repo]
```
kddubey, line 32: can keep `self.repo`, no? am thinking that `self.repo` is more straightforward for codegen b/c codegen tasks originate from a PR from a repo

kddubey (commented):
ty for refactoring!

autofix changes look good

i think there's one bug w/ the codegen changes. after that will approve

PR #2718, opened 2025-06-04
feat(llm-client): LlmClient region & model fallbacking

jennmueng:
Implements a fallback mechanism in the llm client.
- First, the fallback goes regional, keeping the same model and going through the `region_preferences`
- Then, it goes to the `models` list if provided.

1. `models`
    - The `LlmClient` now takes in either the existing `model` param or `models` param to enable model-level fallbacking.
2. `region_preferences`
    - Cleans up the implementation of region handling for LLMs to now be defined inside each provider's llm default configuration.
3. Pass region & model parameters to the `LlmProvider.model` call
    - Can specify region & model parameters in each provider's `.model` call
4. Backoff retriers are now handled at the provider client level like how we had with `GeminiProvider` before instead of in the `AutofixAgent`.
    - Stream timeout errors and 422/429 errors will trigger the fallback but provider specific errors will still trigger the provider-specific retry
    - Stream timeout errors should still retry with the same before triggering regional fallback.

## Next Steps
- Implement model fallbacking for the various use cases
- Implement the "OpenRouter-lite" db-stored model configuration blacklist we talked about as a followup to this
- Provide fallbacks to other use cases such as issue summary etc.


src/seer/automation/autofix/autofix_agent.py
```diff
@@ -208,55 +205,55 @@ def _get_completion(self, run_config: RunConfig, messages: list[Message]):
                 tool_calls.append(chunk)
             elif isinstance(chunk, Usage):
                 usage += chunk
+            elif isinstance(chunk, StreamResetException):
```
jennmueng, line 208: When a stream hits a backoff/retry or a fallback it'll emit this exception to tell the consumer to reset the content from the last stream

src/seer/automation/agent/client.py
```diff
@@ -625,55 +743,64 @@ def generate_text(
...
+                    provider_name=self.provider_name,
+                    usage=usage,
+                ),
+            )
+
+        return backoff_on_exception(
+            is_exception_retryable=self.is_completion_exception_retryable,
+            max_tries=self.backoff_max_tries,
+            sleep_sec_scaler=self.sleep_sec_scaler,
+        )(_generate_text)()
```
jennmueng, line 803: Alot of these include just wrapping the calls with `backoff_on_exception`

src/seer/utils.py
```diff
@@ -218,3 +220,67 @@ def tqdm_sized(
         for batch in iterable:
             pbar.update(length(batch))
             yield batch
+
+
+def backoff_on_generator(
```
jennmueng, line 225: New equivalent backing off util but for generators

src/seer/automation/agent/client.py
```diff
@@ -954,109 +1113,37 @@ def construct_message_from_stream(
...
+                # Leave blank for de as preview models are not available in de
+            },
+        ),
         LlmModelDefaultConfig(
             match=r".*",
             defaults=LlmProviderDefaults(temperature=0.0),
+            region_preference={
+                "us": ["global", "us-central1", "us-east1"],
+                "de": ["europe-west1", "europe-west4"],
+            },
```
jennmueng, line 1133: Intention is the pain today with this big refactor allows us to much more easily define configurations for LLMs as seen and makes it much, much, more clearer to the reader what is happening with our region logic

src/seer/automation/agent/client.py
```diff
@@ -158,7 +202,7 @@ class OpenAiProvider(BaseLlmProvider):
     default_configs: ClassVar[list[LlmModelDefaultConfig]] = [
         LlmModelDefaultConfig(
             match=r"^o.*",
-            defaults=LlmProviderDefaults(temperature=1.0),
+            defaults=LlmProviderDefaults(temperature=1.0, first_token_timeout=90.0),
```
jennmueng, line 205: This replaces the code we used to have in `autofix_agent.py`

```py
            first_token_timeout=(
                90.0
                if (run_config.model.provider_name == LlmProviderType.OPENAI)
                and run_config.model.model_name.startswith("o")
                else 40.0
            ),
```

src/seer/automation/autofix/autofix_agent.py
```diff
@@ -208,55 +205,55 @@ def _get_completion(self, run_config: RunConfig, messages: list[Message]):
...
+                content_chunks = []
+                thinking_content_chunks = []
+                tool_calls = []
+                thinking_signature = None
+
+                self.accumulated_thinking_chunks = []
+
+                with self.context.state.update() as cur:
+                    cur_step = cur.steps[-1]
+                    cur_step.clear_output_stream()
```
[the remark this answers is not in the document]
kddubey: do you need to add `cleared = True`?

kddubey (approved):
🙏  one request to log the model fallback to sentry. can table the nits ofc

src/seer/automation/agent/client.py
jennmueng: I tried the former suggestion of having a `generate_text()` and a `_generate_text()` but now we'll have to define all the arguments and their types all over again to get clean type hints, makes it harder to maintain

I like the latter suggestion, will implement

src/seer/automation/agent/client.py
```diff
@@ -1710,85 +2232,196 @@ def generate_text_stream(
...
+
+                                yield StreamResetException()
+
+                                continue
+                            elif not is_last_model:
+                                # Try next model
+                                logger.warning(
+                                    f"Text stream generation failed with {model_to_use.provider_name} model '{model_to_use.model_name}' (all regions tried) "
+                                    f"due to {type(e).__name__}: {str(e)}. "
+                                    f"Trying next model ({i + 2}/{len(models_to_try)}): {models_to_try[i + 1].provider_name} '{models_to_try[i + 1].model_name}'"
```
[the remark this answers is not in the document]
jennmueng: let's actually log all fallback at error level for monitoring purposes on sentry

PR #2719, opened 2025-06-04
fix(autofix): Claude 3.7 sonnet to global in us

jennmueng:


jennmueng:
Bypassing w/o review to resolve #inc-1188.

PR #2722, opened 2025-06-04
chore(fixability): Disable langfuse

kddubey:
we trace this call in sentry, and store scores in both seer `issue_summary` and sentry `sentry_groupedmessage`, so scores are easy to get.

looking at profiles, the langfuse call can add up to 150ms latency for an op that takes 10-20ms.

### DIFF
--- a/src/seer/automation/summarize/issue.py
+++ b/src/seer/automation/summarize/issue.py
@@ -1,7 +1,7 @@
 import textwrap
 
 import sentry_sdk
-from langfuse.decorators import langfuse_context, observe
+from langfuse.decorators import observe
 from pydantic import BaseModel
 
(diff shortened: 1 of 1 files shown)


src/seer/automation/summarize/issue.py
```diff
@@ -261,6 +258,6 @@ def evaluate_autofixability(
         f"Possible cause: {issue_summary.possible_cause}"
     )
     score = autofixability_model.score(issue_summary_input)
-    is_fixable = score > 0.663  # This flag isn't used. Thresholds are in getsentry/sentry
+    is_fixable = score > 0.66  # This flag isn't used. Thresholds are in getsentry/sentry
```
kddubey, line 261: just to be consistent w/ the threshold in sentry

PR #2727, opened 2025-06-06
chore(automation): Trace repo_client init

kddubey:
exploring feasibility of reading code context for issue summary


roaga (approved):
No harm in adding this but we can see from profiles already that initializing a repo client takes 1-2 seconds usually

PR #2729, opened 2025-06-06
feat(autofix): Model fallbacking + sonnet 4

jennmueng:
Provide lists of models to fallback to for Autofix use cases, upgrade to Claude Sonnet 4, and use our Sonnet 4 PT with good fallbacks

### DIFF
--- a/src/seer/automation/agent/client.py
+++ b/src/seer/automation/agent/client.py
@@ -705,7 +705,7 @@ class AnthropicProvider(BaseLlmProvider):
             project_id=project_id,
             region=selected_region,
             base_url=base_url,
-            max_retries=3,
+            max_retries=1,
         )
 
         return client
--- a/src/seer/automation/agent/client.py
+++ b/src/seer/automation/agent/client.py
@@ -1284,9 +1284,21 @@ class GeminiProvider(BaseLlmProvider):
             langfuse_context.update_current_observation(
                 model=self.model_name,
                 usage=Usage(
-                    completion_tokens=response.usage_metadata.candidates_token_count,
-                    prompt_tokens=response.usage_metadata.prompt_token_count,
-                    total_tokens=response.usage_metadata.total_token_count,
+                    completion_tokens=(
+                        response.usage_metadata.candidates_token_count
+                        if response.usage_metadata
+                        else None
+                    ),
+                    prompt_tokens=(
+                        response.usage_metadata.prompt_token_count
+                        if response.usage_metadata
+                        else None
+                    ),
+                    total_tokens=(
+                        response.usage_metadata.total_token_count
+                        if response.usage_metadata
+                        else None
+                    ),
                 ),
                 metadata={"region": self.region},
             )
--- a/src/seer/automation/agent/client.py
+++ b/src/seer/automation/agent/client.py
@@ -639,11 +639,11 @@ class AnthropicProvider(BaseLlmProvider):
                     "us-east1",
                     "us-east5",
                     "global",
-                ],  # We have sonnet 4 PT for us-east1, us-east5 is anthropic default
+                ],  # We have sonnet 4 PT for us-east1, us-east5 is anthropic default w/ higher quota
                 "de": [
                     "europe-west4",
                     "europe-west1",
-                ],  # We have sonnet 4 PT for europe-west4, europe-west1 is anthropic default
+                ],  # We have sonnet 4 PT for europe-west4, europe-west1 is anthropic default w/ higher quota
                 "*": ["global", "us-east5"],
             },
         ),
@@ -655,7 +655,7 @@ class AnthropicProvider(BaseLlmProvider):
                     "us-east5",
                     "global",
                     "europe-west4",
-                ],  # us-east5 is anthropic default, europe-west4 is where we have it set to before, as a fallback.
+                ],  # us-east5 is anthropic default w/ higher quota, europe-west4 is where we have it set to before, as a fallback.
                 "de": ["europe-west4", "europe-west1"],  # europe-west1 is anthropic default
                 "*": ["global", "us-east5"],
             },
--- a/src/seer/automation/agent/client.py
+++ b/src/seer/automation/agent/client.py
@@ -1281,21 +1281,30 @@ class GeminiProvider(BaseLlmProvider):
                     if each.text:
                         answer += each.text
 
+            usage = Usage(
+                completion_tokens=(
+                    response.usage_metadata.candidates_token_count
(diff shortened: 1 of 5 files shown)


src/seer/automation/agent/client.py
```diff
@@ -631,12 +631,25 @@ class AnthropicProvider(BaseLlmProvider):
     provider_name = LlmProviderType.ANTHROPIC
 
     default_configs: ClassVar[list[LlmModelDefaultConfig]] = [
+        LlmModelDefaultConfig(
+            match=r"^claude-sonnet-4@20250514$",
+            defaults=LlmProviderDefaults(temperature=0.0),
+            region_preference={
+                "us": ["us-east1", "us-east5", "global"],
```
roaga: why prefer `us-east5` over `global` for `us` region?

src/seer/automation/agent/client.py
```diff
@@ -631,12 +631,25 @@ class AnthropicProvider(BaseLlmProvider):
...
+                "de": ["europe-west4", "europe-west1"],
```
roaga: do we have PT in `europe-west4`? Otherwise I don't think this will work for claude sonnet 4

(https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models/claude/sonnet-4)

src/seer/automation/agent/client.py
```diff
@@ -631,12 +631,32 @@ class AnthropicProvider(BaseLlmProvider):
...
 
     default_configs: ClassVar[list[LlmModelDefaultConfig]] = [
+        LlmModelDefaultConfig(
+            match=r"^claude-sonnet-4@20250514$",
+            defaults=LlmProviderDefaults(temperature=0.0),
+            region_preference={
+                "us": [
+                    "us-east1",
+                    "us-east5",
+                    "global",
```
roaga, line 641: why prefer `us-east5` over `global` here?

src/seer/automation/agent/client.py
```diff
@@ -631,12 +631,32 @@ class AnthropicProvider(BaseLlmProvider):
...
             match=r".*",
             defaults=LlmProviderDefaults(temperature=0.0),
             region_preference={
-                "us": ["europe-west4", "global"],
-                "de": ["europe-west4"],
+                "us": [
+                    "us-east5",
+                    "global",
+                    "europe-west4",
+                ],  # us-east5 is anthropic default, europe-west4 is where we have it set to before, as a fallback.
```
roaga, line 658: also here, isn't global better than specific region?

src/seer/automation/agent/client.py
```diff
@@ -689,7 +704,7 @@ def get_client(self, app_config: AppConfig = injected) -> anthropic.AnthropicVer
             project_id=project_id,
             region=selected_region,
             base_url=base_url,
-            max_retries=max_retries,
+            max_retries=3,
```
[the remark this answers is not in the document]
roaga: could set to 0 even and rely on our own retry system

src/seer/automation/autofix/components/root_cause/component.py
```diff
@@ -55,13 +55,23 @@ def invoke(
...
+                de_discovery_config = {
+                    "model": AnthropicProvider.model("claude-sonnet-4@20250514"),
+                    "max_tokens": 8192,
+                }
+                us_discovery_config = {
+                    "models": [
+                        GeminiProvider.model(
+                            "gemini-2.5-flash-preview-04-17",
+                            region="us-central1",  # Only try in this region for this model.
+                        ),
```
roaga, line 67: why? why not global region as a fallback? And wouldn't this be better set as a default config for this model within the client, like you did for claude 4?

src/seer/automation/autofix/components/root_cause/component.py
```diff
@@ -128,17 +142,30 @@ def invoke(
...
+                    "max_tokens": 8192,
+                }
+
+                us_formatter_config = {
+                    "models": [
+                        GeminiProvider.model(
+                            "gemini-2.5-flash-preview-04-17",
+                            region="us-central1",  # Only try in this region for this model.
+                        ),
+                        GeminiProvider.model("gemini-2.5-flash-preview-05-20"),
```
roaga, line 156: same question here

src/seer/automation/autofix/components/solution/component.py
```diff
@@ -178,17 +192,28 @@ def invoke(
...
+                de_config = {
+                    "model": GeminiProvider.model("gemini-2.0-flash-001"),
+                }
+
+                us_config = {
+                    "models": [
+                        GeminiProvider.model(
+                            "gemini-2.5-flash-preview-04-17",
+                            region="us-central1",
+                        ),
```
roaga, line 204: same question here

roaga (approved):

src/seer/automation/agent/client.py
```diff
@@ -1258,6 +1280,17 @@ def _search_the_web():
...
                     if each.text:
                         answer += each.text
+
+            langfuse_context.update_current_observation(
+                model=self.model_name,
+                usage=Usage(
+                    completion_tokens=response.usage_metadata.candidates_token_count,
+                    prompt_tokens=response.usage_metadata.prompt_token_count,
+                    total_tokens=response.usage_metadata.total_token_count,
+                ),
```
kddubey, line 1290: > Description: The code added in the _search_the_web() method of GeminiProvider directly accesses response.usage_metadata.candidates_token_count, response.usage_metadata.prompt_token_count, and response.usage_metadata.total_token_count without checking if response.usage_metadata exists or if the individual token count attributes are present and not None.

looking at the types it seems like this is right?

kddubey: from the [prev edit](https://github.com/getsentry/seer/pull/2729#issuecomment-2950318195) in the comment

src/seer/automation/agent/client.py
```diff
@@ -1258,6 +1280,29 @@ def _search_the_web():
...
                     if each.text:
                         answer += each.text
+
+            langfuse_context.update_current_observation(
+                model=self.model_name,
+                usage=Usage(
+                    completion_tokens=(
+                        response.usage_metadata.candidates_token_count
+                        if response.usage_metadata
+                        else None
```
kddubey, line 1290: `Usage` expects ints so default to `0`. bug predictor didn't catch this type error

kddubey: also type says `response.usage_metadata.candidates_token_count` is int or None so would need to do what's done in `_generate_structured`

PR #2735, opened 2025-06-08
chore(seer): Add link to keep working in pr description

roaga:
Adds a message and a link in the PR descriptions to open Seer directly in the UI and continue working if the user is unhappy with the PR


jennmueng (approved):
great idea

PR #2739, opened 2025-06-09
[fix] Use async task queue to process PR comments

rohitvinnakota-codecov:
The task was not firing previously. Update to use the async task queue similar to other codecov request tasks.


jennmueng (approved):
oh yeah calling invoke on a step directly like that does undefined behavior, never tried it and not sure what it would do

